PR #40, opened 2015-01-02
Adding related property to Activity

matt-leach:
Failed to add a relevant property to the Activity model for get_related_activities in previous pull request.


hozn:
Thanks, I think this also refs #41

PR #48, opened 2015-02-17
fixes for upload_activity

dlenski:
dlenski@772eb42 : The name of the activity should be in parameter `name`, not `activity_name`, and a `description` can be included too. (http://strava.github.io/api/v3/uploads/#post-file)

dlenski@70025d6 : The `upload_activity` response may not contain an `id` field if the response is `401 Unauthorized`


hozn:
Thanks!

PR #51, opened 2015-06-05
Couple of updates

matt-leach:
Various updates:
* removed depreciated pr_distance from Segment model
* added parameter to get_activity to include segment efforts
* added Client.update_athlete method to allow updating the authenticated athlete 

Includes adding an activity_id option to test.ini as some tests require an activity which is owned by the user who owns the access_token.


hozn:
Thank you!

PR #55, opened 2015-06-21
Fix uploading activity as a text.

y0prst:
Unfortunately gave up writing a test. BTW, how do you launch them?

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -625,16 +625,16 @@ class Client(object):
         :type external_id: str
         """
         if not hasattr(activity_file, 'read'):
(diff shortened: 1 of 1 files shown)


hozn:
The tests are a bit tricky, because you have to create your own test.ini that has a strave API token.  And  unfortunately some of the "sample data" that I was using for tests (which was from examples in Strava API) is no longer accessible.  Short of mocking out the Strava side, I'm not sure how to handle the fact that the data in Strava is not especially stable.  (And of course mocking out the Strava API really reduces effectiveness of the test.)

hozn:
Thanks for fix!

PR #57, opened 2015-08-20
Slight fixup for device_watts

djfurie:
Added device_watts to BaseEffort class
Removed a redundant line for device_watts in the Activity class


hozn:
Thanks!

PR #62, opened 2015-10-12
Add support for deauthorization

pR0Ps:
Adds support for the deauthorization functionality of the Strava v3 API detailed at http://strava.github.io/api/v3/oauth/#deauthorization


hozn:
Great thanks!

PR #63, opened 2016-01-01
Change type of undocumated pr_date field on AthleteSegmentStats from …

tnikkel:
…timestamp to date to match what Strava is now sending


hozn:
Thanks!

PR #67, opened 2016-02-26
removed beautifulsoup4 install requirement

tkajstura:
Removed beautifulsoup4 from install_requires in setup.py as it is not used by stravalib.


hozn:
You're right -- beautifulsoup is not needed anymore.  Thanks!

Issue #77, opened 2016-07-14
Webhook Events Support


loisaidasam:
Would be cool to see support for Webhook Events:

http://strava.github.io/api/partner/v3/events/

hozn:
Sure, pull requests welcome!

loisaidasam:
Sweet, looking into it now :)

hozn:
Thank you!

hozn:
Initial commit merged, thank you!  Still outstanding here:
- Remainder of API
- Tests
- Updated user docs

hozn:
Ok, I think this is done.  @loisaidasam maybe you can test out the list/del endpoint when you get this working on your end.  (My app/API key won't get push notifications.)  Date handling should be much improved now with Arrow so sub update event_time will be handled correctly as date.

loisaidasam:
Busy weekend! Nice work. Will definitely test out the list/del endpoints for you. Thanks for the urgency with bringing this functionality in, much appreciated 👍

loisaidasam:
Just tested, and I can confirm that all create, list, and delete functionality do indeed work! 👍

hozn:
Thank you, sir!

PR #79, opened 2016-07-15
Initial webhook events support #77

loisaidasam:
Still TODO: list subscriptions, delete subscriptions, a few other small things

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1328,8 +1328,7 @@ class Client(object):
         """
         callback = model.SubscriptionCallback.deserialize(raw)
         callback.validate(verify_token)
-        challenge = getattr(callback, 'hub.challenge')
-        response_raw = {'hub.challenge': challenge}
+        response_raw = {'hub.challenge': callback.hub_challenge}
         return response_raw
 
     def handle_subscription_update(self, raw):
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1299,6 +1299,68 @@ class Client(object):
         raw = self.protocol.get('/routes/{id}', id=route_id)
         return model.Route.deserialize(raw, bind_client=self)
 
+    def create_subscription(self, client_id, client_secret, callback_url,
+                            object_type=model.Subscription.OBJECT_TYPE_ACTIVITY,
+                            aspect_type=model.Subscription.ASPECT_TYPE_CREATE,
+                            verify_token=model.Subscription.VERIFY_TOKEN_DEFAULT):
+        """
+        Creates a webhook event subscription.
+
+        http://strava.github.io/api/partner/v3/events/#create-a-subscription
+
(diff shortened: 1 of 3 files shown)


stravalib/model.py
```diff
@@ -997,3 +998,51 @@ class Route(LoadableEntity):
...
+    """
+    Represents a Webhook Event Subscription Callback.
+    """
+    def __init__(self, *args, **kwargs):
+        """
+        Special/complicated __init__ because of tricky variable names
+        """
+        setattr(self, 'hub.mode', Attribute(unicode, (SUMMARY, DETAILED)))
+        setattr(self, 'hub.verify_token', Attribute(unicode, (SUMMARY, DETAILED)))
+        setattr(self, 'hub.challenge', Attribute(unicode, (SUMMARY, DETAILED)))
```
hozn: So, I think I see what's going on here ,but I'm wondering if maybe the deserialize() method should be modified to handle "." characters so that we can keep this model definition looking like more traditional models.  I feel like not being able to reference these attributes on the class could be problematic in other areas.

So, for example, change definition to look like:
```python

class SubscriptionCallback(LoadableEntity):
    hub_mode = Attribute(unicode, (SUMMARY, DETAILED))
    # etc.
```

And then change `deserialize()` so that it if detects characters that would be invalid in a python attribute name in the json dictionary keys, it just replaces them with '_'.

Granted, I haven't looked at `deserialize()` code in awhile, but what do you think about that?

stravalib/client.py
```diff
@@ -1299,6 +1299,68 @@ def get_route(self, route_id):
...
+    def list_subscriptions(self, client_id, client_secret):
+        """
+        List current webhook event subscriptions in place for the current application.
+
+        http://strava.github.io/api/partner/v3/events/#list-push-subscriptions
+
+        TODO: Implement this
+        """
+        raise NotImplementedError
+
```
hozn: We probably want to fill in the implementation here before merging this in to master.  Alternatively, we can merge this into a feature branch if you want some assistance implementing this.

stravalib/model.py
loisaidasam: I think that's a great idea, and it should keep the code more consistent. I'll get to it when I have a moment.

stravalib/client.py
loisaidasam: I'm personally a bit eager to get subs set up for a project I'm working on, and listing/deleting subs is a bit less important for me. Happy if someone wants to help on a feature branch, but I also don't see anything wrong with pushing just the create sub functionality for now.

stravalib/model.py
loisaidasam: ✅

stravalib/client.py
hozn: I'll merge it in, though I will hold off on releasing next version until there's implementation.  (I can probably help with that, though I'm gonna tackle the more comprehensive python3 support first.)

PR #82, opened 2016-08-11
added context_entries parameter to get_segment_leaderboard method

jedman:
I added the ability change the number of context entries when requesting a segment leaderboard, documented here: http://strava.github.io/api/v3/segments/#leaderboard


hozn:
Thanks!

PR #83, opened 2016-08-11
use .items() for python2/3 compatibility

Tafkas:
The update_athlete() method in client.py does not work with Python 3.x as 
iteritems() has been removed from Python 3.x.


hozn:
Sorry that I missed this one; thanks for fix!

PR #86, opened 2016-08-14
Fix UnicodeDecodeError

martinogden:
New-style string formatting [doesn't play nice](http://www.voidspace.org.uk/python/weblog/arch_d7_2010_04_10.shtml) when mixing encodings. Use `__unicode__` rather than `__repr__` and ensure everything is unicode.


hozn:
Thanks!

PR #87, opened 2016-08-18
Fix unicode string as name of class

mpavlase:
There exists a list (tuple actually) of accepted types - str or bytes.
Lines like this:
    activity = EntityAttribute("Activity", (SUMMARY, DETAILED))
can be passed as unicode string, that will cause failure.

Adding unicode type to whitelist fix this case.


hozn:
Thanks!

PR #95, opened 2016-11-26
Add new pr_count property.

Wilm0r:
I started getting log noise like this:

WARNING:stravalib.model.Activity:No such attribute pr_count on entity <Activity id=784664167 name=u'Evening commute' resource_state=None>

Looks like this is a new property showing the number of new personal records earned?


hozn:
Thank you!

PR #99, opened 2016-12-23
fix for delete_activity

jonderwaater:
the delete_activity function wasn't working for me. with the proposed change it does.


hozn:
Thanks!

hozn:
Pushed that out in 0.6.6.  Silly mistake; thanks for fixing it.

PR #106, opened 2017-05-19
Add new attributes for Segment

JohnnyLChang:
The Segment's PR can be used to compare with other Athlete PR easily without multiple requests.


hozn:
Thank you!  I need to do a new release with all these fixes!

PR #119, opened 2017-10-04
Adding more info to ActivityUploader and handling errors

bwalks:
I was trying to upload a `tcx.gz` file to Strava, but was hitting timeout issues when waiting for the activity to finish processing. However, the initial request actually failed with a 400. This wasn't captured by `upload_activity` because Strava returned a JSON payload in the format: 

```
{u'errors': [{u'code': u'invalid encoding',
              u'field': u'gzip',
              u'resource': u'upload'}],
 u'message': u'Bad Request',
 }
```

Both `errors` and `message` aren't in the API documentation. Because `errors`, and not `error` was returned, `upload_activity` did not raise an exception.

I also am saving that response info in `ActivityUploader` instances. This allows a consumer to inspect all of the data returned by Strava.

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1596,6 +1596,7 @@ class ActivityUploader(object):
 
         """
         self.client = client
+        self.response = response
         self.update_from_response(response)
 
     def update_from_response(self, response, raise_exc=True):
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1614,7 +1614,7 @@ class ActivityUploader(object):
         self.external_id = response.get('external_id')
         self.activity_id = response.get('activity_id')
         self.status = response.get('status')
-        self.error = response.get('error')
+        self.error = response.get('error') or response.get('errors')
         if raise_exc:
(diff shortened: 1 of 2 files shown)


hozn:
Ok, these changes look fine to merge, though it wasn't clear form this diff if `Client.raise_for_error()` would correctly handle a case where there were multiple errors (i.e. it's a list).

bwalks:
@hozn Looking at the code in `protocol.py`, I don't know why `_handle_protocol_error` didn't catch this. We should have expected `Fault` to get thrown, but it never was.

bwalks:
@hozn let me know if you have any more feedback!

hozn:
@bwalks , sorry, the day job has been crazy.  If you're happy with this code right now, I can merge and we can revisit handling of these errors.  Though it seems like there's a little more work needed there to do this more comprehensively?

bwalks:
@hozn Completely understand. I am happy with it for the time being, and can create an issue to translate the dict to a better error message.

PR #120, opened 2017-10-19
Building from sources

piotr-kubica:
As I'm getting familiar with stravalib I wrote short instruction to build, install and run tests, since this may not be obvious for new contributors.


hozn:
Thanks!

PR #121, opened 2017-11-05
Api rate limiter with functional test closes #111

### LINKED ISSUE #111: Request counting?
It would be nice to have some sort of built in request counter (perhaps even in the logging) to keep track of how many API calls an application is making.

I dug around a little bit in the code but couldn't really find anything (maybe it's hiding right in front of my nose!). Is there something like this? Perhaps at the protocol level?

https://github.com/hozn/stravalib/blob/master/stravalib/protocol.py

Maybe this is already achievable by using some sort of a dummy rate limiter?

piotr-kubica:
Based on the information from HTTP response usage/limits (https://strava.github.io/api/#rate-limiting) a rate
limiter has been implemented.
Changes include:
* creating a new rate limit rule (XRateLimitRule) which handles rate limits and timeouts. The rate limiter is non
blocking - however it throws an appropriate exception when limit reached (RateLimitExceeded) or limit timeout did not expire (RateLimitTimeout) - since the 'short' limit is 15 minutes (600 request can be made asap but than wait 15 from first request for the timeout to expire) I found it more appropriate then using sleep
* modifying RateLimiter and RateLimitRule and protocol so the information from header can be passed
* setting XRateLimitRule as the default rule
* functional test - ClientDefaultRateLimiterTest

I hope it is good enough. It would be great if someone could make a code review :)


piotr-kubica:
I guess this has to merged with #119

jsamoocha:
Thanks for your contribution, I was in need of this and did not have time to work on it yet.

About some details:

- In case of an actual rate limit exceedance, Strava API returns an HTTP 429 error response. This will be raised internally as a `requests.exceptions.HTTPError` by `_handle_protocol_error()`. As a result of this, the call to `rate_limiter` will never happen. So it seems that in practice, the timeout/cooldown mechanism built into `XRateLimitRule` will not work, and neither will it ever raise a `RateLimitExceeded` exception? (This looks like an existing issue and not something that is introduced by this PR).
- It would be nice to have some (optional) wait/sleep functionality in `XRateLimitRule`, similar to that in the existing `RateLimitRule`. End users of a 3rd party app using Strava would find it more tolerable to have multiple latencies of a few seconds each instead of single latencies of potentially several minutes.

PR #124, opened 2017-11-14
More rate limit improvements

jsamoocha:
Introducing a new rate limit rule that uses feedback from the Strava API about actual limit usage. It uses variable cool-down periods after each response processed by the client. The cool-down periods depend on the number of request left in the current period, and the priority given to the rule. There's no need to initialize this rule with explicit limits, it will get these from the API responses as well.

### DIFF
--- a/stravalib/tests/functional/test_client_rate_limiter.py
+++ b/stravalib/tests/functional/test_client_rate_limiter.py
@@ -1,13 +1,11 @@
 from __future__ import division, absolute_import, print_function, unicode_literals
-from stravalib import model, attributes, exc, unithelper as uh
-from stravalib.client import Client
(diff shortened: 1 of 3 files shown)


jsamoocha:
Please review for merge, and test (several functional tests are failing at my end because of dependencies on being a specific Strava user (getting 401 and 404 responses)?)

Thanks!

hozn:
Excellent work.  I'm going to merge, though I have not re-run functional tests yet.  I'll do that before releasing anything.  I need to revisit those yet again to see if there's something better we can do.

hozn:
One thing that would be really useful is any documentation in README.md or in the docs/ directory that provides an overview for using this new XRateLimitRule functionality.  I can help with this too.

PR #133, opened 2018-01-07
Prevent AttributeError when ActivityUploader has no error

facingBackwards:
If there are no errors in the upload response, raise_for_exc raises AttributeError.


hozn:
Thanks!

PR #144, opened 2018-03-29
Add RunningRace model and expose get_running_race(s).

magnific0:
This PR adds the RunningRace class and exposes two get functions:

- client.get_running_races(year)
- client.get_running_race(race_id)

For the first method the year defaults to the current year if empty.

See Strava API here:
https://developers.strava.com/docs/reference/#api-RunningRaces

Also while retrieving club members ```get_club_members(club_id=123)```. I noticed that I got some warnings on three attributes (membership, owner and admin) that are not part of the Athlete class. So I added these to the class too.

Let me know if you have any feedback.

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1338,6 +1338,44 @@ class Client(object):
 
         # Pack streams into dictionary
         return {i.type: i for i in streams}
+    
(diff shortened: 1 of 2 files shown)


hozn:
Cool -- thanks for doing this!

Issue #162, opened 2018-10-16
OAuth Updates


jsamoocha:
See [Strava's message](https://developers.strava.com/docs/oauth-updates/) about the replacement of permanent access tokens by short-lived ones. This may require some thoughts about redesigning the API...

ghn:
Hi! any plan for this update yet?

hozn:
PRs welcome. I haven't had time to look at this yet, no.

hozn:
Released in 0.10.0 -- thanks @david-pacheco

david-pacheco:
I realized there is an issue with my solution. Apparently Strava is not returning the refresh token on the first authentication if the correct "scopes" are not passed in the initial authorization request (according to this post: https://stackoverflow.com/questions/53009519/strava-api-refresh-token-doesnt-issued-after-first-authorization). So the authorize_url method must also be changed...

PR #163, opened 2018-11-01
OAuth Updates #162

david-pacheco:


hozn:
Brilliant -- thanks for taking the time to do this!

PR #167, opened 2018-11-09
Fixes scope verifications after latest update

david-pacheco:
After the latest PR there was still a problem with the scopes validation in need of a fix.


hozn:
Hey, I appreciate the PR; I ended up fixing this slightly differently, but I believe with same result.

PR #169, opened 2018-12-01
Update documentation elaborating on expires_at and refresh_access_token.

jlapenna:
It was a bit ambiguous about how the expires_at token was used; added a small snippet to clarify.


hozn:
Thanks!

PR #200, opened 2020-08-24
Update docs for Client.authorization_url

graham33:
The scope parameter docs were incorrect (presumably out of date).

I copied the docs from the corresponding function in protocol.py.


hozn:
Thank you!  Just cleaning up some open tickets.

Issue #153, opened 2018-05-30
Can't get ID of Activity by using activities = client.get_activities()


prof-schacht:
Hi,

I was wondering how I can get the id of an activity by using the client.get_activities() function. 
When I look into this function and also the activity model behind it, I can't see the field ID. 

My intention was, that I can download all activities from an athlete and then enable the possibility to get in addition the details (streams) of an activity by pressing an button.

When I look into the developer api from strata https://developers.strava.com/docs/reference/#api-Activities-getLoggedInAthleteActivities. I can see that there is an id field which is provided by the api.

Did somebody have an idea how to get the id for collecting the stream data from an activity by having the list of activities?

BR

hozn:
Does this not work for you ... ?
```python

activities = client.get_activities()
for activity in activities:
  print(f"Activity ID: {activity.id}")
```

adv27:
Hi, the result method `to_dict()` not include `id` field

bapBardas:
I'm running into the same issue

darthale:
the `id` in the Activity entity is named `upload_id` rather than `id`.

bapBardas:
> the `id` in the Activity entity is named `upload_id` rather than `id`.

No, this is not correct : for instance i retrieve an `activity.upload_id = '3125668582'` while the `activity.id = 2936266820`
btw upload_id is also a field of the Activity before converting it to dict.

jakearmendariz:
Same issue. Upload ID does not match the id

Great library thank you!

PR #201, opened 2020-09-25
Fix a few fields in Activity model

liskin:
Source:
http://developers.strava.com/docs/reference/#api-models-SummaryActivity,
http://developers.strava.com/docs/reference/#api-models-DetailedActivity

Fixes: #153


hozn:
Thanks for this fix -- just finally getting around to cleaning up some lingering PRs.

PR #203, opened 2020-10-06
Make to_dict() include attributes of parent classes

pR0Ps:
Before this change `to_dict()` would only return values that were defined on the
local class.

For example, `Bike.to_dict()` would return a dict with a single `frame_type` key
since it's the only `Attribute` defined on the class. The rest of the
attributes are defined on the parent classes (`Gear`, `IdentifiableEntity`, etc.)

This commit makes `to_dict()` also look for `Attributes` defined on parent
classes and includes them in the output as well.

### DIFF
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -44,10 +44,11 @@ class BaseEntity(object):
         :rtype: Dict[str, Any]
         """
         d = {}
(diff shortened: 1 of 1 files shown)


hozn:
Thank you!  I agree -- useful improvement.

PR #204, opened 2020-10-06
Now webhooks use the same domain as the rest of API.

thiago-sylvain:
+ Removed unused attributes in webhook creation.


hozn:
Thank you!

PR #205, opened 2020-10-31
Tweak __repr__ of BaseEntity objects

pR0Ps:
Changes:
 - Will only list the `resourse_state` property if it is set to something
 - Won't include a trailing space after the class name if none of the attrs are set


hozn:
Thanks for providing this fix.  Just catching up on some (really old) tickets!

PR #216, opened 2021-08-13
Fixes IndexErrors when deserializing empty lists as GPS locations

### LINKED ISSUE #215: Strava API change for activities without GPS lead to deserialization error
It seems there was an (as far as this moment) unannounced API change in Strava where activities without GPS are now returned with `[]` as the value of the `start_latlng` attribute. I don't know what value was returned before, but now the following snippet in `attributes.py` fails with an `IndexError`:

```
    def unmarshal(self, v):
        """
        """
        if not isinstance(v, LatLon):
            v = LatLon(lat=v[0], lon=v[1])
        return v
```

jsamoocha:
Fixes #215. Since 2021-08-11, the Strava API returns an empty list for e.g. the `start_latlng` attribute.
This fix prevents `IndexError`s when deserializing these attributes.

### DIFF
--- a/stravalib/attributes.py
+++ b/stravalib/attributes.py
@@ -181,7 +181,7 @@ class LocationAttribute(Attribute):
         """
         """
         if not isinstance(v, LatLon):
-            v = LatLon(lat=v[0], lon=v[1])
+            v = LatLon(lat=v[0], lon=v[1]) if v else None
         return v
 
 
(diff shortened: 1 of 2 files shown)


hozn:
Thanks for providing this fix!

yvetterowe:
hey @hozn, can we bump the release version to incorporate this fix? Much appreciated!

hozn:
Yeah, I will aim to get a bumped version released this weekend or early
next week.

ColinChargy:
Hi,
I'm also waiting for the new release to fix this !
Thanks for everything!
Regards,
Colin Chargy

Issue #127, opened 2017-12-04
Create to_dict() method on the model objects


hozn:
Driven most recently by #126 by @dizhu1991, I think it would be useful to have a `to_dict()` method to export loaded attributes.  For initial version we can probably ignore any of the lazy-loaded objects (in future we could add a flat to `to_dict()` to incldue these).

PR #218, opened 2021-08-18
fix: #127

yihong0618:


hozn (commented):
I think I will tweak this slightly to use `six.text_type`, but thanks for demonstrating the fix.

hozn:
Thanks again -- pushed out 0.10.4.

yihong0618:
@hozn Thank you for great repo.~

PR #223, opened 2021-10-08
Refactor deprecated unittest aliases for Python 3.11 compatibility.

tirkarthi:
Deprecated unittest aliases were removed in Python 3.11 . The recommended aliases are present in both Python 2.7 and Python 3. So the PR is backwards compatible.

assertEquals -> assertEqual
assertAlmostEquals -> assertAlmostEqual

Ref : https://github.com/python/cpython/pull/28268


lwasser:
@hozn are you comfortable with merging this one as well? i was starting to implement this when running the tests. it will fix some of the warnings we are getting.

hozn:
Yes, I'm comfortable with this one too.  Thanks!

lwasser:
great! @tirkarthi thank you - merging this!!

PR #226, opened 2021-10-17
Docs build p2

lwasser:
getting closer. this is a cleaner version of the docs with many improvements. i'm concerned that i may have broken things however because i can't run the test suite.


lwasser:
oops sorry wrong repo - closing.

PR #241, opened 2022-10-31
Improved unknown time zone handling

jsamoocha:
Some activities have unknown or incorrect time zones in the JSON response from Strava, e.g., '(GMT+00:00) Factory'. This change will make the unmarshaller from `TimezoneAttribute` return `None` instead of raising an `UnknowTimeZoneError` (and crashing the overall unmarshalling process).

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1374,48 +1374,6 @@ class Client(object):
         # Pack streams into dictionary
         return {i.type: i for i in streams}
 
-    def get_running_race(self, race_id):
-        """
-        Gets a running race for a given identifier.
-
-        TODO: This has been deprecated by strava as of Nov 1 2021. See
-        https://developers.strava.com/docs/changelog/
-        https://developers.strava.com/docs/reference/#api-models-RunningRace
-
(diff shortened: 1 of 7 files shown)


stravalib/client.py
```diff
@@ -1374,48 +1374,6 @@ def get_segment_streams(self, segment_id, types=None, resolution=None,
...
-        if year is None:
-            year = datetime.datetime.now().year
-
-        params = {"year": year}
-
-        result_fetcher = functools.partial(self.protocol.get,
-                                           '/running_races',
-                                           **params)
-
-        return BatchedResultsIterator(entity=model.RunningRace, bind_client=self,
```
yihong0618, line 1415: can you add some explain why delete this in this PR or commit message

jsamoocha: These endpoints don't exist anymore at Strava, and the current build fails because of that.

yihong0618 (approved):
LGTM

lwasser:
y'all should have superpowers to merge when this is ready! i'll work on pypi this week

Issue #231, opened 2021-12-17
[Feat] only support python3.


yihong0618:
Can we only support python3  because python2 is no longer maintained that can make the code clean.

hozn:
Certainly I haven't used python 2 in forever. It probably is time to move
on :)

jsamoocha:
What needs to be done for this other than removing references to python 2 in the documentation?

PR #245, opened 2022-11-02
Test Refactoring

jsamoocha:
This adds a new job to the (now renamed) `build-docs.yml` workflow file that runs flake8 and pytest (and as such replaces the existing nosetest framework by the more modern pytest). Please note that only the _unit_ and _integration_ tests are included in the CI build.

**TODO**:

- [x] Review
- [x] Update Contributing documentation 
- [x] Create separate actions for lint/test/doc, testing should also run on multiple OS


.github/workflows/build-test-docs.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install dependencies
+        run: python3 -m pip install -r ./requirements.txt
+
+      - run: make docs -B
+
+  build-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.7", "3.8", "3.9", "3.10" ]
```
yihong0618, line 47: do we still support 2.7?

yihong0618: https://github.com/stravalib/stravalib/issues/231

yihong0618 (commented):
Is it better to separate `doc yml` and `test yml`

yihong0618:
I think I will do `drop python2` work later we can save some codes.

.github/workflows/build-test-docs.yml
jsamoocha: > Is it better to separate `doc yml` and `test yml`

Sorry for my confusion: is this an open question or suggestion for improvement? I don't consider myself an expert on Github Actions, so any advice is welcome.

> do we still support 2.7?

Good question, TBH I just copy-pasted the versions mentioned in the starter workflow. We may also need to add the recent 3.11?

yihong0618: no need to add 3.11 for now I think.

yihong0618: > Sorry for my confusion: is this an open question or suggestion for improvement? I don't consider myself an expert on Github Actions, so any advice is welcome.

its an open question.~

jsamoocha:
I added minimal working examples of testing client endpoints with a mocked Strava API.

The mocked Strava API can be used as a pytest fixture and is implemented as a thin wrapper around the `RequestsMock` type from the `responses` package. Instead of specifying the entire expected JSON response body for every request, the published example responses by Strava (see the swagger.json file in RESOURCES) are used. They can be tweaked by providing a `response_update` argument to the mock methods.

TODO: dynamically choosing a response from the swagger.json based on the status code in case of non-200 responses.

@yihong0618 @lwasser @hozn, please let me know if this approach works for you. If so, we can continue migrating the other tests.

yihong0618:
> I added minimal working examples of testing client endpoints with a mocked Strava API.
> 
> The mocked Strava API can be used as a pytest fixture and is implemented as a thin wrapper around the `RequestsMock` type from the `responses` package. Instead of specifying the entire expected JSON response body for every request, the published example responses by Strava (see the swagger.json file in RESOURCES) are used. They can be tweaked by providing a `response_update` argument to the mock methods.
> 
> TODO: dynamically choosing a response from the swagger.json based on the status code in case of non-200 responses.
> 
> @yihong0618 @lwasser @hozn, please let me know if this approach works for you. If so, we can continue migrating the other tests.

Looks good to me.

lwasser:
You rock! I will look today.

jsamoocha:
> Looks good to me. can this `PR` be reviewed now or still WIP

Yes, please review it when you have time. The module `tests.integration.test_client` has some examples of how to use the mock for different purposes, which may be a good starting point.

.github/workflows/build-test-docs.yml
```diff
@@ -0,0 +1,78 @@
+name: Build Test and Documentation
+# Build on all branches
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
```
yihong0618, line 9: when pull request only master branch is OK

stravalib/tests/integration/strava_api_stub.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def _api_method_adapter(api_method: Callable) -> Callable:
+    @wraps(api_method)
+    def method_wrapper(
+            *args,
+            response_update: Dict[str, Any] = None,
+            n_results: Optional[int] = None,
+            **kwargs
+    ) -> BaseResponse:
```
yihong0618, line 48: FYI, if we use type hint here we can not run these tests in python2.
But its fine, since we will drop python2 later

yihong0618 (approved):
LGTM, amazing work!

.github/workflows/build-test-docs.yml
lwasser: hi! i think dropping python 2.x makes sense and we should do it. Most packages are doing this as it's end of life. i think generally supporting the latest several versions of python makes sense. we can always add 3.11 later as well. 

i'm going to look at the full build here in just a few but we should also setup a matrix to test on windows, mac and linux -

.github/workflows/build-test-docs.yml
lwasser, line 9:
```suggestion
on:
  pull_request:
  push:
    branches:
      - master
```
This will then build on merge to master and on pr but won't try to do both when someone submits a pr

.github/workflows/build-test-docs.yml
lwasser: so i think we should add a matrix  build. in this example we'd break the build into a docs only build that runs on linux or whatever os we want. Then a second that uses the matrix to run tests across versions of python and operating systems. 

Here is an example of what the matrix build would look like. i'm happy to help with this

https://github.com/earthlab/abc-classroom/blob/main/.github/workflows/run-tests-mac-win-linux.yml#L14
But we don't need in this example the various environments, we just need the os matrix part 
```
  setup-build:
    name: Ex1 (${{ matrix.python-version }}, ${{ matrix.os }})
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix:
        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
        python-version: [3.7, 3.8, 3.9]

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python ${{ matrix.python-version }}
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python-version }}
```

.github/workflows/build-test-docs.yml
```diff
@@ -0,0 +1,78 @@
...
+          python -m pip install --upgrade pip
+          pip install flake8 pytest
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Lint with flake8
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+      - name: Test with pytest
```
lwasser, line 76: so psyched you're using pytest!! yay!

.github/workflows/build-test-docs.yml
```diff
@@ -0,0 +1,78 @@
...
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-py${{ matrix.python-version }}-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-py${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8 pytest
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Lint with flake8
```
lwasser, line 70: so we wouldn't need to lint across os' either. i can ask some friends tomorrow about how they handle linting vs doc build vs tests where you really want a matrix. but my gut is telling me a build for tests and a build for docs and linting OR conditional statements here that only look at docs on one element of the workflow matrix ie 

if python 3.8 and mac os, build docs, lint run tests
else 
only run tests

README.md
```diff
@@ -72,17 +72,14 @@ will build and install *stravalib* in your *pip* package repository.
...
-shell$ nosetests
+shell$ pytest stravalib/tests/unit
 ```
-or
-```
-shell$ nosetests-3
-```
-if you use Python3.
 
 To run **integration** tests you will need to rename *test.ini-example* (which you can find *<your-root-proj-dir>*/stravalib/tests/) to *test.ini*
```
lwasser, line 78: is this replaced by the mock element? i haven't dug into the actual tests yet :)

stravalib/protocol.py
```diff
@@ -163,7 +163,7 @@ def refresh_access_token(self, client_id, client_secret, refresh_token):
 
         return access_info
 
-    def _resolve_url(self, url):
+    def resolve_url(self, url):
```
lwasser, line 166: is this a helper? i'm curious is the rename just for clarity? i'm fine either way. i'm used to [helper  / non public methods using an underscore to follow PEP8 ](https://peps.python.org/pep-0008/#method-names-and-instance-variables)

stravalib/tests/integration/strava_api_stub.py
lwasser: if y'all are ok with dropping python 2.x i really think we should!

lwasser (commented):
This looks really good to me. I am traveling for the next few days and wont be able to run this locally. I defer to you as the expert on how the test suite is designed but would love to better understand how the mock works. I'm wondering if you're using that swagger json response as the expected structure of the return from the actual API?

Kind of like we might do with a cassette but that might be better and simpler? 

In terms of changes - i do think we should run the test suite on a matrix. And then let's just have separate builds for 
tests
docs 
linting? or combine docs and linting?

 a separate build might be easier to look at from a maintenance perspective because with the matrix will will then have python 3.9, windows, 3.9 mac etc etc...

Here is a build from one of my favorite people who is really creating nice clean docs and such in their community. They have a build for  tests, docs, style (linting)

https://github.com/fatiando/boule/tree/main/.github/workflows 

let me know if i can help here!

yihong0618:
> if y'all are ok with dropping python 2.x i really think we should!
will do this after this `PR` merge

stravalib/tests/integration/strava_api_stub.py
jsamoocha: I guess we agree that we'll not be supporting Python 2 any longer, so I'll keep the type hints.

.github/workflows/build-test-docs.yml
jsamoocha: Supposedly, it runs the legacy (unittest) tests without any changes. So why not? ¯\_(ツ)_/¯

README.md
jsamoocha: I still need to modify the documentation in a.o. README.md, based on this review.

stravalib/protocol.py
jsamoocha: I decided to reuse this method outside of the `ApiV3` class to help set up the mocks, so made this method public. Alternatively, we could also refactor by completely moving the server address, API base, and resolve_url function to some generic util module or package __init__ module. Happy to make that change if you think the latter is necessary.

.github/workflows/build-test-docs.yml
jsamoocha: Ok, I'll create separate actions for test/lint/doc, where test is using a matrix as you mentioned above.

stravalib/protocol.py
lwasser: @jsamoocha got it. this makes sense. I will default to your expertise here in setting this up. What would be the benefit of a complete refactor?

thank you again for this work here. it's awesome to have you here helping maintain stravalib.

stravalib/tests/integration/conftest.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+from stravalib import Client
+from stravalib.tests.integration.strava_api_stub import StravaAPIMock
+
+
+@pytest.fixture
+def mock_strava_api():
```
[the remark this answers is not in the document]
hozn: @lwasser , I'll jump in and provide a little context around this.  What @jsamoocha is doing -- and honestly, this is brilliant -- is using Strava's published Swagger/OpenAPI JSON (https://developers.strava.com/swagger/swagger.json) to extract all the known API paths and the examples (which are in that document) that Strava provides as a sort of documentation.  To address your question more specifically, the mock here serves to intercept and replace what would be the real HTTP requests out to the Strava API (which would require an application ID / secret / athletes / etc.) with the responses that were provided by Strava as examples in their Swagger.

I particularly like this approach as:
1. As mentioned above, this removes the need for having to set up any real infrastructure (accounts / apps / etc.) to run the tests
2. The example responses are pulled down with every test run.  Since these will be kept up-to-date by Strava we'll know that if stravalib starts breaking that something in the API is no longer as expected.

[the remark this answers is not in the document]
jsamoocha: Thanks for the explanation, @hozn, while I was busy adding some documentation to the code!

@lwasser, to address your question regarding passthru: because the fetching of the swagger.json file from Strava is happening within the context of the mocked API, that request would've been mocked as well. So we need to use `add_passthru()` to tell the mocker not to interfere with that request.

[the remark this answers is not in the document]
lwasser: awesome. @jsamoocha thank you for all of this work and for the explanation about how this is working. i'll defer to any issues @hozn sees in the PR which it sounds like they really appreciate your approach !! so please don't let me slow down the process of merging this. it all looks good to me !

.github/workflows/build-test-docs.yml
jsamoocha: Yes, good suggestion! It is included in the 3 separate workflow files now.

Issue #250, opened 2022-11-11
Provide (fully covered) integration tests for all client methods


jsamoocha:
For effective CI and PR merging, all existing client endpoint tests from `stravalib.tests.functional` should be moved to or re-recreated in `stravalib.tests.integration`, and use the new `mock_strava_api` fixture to be independent of an actual connection to the Strava API. The tests in `stravalib.tests.integration.test_client` serve as examples.

For each test, let's try to cover all branching logic (and maybe use automated coverage tooling for that?)

The current swagger.json defines the following endpoints that need to be tested:

- [ ] /athletes/{id}/stats
- [ ] /athlete
- [ ] /athlete/zones
- [ ] /segments/{id}
- [ ] /segments/starred
- [ ] /segments/{id}/starred
- [ ] /segment_efforts
- [ ] /segments/explore
- [ ] /segment_efforts/{id}
- [ ] /activities
- [ ] /activities/{id}
- [ ] /athlete/activities
- [ ] /activities/{id}/laps
- [ ] /activities/{id}/zones
- [ ] /activities/{id}/comments
- [ ] /activities/{id}/kudos
- [ ] /clubs/{id}
- [ ] /clubs/{id}/members
- [ ] /clubs/{id}/admins
- [ ] /clubs/{id}/activities
- [ ] /athlete/clubs
- [ ] /gear/{id}
- [ ] /routes/{id}
- [ ] /athletes/{id}/routes
- [ ] /routes/{id}/export_gpx
- [ ] /routes/{id}/export_tcx
- [ ] /uploads
- [ ] /uploads/{uploadId}
- [ ] /activities/{id}/streams
- [ ] /segment_efforts/{id}/streams
- [ ] /segments/{id}/streams
- [ ] /routes/{id}/streams

 This is quite a lot of work, so let's gradually add these tests in small PRs and check them off the list above.

jsamoocha:
#256 takes care of /uploads, /activities{id}, and /athlete.

jsamoocha:
Oops, I forgot to take into account multiple HTTP methods per endpoint, so added these to the checklist.

PR #256, opened 2022-11-15
Test client - upload_activity

jsamoocha:
Fully tests upload_activity(). (part of #250)

I could use some advice on if/how to use `FutureWarning`s in case of seemingly deprecated API parameters.

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -235,9 +235,7 @@ class Client(object):
                                       result_fetcher=result_fetcher,
                                       limit=limit)
 
-
-    # TODO: Doble check these endpoint doc URLs are correct
-    def get_athlete(self, athlete_id=None):
+    def get_athlete(self):
         """
         Gets the specified athlete; if athlete_id is None then retrieves a
         detail-level representation of currently authenticated athlete;
@@ -250,14 +248,7 @@ class Client(object):
         :return: The athlete model object.
         :rtype: :class:`stravalib.model.Athlete`
         """
-        if athlete_id is None:
-            raw = self.protocol.get('/athlete')
-        else:
-            raise NotImplementedError("The /athletes/{id} endpoint was removed by Strava.  "
-                                      "See https://developers.strava.com/docs/january-2018-update/")
-
-            # raw = self.protocol.get('/athletes/{athlete_id}', athlete_id=athlete_id)
-
+        raw = self.protocol.get('/athlete')
         return model.Athlete.deserialize(raw, bind_client=self)
 
     # TODO: this endpoint was removed so do we want to remove the URL altogether?
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -5,12 +5,14 @@ Provides the main interface classes for the Strava version 3 REST API.
 """
 
 from __future__ import division, absolute_import, print_function, unicode_literals
+
+import json
 import logging
-# import warnings # unused import
 import functools
 import time
 import collections
 import calendar
+import warnings
 from io import BytesIO
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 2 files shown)


lwasser:
oh @jsamoocha i can ask some of my colleagues about what the best practices are for this. but question - are we dropping those API elements now? (i am asking to understand better how future warning works). Is future warning for an instance where we might drop this in the next release vs it no longer exists now? 

Also it looks like this PR has some merge conflicts

lwasser:
or maybe @hozn has thoughts on how we handle deprecated api endpoints? it seems like we are dropping them now

stravalib/client.py
```diff
@@ -686,13 +683,24 @@ def upload_activity(self, activity_file, data_type, name=None, description=None,
                               nordicski, alpineski, backcountryski, iceskate, inlineskate,
                               kitesurf, rollerski, windsurf, workout, snowboard, snowshoe
                               Type detected from file overrides, uses athlete's default type if not specified
+                              WARNING - This param is supported (as of 2022-11-15), but not
```
lwasser, line 686: @jsamoocha so is this not documented in the strava API docs?? or in our docs?  many thanks

stravalib/client.py
```diff
@@ -686,13 +683,24 @@ def upload_activity(self, activity_file, data_type, name=None, description=None,
...
+                              documented and may be removed in the future.
         :type activity_type: str
 
-        :param private: (optional) set to True to mark the resulting activity as private, 'view_private' permissions will be necessary to view the activity
+        :param private: (optional) set to True to mark the resulting activity as private,
+                        'view_private' permissions will be necessary to view the activity.
```
lwasser, line 691: similar question to above - if it's not supported in the strava API does that mean it just no longer works? so we should remove it and add a warning to the user that this particular API feature has been removed by strava?

jsamoocha:
> oh @jsamoocha i can ask some of my colleagues about what the best practices are for this. but question - are we dropping those API elements now? (i am asking to understand better how future warning works). Is future warning for an instance where we might drop this in the next release vs it no longer exists now?
> 
> Also it looks like this PR has some merge conflicts

In this PR, no API elements are dropped from stravalib.

I'm not in favor of removing deprecated request parameters as long as Strava doesn't send an error response because of them (I manually checked on the actual Strava API with a test account). Dropping them may break code that has dependencies on stravalib. So instead, by using [`warnings`](https://docs.python.org/3/library/warnings.html) (and logging them), users using the deprecated parameters will have time to address these issues while their code remains functioning.

I was mainly wondering about the best way to import/configure warnings. I did everything on the module level (i.e., in `client.py`), but maybe there are better ways?

stravalib/client.py
jsamoocha: `activity_type` doesn't officially exist anymore as a request parameter per the Strava API docs. However, it still seems to be functioning - I manually checked that providing this parameter makes strava assign the type to the uploaded activity.

stravalib/client.py
jsamoocha: `private` also doesn't exist anymore per the Strava API docs. But contrary to `activity_type`, it seems to be ignored by Strava: providing private=True/"true"/1/etc doesn't make the uploaded activity private.

yihong0618 (approved):
LGTM

hozn:
> or maybe @hozn has thoughts on how we handle deprecated api endpoints? it seems like we are dropping them now

If the API endpoints are simply deprecated, my general thought would be:
- Emit a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) from the method in question for the next point or even patch-level release.
- Remove the method in the next major release.  (Which may mean forcing cut of major release due to Strava dropping the method.)

PR #259, opened 2022-11-18
FEAT: setuptools_scm for versions, build for packaging/dist & remove setup.py & CI for pypi release

lwasser:
closes #251 

Ok y'all this one needs a solid review. it does the following 

✅ It adds a CI build that will run 
   * when we push to master (ie merge a PR) it will push to test pypi so we can see what the distro looks like and how it installs
   * When we create a new tagged release, it will push to the real pypi. i haven't tested a real push yet but am assuming test pypi is working so the real should work. 
✅ It moves away from old packagin standards of using setup.py and to the new preferred build package which removes the need for setup.py and supports pyproject.toml and setup.cfg 
✅ It updates the changelog for the past few changes made by @yihong0618 and @jsamoocha (just realized i need to add this change too). 
✅ it moves the test suite to run on the installed wheel vs the flat files in the package. this better represents the experience that the user will have : NOTE: it's actually possible that it's not running the tests on the installed package given we have the code in the main dir rather than src/dir and test are run in the main dir. but that can be another discussion
✅ it updates docs to build and grab the new version which i am grabbing from the scm generated version_generated file 

Please ask questions if any of this doesn't make sense. i had to spend some time understanding how SCM works on the back end and what the new packaging standards are. so I am still learning here too. but i have tested this.

If it all works, we can update docs, readme and bump a version. but i really would love some feedback on this before we merge :) it's a significant package restructure that I think brings us into modern python packaging space.

PR #263, opened 2022-11-20
FEAT: Cleanup make files and add docs autobuild

### LINKED ISSUE #261: Update build steps to use a better cleaned up Makefile(s)
Hi everyone.
Right now I created a really ugly makefile to build the docs that needs to be cleaned up.

Recently i've been using Nox a good bit and enjoying it as it's python based. But also it has some functionality including virtual environments built in. 

Is anyone against my setting up nox to do things like

make tests
make docs 
make clean 
?

instead it would be 

nox -s docs
nox -s docs-live (to have a live doc version that updates as you update the code
nox -s tests
etc

[here is a post on using nox](https://codewitholi.com/_posts/python-nox-automation/)

i think nox can really simplify a newer user getting started. but it does change how those of us who are used to certain tools work just a bit . 

UPDATE: i think we can do most of the same things with make and i have a local refresh of our make build too. Nox is just python and a bit more modern and cross platform compatible and supports virtual envs. 

but i'm open to either if y'all want to stay with vanilla make too.

lwasser:
closes #261 

So this is an update to the make build that 

1. cleans up the make builds as they were a bit messy especially the docs one
2. adds  sphinx autobuild which is really nice for editing docs as it allows you to run the docs live as you edit them and view your updates in realtime in the browser. this saves a ton of time. 
3. Adds a clean step to clean out files in your dir that are crteated via builds that you'd never want to be in version control 

What we don't get with make is the virtual enviornment setup which might make it easier for a new contributor BUT make does the job. 

Once this PR which is small and the other big one are merged I will submit a new pr to update the builds to call make to streamline everything. That way we have the commands to do things like run CI tests in only one place so if we add more CI rules we can just update the makefile and the entire build will update!

NOTE: right now i don't have make setup to build the local tests that require api authentication. but we can talk about that later. that is a bit more complex 

----
UPDATE: i just closed a really old PR to update the docs link in the readme. i just added it to this PR as it's a small fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,34 @@
-docs: docs/*.rst docs/conf.py docs/Makefile stravalib/*.py  ## generate html docs
-	# -fMeET -o docs/api stravalib stravalib/tests/ stravalib/tests/unit stravalib/tests/functional # omit tests dir from build
-	$(MAKE) -C docs clean
-	$(MAKE) -C docs html
-	$(MAKE) -C docs linkcheck
+TESTDIR=tmp-test-dir-stravalib
+PYTEST_ARGS=stravalib/tests/unit stravalib/tests/integration
+
+## I still need to add code cov to this build
+help:
+	@echo "Commands:"
+	@echo ""
+	@echo "  install   install package in editable mode"
+	@echo "  test      run the test suite (including doctests) and report coverage"
+	@echo "  build     build source and wheel distributions"
+	@echo "  clean     clean up build and generated files"
+	@echo ""
+
+
+build:
+	python -m build .
+
+install:
+	python -m pip install --no-deps -e .
+
+test:
+	# Create tmp dir to ensure that tests are run on the installed version of stravalib
+	mkdir -p $(TESTDIR)
+	cd $(TESTDIR); 
+	pytest $(PYTEST_ARGS) $(PROJECT)
+	## cp $(TESTDIR)/.coverage* .
+	rm -r $(TESTDIR)
+
+## Clean up all unneeded files and directories and things that shouldn't be under version control
+clean:
+	find . -name "*.pyc" -exec rm -v {} \;
+	find . -name "*.orig" -exec rm -v {} \;
+	find . -name ".coverage.*" -exec rm -v {} \;
+	rm -rvf build dist MANIFEST *.egg-info __pycache__ .coverage .cache .pytest_cache $(PROJECT)/_version_generated.py
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ help:
 
 
 build:
-	python -m build .
+	python -m build
 
 install:
 	python -m pip install --no-deps -e .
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,5 +1,34 @@
...
+	@echo ""
+
+
+build:
+	python -m build .
+
+install:
+	python -m pip install --no-deps -e .
+
+test:
```
lwasser, line 21: This runs the tests in a temp directory. It removes the temp directory after they are run. This is useful because we should be running tests on the installed package rather than the flat files as things can go wrong in the installed package that we won't see until they are on pypi.

docs/Makefile
```diff
@@ -13,166 +12,41 @@ $(error The '$(SPHINXBUILD)' command was not found. Make sure you have Sphinx in
...
+	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
-pseudoxml:
-	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
-	@echo
-	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+## Serve live docs in a browser
+serve:
+	pwd
+	sphinx-autobuild ../docs _build/html
```
lwasser, line 52: This serves the docs live in the browser and updates when you update any of the doc files.

lwasser:
I cant properly update the changelog until we merge #259 so once we merge that i'll add the changelog update to this pr.

jsamoocha (approved):
Ok, I'm not an expert on this. Do I understand correctly that the different makefiles (for testing/building/documentation/linting/etc.) are there to hide complexity from the workflow .yml files and keep the latter as small/simple as possible?

changelog.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## Unreleased
+
+* Update make to build and serve docs and also run current tests (@lwasser,#263)
```
lwasser, line 5:
```suggestion

* Add an option to mute Strava activity on update (@ollyajoke, #227)
* Update make to build and serve docs and also run current tests (@lwasser,#263)
```

changelog.md
```diff
@@ -1,6 +1,9 @@
 # Change Log
 
 ## Unreleased
+
+* Add an option to mute Strava activity on update (@ollyajoke, #227)
```
lwasser, line 5:
```suggestion
* Fix: add new attributes for bikes according to updates to Strava API to fix warning message (huaminghuangtw, #240)
* Add an option to mute Strava activity on update (@ollyajoke, #227)
```

changelog.md
```diff
@@ -1,6 +1,9 @@
 # Change Log
 
 ## Unreleased
+* Fix: add new attributes for bikes according to updates to Strava API to fix warning message (huaminghuangtw, #240)
```
lwasser, line 4:
```suggestion
* Fix: add new attributes for bikes according to updates to Strava API to fix warning message (huaminghuangtw, #239)
```

PR #265, opened 2022-11-21
FIX: pypi build with version & simplify build ✨ 

lwasser:
This PR:

Fixes the issue where it couldn't find the dist/ dir. essentially to build that way i'd need to upload the archive and then download it in the publish step. That just was getting a lot more complex than it needed to be. 

Thus i opted to simplify the build and just have it all run in one step with conditionals. it also should only run in the stravalib main repo so other contributors don't have to worry about it.


lwasser:
This is just a fix so i'm going to merge it.

PR #271, opened 2022-11-23
Fix: Add readthedocs config file to ensure that stravalib is installed in the build using pip

### LINKED ISSUE #270: FIX documentation build in both actions and RTD
[See this thread here. ](https://github.com/stravalib/stravalib/issues/267#issuecomment-1324147035)

Right now we are using setuptools_scm to keep track of versions. Our docs are being built automatically using readthedocs build engine. we are testing the docs build here. because

1. we watn to test for broken links (to be added)
2. the API is documneted with sphinx so if a doc string is updated incorrectly it would break the docs so we want to test
3. we could eventually setup doc test for in doc string vignettes like many packages do. 

Previously, the documentation was built on actions and pushed to gh-pages for direct hosting. The downside of this is that there is no GUI there for previous documentation versions.  See issue #222 for previous discussions

The build is breaking because the docs needs to access the _Version__ attribute of the package which is accessed via the _version_generated.py file that scm creates. so we need to build the package to access that file. 

I just need to 
1. fix the action build to accomodate the file and 
2. look at modifying the RTD build to make sure it too has access OR push directly from github to RTD if that is an option. 

i will look into both. RTD builds can be a bit slow.

lwasser:
closes #270 

in part but i think we should talk this through in that issue just to ensure we are all on the same page. 

This PR just adds a yml file with instructs readthedocs to build our package using pip. i think that with pip, the version_generated file must get created as long as you have tags locally the version should work.  

It also adds 3 badges to the readme:
* docs build or pass from readthedocs. that will make it easier to identify when a  build goes wrong
* python versions that we support
* current release version on pypi

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,7 +17,7 @@ Welcome to Stravalib!
 `Strava <https://www.strava.com>`_  API.
 
 This library is designed provide a simple and easy-to-use object model paradigm for interacting
-with the API, support modern versions of Python (2.7, 3.2+), and expose the full functionality of the REST API.
+with the API, support modern versions of Python (3.7+), and expose the full functionality of the REST API.
 
 **Why use stravalib?**  The Strava REST API is fairly straightforward.  The main reasons to use something like
 stravalib would be:
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -34,5 +34,5 @@ jobs:
       - name: Install dependencies
         run: |
           python3 -m pip install -r ./requirements.txt
-          python3 -m pip install -e .
-      - run: make docs -B
+          python3 -m pip install .
+      - run: make -C docs html
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,13 @@
+version: 2
+
+sphinx:
+  configuration: docs/conf.py
+
+formats: all
+
+python:
+  version: 3.8
+  install:
+    - method: pip
+      path: .
+    - requirements: requirements.txt
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ See the [online documentation](https://stravalib.readthedocs.io/) to learn more.
 
 The package is available on PyPI to be installed using `easy_install` or `Or you can
 
-
 ## How to Contribute to Stravalib
 
 ### Get Started!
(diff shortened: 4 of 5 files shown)


docs/index.rst
```diff
@@ -17,7 +17,7 @@ Welcome to Stravalib!
 `Strava <https://www.strava.com>`_  API.
 
 This library is designed provide a simple and easy-to-use object model paradigm for interacting
-with the API, support modern versions of Python (2.7, 3.2+), and expose the full functionality of the REST API.
+with the API, support modern versions of Python (3.7+), and expose the full functionality of the REST API.
```
lwasser, line 20: This is here so we can see the docs build did in fact update. i have a larger documentation update coming.

lwasser:
@jsamoocha you are absolutely correct, I was focused on readthedocs and not on our build and didn't understand what you were pointing out. The docs build was wrong. We have to call `make -C docs` to tell make to look in the docs directory for the make file to build docs and then we can use the flags there such as `make -C docs html` etc. SO i've updated both the readme and the build file. let's see how this runs. it should work now. 

I will submit another PR that clarifies how to build. if you install with pip you will get the version file required to installing in editable mode locally will work - you don't have to build the wheel unless you want to `make -C docs` locally as that will require the version file. i'll clarify with better documentation.

lwasser:
so this time the docs did indeed build properly in actions. goodness sorry i was misunderstanding your comments and not seeing the issue. have a look now and let me know what you think. and really good catch!

jsamoocha (approved):
LGTM, big changes like the last one are so super hard to get right in one go with many moving parts that can hardly be tested in isolation :-)

yihong0618 (approved):

lwasser:
@jsamoocha thank you!! so true! and i'm so so glad that you caught this because I was more focused on the package structure and also then readthedocs building. Thank you for the catch and the review. I'm merging this. 

Given also we have two approvals @yihong0618 and @jsamoocha always feel free to merge my PR's when you both approve :) and thank you both for being such a great maintainer team to work with!! ✨ !I'm enjoying working with y'all and @hozn !

lwasser:
In theory this should trigger a RTD build and the  badge on the readme for the RTD docs build should turn green. let's see.

PR #272, opened 2022-11-24
Fix:  deprecated set-output command 

yihong0618:
see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - fix_actions
+      - master
 
 jobs:
   build-doc:
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - fix_actions
 
 jobs:
   build-doc:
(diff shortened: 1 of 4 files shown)


jsamoocha (approved):
LGTM, thanks!

lwasser:
@yihong0618 this is so awesome thank you. Are you able to update the changelog.md file as well and then we can merge this? Many thanks.

Also as a learning opportunity for me there is one part that you added that I don't understand. if you could help me understand what it is doing, i'd appreciate it :) Thank you so  much!

.github/workflows/build-docs.yml
```diff
@@ -21,16 +21,17 @@ jobs:
         run: |
           # install pip=>20.1 to use "pip cache dir"
           python3 -m pip install --upgrade pip
-      - name: Get pip cache dir
-        id: pip-cache
-        run: echo "::set-output name=dir::$(pip cache dir)"
+      - name: Set Variables
+        id: set_variables
+        shell: bash
+        run: |
```
lwasser, line 27: So can you help me understand what this run step is doing @yihong0618 no critique here i'm just trying to understand the fix! i see that we need to implement the fix from that blog you linked to!!

yihong0618: image

different python version + executable generate different hashid as key

lwasser: thanks!

PR #277, opened 2022-11-26
FIX: update documentation to use cleaner furo theme and add markdown support via myst

lwasser:
closes #276 
This pr does a bunch of cleanup things

* It adds support for myst which allows us to include markdown in our docs for easier contributing! 
* It cleans up the landing page of the docs to make it easier for users to find content
* It embeds the CONTRIBUTING.md file into the documentation for easier discovery
* It just cleans up extra stuff from the conf.py file that we don't need! 
* I finessed the documentation styles so its just a bit cleaner as well :) 

In case you don't want to build locally i've added some screenshots of what the new theme and landing page look like! 

# Documentation landing page
(I might make the grid cards grey or add more of a drop shadow at some point) 

Screen Shot 2022-11-26 at 11 37 22 AM

## Landing page lower down 
Screen Shot 2022-11-26 at 11 37 29 AM

## New sidebar navigation 

so nav is not split between top and side which i think is confusing .
Screen Shot 2022-11-26 at 11 37 11 AM

PR #278, opened 2022-11-26
FIX: Add development guide and more contributing resources

lwasser:
This pr (in draft as it includes elements from #277 so it will be smaller once that is merged):

* Creates a development guide (which needs a good review) that overviews our workflow here and infrastructure that we've created. It just occurred to me that I should add a release cycle page as well to overview how we do releases. I will do that. 
* Removes the old contributing.rst file
* adds just a small tutorial link for the authentication page because that will help users get started creating a strava app. 
* Cleans up the get-started intro page (drop easyinstall and adds cards)

Screen Shot 2022-11-26 at 12 16 38 PM


lwasser:
i am closing this. i don't like how the rebase went. will open a new PR once i've tested locally.

Issue #282, opened 2022-11-30
Missing implementation for /segments/{id}/starred (PUT) endpoint


jsamoocha:
There's no implementation for this endpoint in `client`.

lwasser:
@jsamoocha i see this issue was called in a few other issues and pr's. but i don't see a put method for starred segments but [do see this in the docs](https://developers.strava.com/docs/reference/#api-Segments-starSegment)
 which suggests we could potentiall star or unstar segments (if i'm reading that right)?

is this something that is still on our TODO list or can we close this? if this is supported by strava and there is some intent to implement it i can create a milestone just to remind us it is an active potential new feature. just lemme know!

jsamoocha:
This is still open as we currently don't support this API endpoint.

Issue #222, opened 2021-09-16
Update Documentation Infrastructure for Stravalib


lwasser:
In issue #219 , @samryan18 points out that the documentation is out of date. This is likely because the python hosted service doesn't seem to be operating as it used to operate currently. The docs here are already using sphinx which makes it super easy to port to a new platform quickly! We have two options. 

## [Read the Docs](https://readthedocs.org/): 

This platform will automagically connect to a github repo with sphinx setup and allow you to build documentation. it supports versioning, custom environments and branch development of documentation if needed. 

### Potential Drawbacks:
   * maintainers need to setup an account (you can use github to login) 
   *  if docs are failing sometimes you will have to fix things within read the docs (this is rare but i have to do it maybe once or twice a year for our packages). 
   * Advertisements on the docs - see examples below:
[Earthpy docs using standards sphinx theme](https://earthpy.readthedocs.io/en/latest/) 
[Geopandas docs using read the docs (pydata theme)](https://geopandas.readthedocs.io/en/latest/docs.html )
[GEOPANDAS docs running on gh pages - note the lack of versions at the bottom but also no ads: ](https://geopandas.org/index.html)
   * i've run into memory issues with some dev environments using RTD especially if you use conda

What i like to do with this build is create a local github action that tests the documentation build. often you can see the point of error and correct it in github without even having to touch read the docs. every once in a while you need to go into the read the docs admin interface. the new url for docs would be: stravalib.readthedocs.io 

### Benefits
* Doc build maintained by RTD platform
* All versions of your package documentation including latest available to users
* will be super fast to get up and running

##  2. Github Pages Build
In this scenario we'd build the docs via github actions and push them to a new branch here called gh-pages. We'd turn on github pages on the repo and the new docs url will be stravalib.github.io/ . This could work as well but the onus would be on stravalib maintainers to maintain the sphinx website. it would be a bit more infrastructure development here as well to maintain. This solution would also not provide versioned docs so it would always be "latest". Read the docs keeps that history online which is nice. On the flip side in this scenario no one needs a separate set of permissions to work on doc build issues. So in the long term it could be easier if you want more people to be able to help update builds, docs etc.

### Drawbacks
  * A bit more work up front to setup CI infrastructure for the build (but probably worth it)
  * No versioned built docs available to the user online
 
### Benefits
  * No need for an additional login as it all lives here in github / github pages
  * Easier for anyone to update doc builds and such without need special permissions to authenticate with RTD
[...]

PR #283, opened 2022-11-30
Add integration tests for all POST/PUT client methods

jsamoocha:
These new tests make sure that all (implemented, see #282) endpoints used for creating and updating objects are covered.

It adds several deprecation warnings (both at runtime and in the documentation) for method parameters that are not supported anymore by Strava.

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -309,9 +309,25 @@ class Client(object):
         https://developers.strava.com/docs/reference/#api-Athletes-updateLoggedInAthlete
 
         :param city: City the athlete lives in
+
+                     .. deprecated:: 1.0
+                     This param is not supported by the Strava API and may be
+                     removed in the future.
         :param state: State the athlete lives in
+
+                      .. deprecated:: 1.0
+                      This param is not supported by the Strava API and may be
+                      removed in the future.
         :param country: Country the athlete lives in
+
+                        .. deprecated:: 1.0
+                        This param is not supported by the Strava API and may be
+                        removed in the future.
         :param sex: Sex of the athlete
+
+                    .. deprecated:: 1.0
+                    This param is not supported by the Strava API and may be
+                    removed in the future.
         :param weight: Weight of the athlete in kg (float)
 
         :return: The updated athlete
@@ -636,14 +652,18 @@ class Client(object):
                               iceskate, inlineskate, kitesurf, rollerski,
                               windsurf, workout, snowboard, snowshoe
         :param private: Whether the activity is private.
-                        WARNING - This param is not supported by the Strava API and may be
+
+                        .. deprecated:: 1.0
+                        This param is not supported by the Strava API and may be
                         removed in the future.
         :param commute: Whether the activity is a commute.
         :param trainer: Whether this is a trainer activity.
         :param gear_id: Alpha-numeric ID of gear (bike, shoes) used on this activity.
         :param description: Description for the activity.
         :param device_name: Device name for the activity
-                            WARNING - This param is not supported by the Strava API and may be
+
+                            .. deprecated:: 1.0
+                            This param is not supported by the Strava API and may be
                             removed in the future.
         :param hide_from_home: Whether the activity is muted (hidden from Home and Club feeds).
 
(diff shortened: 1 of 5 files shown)


stravalib/client.py
```diff
@@ -291,9 +309,25 @@ def update_athlete(self, city=None, state=None, country=None, sex=None, weight=N
         https://developers.strava.com/docs/reference/#api-Athletes-updateLoggedInAthlete
 
         :param city: City the athlete lives in
```
lwasser, line 311: @jsamoocha are you open to migrating to numpy style docstrings? it seems to be a standard now and i think it's much easier to read. i started to implement this [back in this issue](https://github.com/stravalib/stravalib/issues/222#issuecomment-924501383) and had a local branch but that branch is super dated now . if you are open to it, when the time comes i'm happy to help with a bigger pr. i plan to do that with black anyway. But i wanted to run it by you first. i can create a specific issue about this.

there is a nice tool that does the conversion for you that I was using. and we don't have to implement it in this PR it's just something i've been thinking about.

jsamoocha: Sure, especially if there's a tool out there that will save us from doing it all manually!

lwasser: there is! we can work on that in another PR. i found this great command line tool that automagically converts docstring format :) yay!

stravalib/client.py
```diff
@@ -29,6 +29,24 @@
 logging.captureWarnings(True)
 
 
+def warn_param_deprecation(param_name: str):
```
lwasser, line 32: @jsamoocha this is great to have functions that handle our warnings. Do you think they belong in the exc.py module rather than the client module as helpers?

stravalib/client.py
```diff
@@ -715,38 +764,28 @@ def upload_activity(self, activity_file, data_type, name=None, description=None,
...
+        if data_type not in valid_data_types:
+            raise ValueError(f'Invalid data type {data_type}. Possible values {valid_data_types!r}')
 
         params = {'data_type': data_type}
         if name is not None:
             params['name'] = name
         if description is not None:
             params['description'] = description
         if activity_type is not None:
-            warnings.warn(
```
lwasser, line 727: just curious is the activity_type no longer undocumented?

stravalib/client.py
```diff
@@ -715,38 +764,28 @@ def upload_activity(self, activity_file, data_type, name=None, description=None,
...
-                stacklevel=2
-            )
             if not activity_type.lower() in [t.lower() for t in model.Activity.TYPES]:
-                raise ValueError("Invalid activity type: {0}.  Possible values: {1!r}".format(activity_type, model.Activity.TYPES))
-            params['activity_type'] = activity_type
+                raise ValueError(f'Invalid activity type: {activity_type}. Possible values: {model.Activity.TYPES!r}')
+            warn_param_unofficial('activity_type')
+            params['activity_type'] = activity_type.lower()
         if private is not None:
-            warnings.warn(
```
lwasser, line 737: same question here. just a general question so i understand!

stravalib/unithelper.py
```diff
@@ -42,27 +42,3 @@ def c2f(celsius):
 
     """
     return (9.0 / 5.0) * celsius + 32
-
-
-def timedelta_to_seconds(td):
```
lwasser, line 47: @jsamoocha this pr is really helping me understand how you created the tests and the mock object. super clever! 

I thijnk it's great to remove this as it's built in but we should update the docs with unit conversions if they don't exist. so i'm going to open an issue about that right now. i know we have a ton of of issues but we can still keep chipping away!

lwasser (approved):
@jsamoocha this can be merged - i think it looks great. the only question i had was about whether we want those warning helpers to live in the exc.py file with other warning related classes (i have questions actually about that module too but will save that for an issue :) 

as it's not related to this pr.

lwasser:
oh i think i'll also work next on adding codecov to this repo! it's about time we add it. :) thank you for creating such an awesome test suite and for being so clever with the design of it!!

stravalib/client.py
jsamoocha: No, I moved the warning after the check for valid activity types. That ValueError should be raised anyway in case of an invalid type. This move allowed me to reduce complexity of writing the tests (because I only need to check if either an exception or warning is raised, but not both).

stravalib/client.py
jsamoocha: It's still deprecated, but I moved the actual warning into a separate function.

stravalib/client.py
jsamoocha: Ah, yes - good suggestion! The `exc` module makes more sense. I moved them there.

PR #296, opened 2022-12-13
Initial pass - blue

lwasser:
hey there @jsamoocha i am working on the model file (format / docstrings etc) and i am finding the code to be complex. i wanted to point out a few things so we could have a discussion about this -


stravalib/model.py
```diff
@@ -11,11 +11,20 @@
...
-from stravalib.attributes import (META, SUMMARY, DETAILED, Attribute,
-                                  TimestampAttribute, LocationAttribute,
-                                  EntityCollection, EntityAttribute,
-                                  TimeIntervalAttribute, TimezoneAttribute,
-                                  DateAttribute, ChoicesAttribute)
+from stravalib.attributes import (
+    META,
+    SUMMARY,
+    DETAILED,
+    Attribute,
```
lwasser, line 18: we have a class called `Attribute`. because attribute is a term for an element  in a Python class it confused me. initially i thought incorrectly that it was an actual attribute called in a way that i didn't understand but afterin talking with friends i am clear it's a class. it's a class called within another class over and over. 

Should we consider naming conventions that do not use existing Python terms?

stravalib/model.py
```diff
@@ -81,40 +116,60 @@ def __repr__(self):
...
+    )  #: The detail-level for this entity.
 
 
 class IdentifiableEntity(ResourceStateEntity):
     """
     Mixin for entities that include an ID attribute.
     """
-    id = Attribute(int, (META, SUMMARY, DETAILED))  #: The numeric ID for this entity.
+
+    id = Attribute(
```
lwasser, line 145: I am really confused here (maybe i am just confused but..) here we use ATTRIBUTE to create a class within another class identifiableEntity and then id= is used.

But [id is also a python function](https://docs.python.org/3/library/functions.html#id) 

so i'm just really getting confused by the nesting of elements in this module.

wondering if @hozn could help me understand. i know that there is some complexity in this package that I need to better understand but also i wonder if some of it could be simpler which might make maintenance easier? 

many thanks for any input @jsamoocha @hozn

stravalib/model.py
```diff
@@ -81,40 +116,60 @@ def __repr__(self):
...
 
 class BoundEntity(BaseEntity):
     """
-    Base class for entities that support lazy loading additional data using a bound client.
+    Base class for entities that support lazy loading additional data using a 
+    bound client.
+
+    Attributes
+    ----------
+    bind_client : stravalib.client.Client
```
lwasser, line 157: So here  i suspect that bind_client is also a class because it's a Client object. 
Soi need to remove it from being an attribute. i'm just really getting confused by the complexity of the various objects called within objects. in some moduleslike exc we have classes that are this simpe

```
class LoginFailed(AuthError):
    pass
```

Screen Shot 2022-12-13 at 8 44 07 AM

lwasser:
[this video is related to my comments from pycon 2012 about classes and extra complexity (sometimes they are great but i just wonder here if we can simplify.](https://www.youtube.com/watch?v=o9pEzgHorH0)

docs/conf.py
```diff
@@ -89,7 +102,7 @@
 # The title in the left hand corner of the docs
 html_title = "Stravalib Docs"
 # Theme and css
-html_theme = "furo"
+html_theme = "pydata_sphinx_theme"
```
lwasser, line 105: This was just a temporary switch from furo to see if the theme rendered api docs differently or if it was in fact settings or my changes to the cod docs

stravalib/model.py
jsamoocha: Are you familiar with Python's descriptor protocol? All these `Attribute` objects within model objects are descriptors.

I agree that the naming may be confusing. Whenever we change to Pydantic (and generated model classes), all this logic within the `attributes` module will go (at least all the (un-) marshaling and typing).

stravalib/model.py
jsamoocha: As I mentioned above, switching to Pydantic will remove much of this complexity.

stravalib/model.py
lwasser: thank you @jsamoocha i think i need to learn a bit moer about descriptor protocols. I first looked at the code quickly and thought perhaps it was an actual attribute. but then as the docs were rendering i realized (with input from a friend) that it was actually a class called Attribute. but it sounds like i need to read about descriptor protocols.

stravalib/model.py
lwasser: perfect.

lwasser:
ok @jsamoocha i will close this. i appreciate your feedback

so given the move to pydantic, should i hold off on a format with black, isort docstring cleanup? if we are doing a big overhaul of the API via pydantic it seems like it might be a better use of my time to work on tutorials and workflows which will both test the API as i run them but also provide users with more guidance on using stravalib.

If we setup [sphinx gallery](https://sphinx-gallery.github.io/stable/index.html) the code in my tutorials can actually run. i can include a small api return dataset in the package eventually to support the tutorials being run in CI. the only thing we can't test that way in a workflow is API elements. which i could include secrets in the repo for that but not sure it's worth it at this point

NOTE: i haven't tested gallery with this theme yet but could. The first iteration could just be straight forward tutorials.

lwasser:
thanks @jsamoocha this is really helpful - https://docs.python.org/3/howto/descriptor.html i understand better what the intent of classes within classes is! i don't have a CS background so am learning some new things here which is great!

jsamoocha:
> so given the move to pydantic, should i hold off on a format with black, isort docstring cleanup? 

See [this](https://koxudaxi.github.io/datamodel-code-generator/formatting/): it seems that using a (generated) pydantic model doesn't conflict with your efforts to introduce (automatic) formatting and import sorting etc.

About tutorials (and adding more tooling to support these): What would you like to add?

PR #297, opened 2022-12-13
Units2pint

### LINKED ISSUE #281: Units dependency seems to be abandoned
The `units` package seems not to be maintained any longer (the latest release on PyPi is from 2017, and its source repo has vanished).

In line with #248, we need to start thinking about an alternative, possibly from [this list](https://socialcompare.com/en/comparison/python-units-quantities-packages).

jsamoocha:
closes #281

## Description

Replaces the dependency on `units` with the `pint` library since the former is no longer maintained. Adds helper classes and functions to maintain backward compatibility. Unless client code uses internals of the `units.Quantity` type, the result of this change should be compatible with existing code.

## Type of change

- [x] Other (please describe)

The `units` package is no longer maintained and may lead to dependency resolution errors (e.g., in pipenv 2022.11.30).

## Does your PR include tests

- [x] Yes

### DIFF
--- a/stravalib/exc.py
+++ b/stravalib/exc.py
@@ -113,3 +113,13 @@ def warn_param_unofficial(param_name: str):
         FutureWarning,
         stacklevel=3
     )
+
+
+def warn_units_deprecated():
+    warnings.warn(
+        'You are using a Quantity object or attributes from the units library, which is '
+        'deprecated. Support for these types will be removed in the future. Instead, '
+        'please use Quantity objects from the Pint package (https://pint.readthedocs.io).',
+        DeprecationWarning,
+        stacklevel=3
+    )
--- a/stravalib/tests/unit/test_unithelper.py
+++ b/stravalib/tests/unit/test_unithelper.py
@@ -1,0 +1,56 @@
+from dataclasses import dataclass
+
+import pint
+import pytest
+from pint import Unit
+
+from stravalib import unithelper as uh
+from stravalib.unithelper import UnitsQuantity
+
+
+@dataclass
+class UnitsLikeQuantity(UnitsQuantity):
+    num: float
+    unit: str
+
+
+@pytest.mark.parametrize(
+    'from_unit,to_unit,expected_magnitude,expected_unit',
+    (
+        (None, 'meter', 1, 'meter'),
+        ('meter', 'meter', 1, 'meter'),
+        ('km', 'meter', 1000, 'meter'),
+        ('m/s', 'kph', 3.6, 'kilometer / hour')
+    )
+)
+class TestUnitConversion:
+    def test_conversion_legacy(self, from_unit, to_unit, expected_magnitude, expected_unit):
+        quantity = UnitsLikeQuantity(1, from_unit) if from_unit else 1
+        with pytest.warns(DeprecationWarning):
+            converted_quantity = getattr(uh, to_unit)(quantity)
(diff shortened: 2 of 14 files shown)


jsamoocha:
Ok, python 3.7 build is failing because it doesn't have `typing.Protocol`. Pint also doesn't (officially) support python 3.7. Maybe it's time to drop support for 3.7?

stravalib/exc.py
```diff
@@ -113,3 +113,13 @@ def warn_param_unofficial(param_name: str):
         FutureWarning,
         stacklevel=3
     )
+
+
+def warn_units_deprecated():
```
jsamoocha, line 118: A warning to discourage users from using the legacy `units` types and attributes so we can eventually get rid of the monstrosities in the `unithelper` module.

stravalib/tests/unit/test_unithelper.py
```diff
@@ -0,0 +1,56 @@
...
+    'from_unit,to_unit,expected_magnitude,expected_unit',
+    (
+        (None, 'meter', 1, 'meter'),
+        ('meter', 'meter', 1, 'meter'),
+        ('km', 'meter', 1000, 'meter'),
+        ('m/s', 'kph', 3.6, 'kilometer / hour')
+    )
+)
+class TestUnitConversion:
+    def test_conversion_legacy(self, from_unit, to_unit, expected_magnitude, expected_unit):
```
jsamoocha, line 27: For testing backward compatibility

stravalib/unithelper.py
```diff
@@ -2,28 +2,102 @@
...
+        return getattr(self.q, item)
+
+    def __str__(self):
+        return str(self.q)
+
+    def __repr__(self):
+        return repr(self.q)
+
+
+class UnitConverter:
```
jsamoocha, line 60: This class ensures we have a callable that behaves as the legacy `units`-based converters. It wraps the converted `pint.Quantity` object such that it can be interacted with as if it was a `units.Quantity` object.

lwasser:
> Ok, python 3.7 build is failing because it doesn't have `typing.Protocol`. Pint also doesn't (officially) support python 3.7. Maybe it's time to drop support for 3.7?

Yes to add pint we definitely want to drop 3.7 support!! otherwise we could run into conflicts later - eeks. i suggest adding 3.11 and removing 3.7 which seems to be in line with what other packages are supporting.

stravalib/exc.py
lwasser: 😆 thanks for the laugh. cya monstrosities!

lwasser:
you rock THANK YOU for pulling this together so quickly. can test locally tonight or tomorrow.

lwasser:
ok @jsamoocha i'm running into a few issues locally not related to this PR. i'm opening issues as i see them.

This works great on an activity that i pulled from strava:

```python
print(unithelper.miles(an_activity.distance))
print(unithelper.kilometers(an_activity.distance))
```

All of this seems to work. 

Are there other things that I should test here? i can't break things locally so i think we can approve but with the caveat that we move to python 3.11 and drop 3.7.

lwasser:
I made a small change to hopefully support docs building. the dev guide link in the CONTRIBUTING was incorrect. Not sure how i missed that in other pr's but let's see if all builds ✅

lwasser (approved):
All is green now. ready to merge whenever you are ready for it @jsamoocha !
Great contribution. I hope this fixes our wheel / dependency issue now!

PR #300, opened 2022-12-15
Release 1.1.0

jsamoocha:
## Release checklist
- [x] Be sure to clearly specify what version you are bumping to in the PR title: Example: Bump to version x.x
- [x] Add the version of this release to our changelog
- [x] Organize items changes under the new version in groups as follows: Added, Fixed and Changed
- [x] Add all contributors to the release below those sections 
- [x] Wait for a maintainer to approve the pull request. Then merge! You are now ready to create the release. 
- [x] Successfully installed from test pypi and manually tested athlete/activity fetching.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,18 +1,31 @@
 # Change Log
 
 ## Unreleased
+
+## v1.1.0
+
+### Added
 * Add: Development & build/release guide to documentation, edit button to documentation theme, pr template for release (@lwasser, #289)
 * Add: Integration tests for /routes/{id} and /segments/starred (GET) (@jsamoocha, #250 (partial))
-* Fix: Move docs to `furo` theme, add `myst` support for markdown, include CONTRIBUTING.md in documentation, enhance intro documentation page and add linkcheck to docs build (@lwasser, #276)
 * Add: Add integration tests for all POST/PUT client methods (@jsamoocha, #250 (partial))
 * Add: code cov to test suite (@lwasser, #262)
 * Add: add code of conduct to the repo, update contributing guide + readme badges (@lwasser, #269, #274) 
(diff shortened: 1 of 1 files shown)


lwasser (approved):
Nice! tested from testpypi locally and all looks great! Nice job with dropping that units dependency so quickly! 

Approving and ready to merge!

@jsamoocha we haven't talked about this but what are your thoughts on squashing merge commits vs leaving. I noticed a merge commit in this PR. 

Some really like to not squash only because it helps contributors if people look at the contributor insights - the more commits contributes to your "stats". For me, i like having a cleaner commit history to troubleshoot when you have errors. so in this PR i would squash and merge and rename the commit to release x.x 

are you comfortable with squashin g this pr in this case? i've done it previously I just didn't think to ask you what you thought about it. if there is a PR with really descriptive commits i think it's ok to not squash. but this one i think it's worth squashing. thoughts? 

so i'd prefer to see a commit for this one that says RELEASE: 1.1 or something like that.

jsamoocha:
You're right. The merge commit was not informative, so I squashed it.

I generally like to keep commits intact for documentation reasons (unless the commit history becomes polluted with many edits/fixes). However, when these are squashed and force-pushed during PR review, Github sometimes loses track of what has changed since the previous review iteration. So I guess the default opinion holds: "it depends" ¯\_(ツ)_/¯

lwasser:
sounds great! just wanted to check in. i'll try to do the same and can clean up as i go to make individual commits more informative as well rather than squashing everything!

lwasser:
AFAIKT everything looks good. it released to pypi. seems like our release workflow is working well! 🎆

jsamoocha:
Yes, the release doc you created made it really easy 👍

Issue #248, opened 2022-11-09
Discussion: Consider Pydantic as a replacement for homegrown Models / Attributes framework ?


hozn:
A key value-add, in my opinion, for Stravalib is the modeling of the entities from Strava in a way that makes them easy for API users to integrate with their applications.  Not just the stronger typing that is increasingly helpful in larger Python applications, but also handling speed/distance units and conversions.   Since this library has been written, we've had some great new packages added to the python ecosystem.  We've also had newer python introduce native dataclasses.  For awhile [Marshmallow](https://marshmallow.readthedocs.io/en/stable/) was something I'd been looking at to change how we were modeling the Strava entities; however, I think the clear front-runner in recent years is [Pydantic](https://pydantic-docs.helpmanual.io/). 

I think the main benefit to this would be embracing what is becoming (or feels like it is becoming) the standard for modeling (de)serializable data classes in python.  Pydantic is easily customized to handle more complex deserialization (or serialization) of these dataclass models.

More thought would need to be given to how to handle units & type conversions, the current way of expressing the different detail levels and which attributes are present, etc.  (The latter may not be strictly necessary either?)

I haven't fully thought this through or proven it out; this is more just an observation that stravalib rolls its own (de)serialization & modeling framework, which may be better handled by purpose built tooling in 2022.  So just wanted to propose this idea for consideration if there is a plan to do a major version bump & break backwards compatibility.  (This would also mean supporting only newer Python 3, but I think that is probably acceptable with a major version bump.)

lwasser:
@hozn i always think that if there is a really well maintained tool that implements functionality that we'd be maintaining manually here, it's worth considering adding it as a dependency and refactoring code. It seems like we're already moving away from python 2.x and i think we are at >=3.7 at this point given the CI changes implemented already! 

So i suppose the question is, how much work would this entail? We also should probably get all of our infrastructure, tests, etc updated first as well :)

jsamoocha:
[...]

PR #302, opened 2022-12-30
Update model

jsamoocha:
## Description

This PR adds a workflow that periodically or manually generates a Pydantic model from the published Strava API. When this generated model differs from the previous version (i.e., the API has changed), a pull request is generated for merging the changes. This addresses one of the tasks in #248 .

The added workflow uses a custom GitHub Action that I recently added to the stravalib organization and can be found [here](https://github.com/stravalib/strava_swagger2pydantic)

## Type of change

This particular PR has no effect on existing functionality as the generated/updated model is not used (yet)

## Does your PR include tests

No, the only changes are in workflow - and configuration files

NA

### DIFF
--- a/.github/workflows/check-strava-api.yml
+++ b/.github/workflows/check-strava-api.yml
@@ -1,0 +1,29 @@
+name: Check Strava API
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  update-model:
+    name: Update Model
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Fetch Latest API
+        uses: stravalib/strava_swagger2pydantic@v1
+        with:
+          model_file: 'stravalib/strava_model.py'
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          add-paths: |
+            stravalib
+          commit-message: Strava API Change
+          branch: api-change
+          delete-branch: true
+          title: '[CHANGE] Strava API Change'
+          body: |
+            There were changes in the Strava API:
+            [Please edit this comment to indicate what has changed]
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ and quantities through the [Python Pint library](https://pypi.org/project/Pint/)
 
 * Python 3.8+
 * Setuptools for installing dependencies
-* Other Python libraries (installed automatically when using pip/easy_install): requests, pytz, pint, arrow, pydantic
+* Other Python libraries (installed automatically when using pip): requests, pytz, pint, arrow, pydantic
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ and quantities through the [Python Pint library](https://pypi.org/project/Pint/)
 
 * Python 3.8+
 * Setuptools for installing dependencies
-* Other Python libraries (installed automatically when using pip/easy_install): requests, pytz, pint, arrow
+* Other Python libraries (installed automatically when using pip/easy_install): requests, pytz, pint, arrow, pydantic
 
 ## Installation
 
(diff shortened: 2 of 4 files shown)


.github/workflows/check-strava-api.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: stravalib/strava_swagger2pydantic@v1
+        with:
+          model_file: 'stravalib/strava_model.py'
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          add-paths: |
+            stravalib
+          commit-message: Strava API Change
+          branch: api-change
```
lwasser, line 24: @jsamoocha can you help me understand the workflow here. i see a pydantic model branch but this will pull against api-change

README.md
```diff
@@ -15,7 +15,7 @@ and quantities through the [Python Pint library](https://pypi.org/project/Pint/)
 
 * Python 3.8+
 * Setuptools for installing dependencies
-* Other Python libraries (installed automatically when using pip/easy_install): requests, pytz, pint, arrow
+* Other Python libraries (installed automatically when using pip/easy_install): requests, pytz, pint, arrow, pydantic
```
lwasser, line 18:
```suggestion
* Other Python libraries (installed automatically when using pip): requests, pytz, pint, arrow, pydantic
```

lwasser: good to get rid of easy install as its [deprecated](https://packaging.python.org/en/latest/discussions/pip-vs-easy-install/)!

lwasser (commented):
just a few questions. specifically i'm confused about the branch workflow that is being setup here.

.github/workflows/check-strava-api.yml
jsamoocha: This workflow has three steps:

1. Check out the stravalib repo
2. Generate the Pydantic model from Strava's published API, which _may_ update the file `stravalib/strava_model.py`. This module contains the classes of the Pydantic model.
3. Create a temporary branch `api_change`, and in case of any git diffs in the path `stravalib`, create a PR with the provided commit message, title, etc.

So the `branch` key is just a parameter for the [create-pull-request action](https://github.com/marketplace/actions/create-pull-request). I agree the authors of that action could've prevented some confusion by naming that key `tmp-branch-name` or whatever else.

This workflow will run (AFAIK) only on the master branch when scheduled and can run on any branch when triggered manually.

I'm new to GitHub Actions and Workflows, so please feel free to ask more in case I don't make sense.

.github/workflows/check-strava-api.yml
```diff
@@ -0,0 +1,29 @@
...
+  workflow_dispatch:
+
+jobs:
+  update-model:
+    name: Update Model
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Fetch Latest API
+        uses: stravalib/strava_swagger2pydantic@v1
```
lwasser, line 15: Ok wow - you made a new action to do this. this is what i didn't understand. ive never made an action before. i will have to clone that repo to see what the code actually does when it runs.

.github/workflows/check-strava-api.yml
lwasser: ok i think i follow now! let's merge as it's not going to modify our main code base at this point yet. 

i will play with the script that you wrote to actually create the API when i have a bit more time - more for my learning than as a test :)

lwasser (approved):
Let's merge this @jsamoocha thank you for all of the work here. i see where it's going. i'm going to try to carve out some time to play with the action that you created in the next week. i can also update my PR with the new dep once this is merged

.github/workflows/check-strava-api.yml
jsamoocha: I tested it on a private toy repo. I just gave you access to it to save some boilerplate code/time. Feel free to modify whatever you want for testing/learning purposes.

PR #309, opened 2023-01-06
fix(upload_activity):  - fix unable to read activity_file with type bytes

### LINKED ISSUE #308: unable to upload_activity if its bytes type
When calling `client.upload_activity` and `activity_file` type is `bytes`, 
I get this error 
`Invalid type specified for activity_file: bytes`

gitexel:
closes #308

## Description
When calling `client.upload_activity` and `activity_file` type is `bytes`,
I get this error
`Invalid type specified for activity_file: bytes`

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -734,7 +734,7 @@ class Client(object):
         if not hasattr(activity_file, 'read'):
             if isinstance(activity_file, str):
                 activity_file = BytesIO(activity_file.encode('utf-8'))
-            elif isinstance(activity_file, str):
+            elif isinstance(activity_file, bytes):
                 activity_file = BytesIO(activity_file)
             else:
                 raise TypeError("Invalid type specified for activity_file: {0}".format(type(activity_file)))
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -694,7 +694,7 @@ class Client(object):
         https://developers.strava.com/docs/reference/#api-Uploads-createUpload
 
         :param activity_file: The file object to upload or file contents.
-        :type activity_file: file or str
+        :type activity_file: TextIOWrapper, str or bytes
 
         :param data_type: File format for upload. Possible values: fit, fit.gz, tcx, tcx.gz, gpx, gpx.gz
         :type data_type: str
--- a/stravalib/tests/unit/test_client_utils.py
+++ b/stravalib/tests/unit/test_client_utils.py
@@ -1,5 +1,8 @@
+import os
+from unittest import mock
+
 from stravalib.client import Client
-from stravalib.tests import TestBase
+from stravalib.tests import TestBase, RESOURCES_DIR
 import datetime
 import pytz
 from urllib import parse as urlparse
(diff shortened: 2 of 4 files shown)


stravalib/client.py
```diff
@@ -734,7 +734,7 @@ def upload_activity(self, activity_file, data_type, name=None, description=None,
         if not hasattr(activity_file, 'read'):
             if isinstance(activity_file, str):
                 activity_file = BytesIO(activity_file.encode('utf-8'))
-            elif isinstance(activity_file, str):
+            elif isinstance(activity_file, bytes):
```
jsamoocha, line 737: Can you please also update the docstring of this function, so the parameter has the correct types `TextIOWrapper`, `str`, and `bytes`?

stravalib/tests/unit/test_client_utils.py
```diff
@@ -54,3 +57,34 @@ def test_params(self):
...
+
+class TestClientUploadActivity(TestBase):
+    client = Client()
+
+    def test_upload_activity_file_with_different_types(self):
+        """
+        Test uploading an activity.
+
+        NOTE: This requires clearing out the uploaded activities from configured
+        writable Strava acct.
```
jsamoocha, line 70: Is this statement true, given that `ApiV3.post` is mocked? I guess the Strava API is not hit?

jsamoocha (commented):
Hi, many thanks for your contribution! It seems that `bytes` was never supported.

Pending the comments, this looks good to merge.

Can you please also update the changelog with your contribution under the `##Unreleased` heading?

@lwasser , do you know why pre-commit.ci already tries to process this PR while we only merged its config file into that other feature branch?

stravalib/client.py
lwasser: @jsamoocha @gitexel What is an example of a file type format that would be read as bytes? (just for my understanding) are gpx and tcx read in that format? many thanks

lwasser:
hey @gitexel thank you for this pr!

@jsamoocha can you show me where ou see the bot tried to commit here? i only see two commits and then it errored. it error because the config file is NOT here yet. But the bot it setup to run on the repository - it's a github based installation on the entire repo so it needs the config file to run

Screen Shot 2023-01-06 at 1 24 28 PM

if you see something different please say the word! we probably should just add the config here as well. i don't know that there is a way to customize the bot to ONLY run on specific branches without a config file but i'll look into that if you want to hold off on running code format pr's here.

stravalib/tests/unit/test_client_utils.py
gitexel: I forgot to remove the statement. I will remove it.

lwasser:
i might be able to set it up here to 

1. NOT fix pr's ?
2. OR skip all of the ci checks - https://pre-commit.ci/#configuration-skip  
but maybe we just should add it here and we could turn off the auto pr feature and run it manually on pr's before they are merged (if we are worried about it) i am not worried about it at this point!

stravalib/client.py
gitexel: @jsamoocha done

gitexel: @lwasser when you generate or download a file with `.fit` format it will be `bytes` type, so it will be much easier to pass the bytes directly to `upload_acitvity`

gitexel: @lwasser when you generate or download a file with `.fit` format it will be `bytes` type, so it will be much easier to pass the bytes directly to `upload_acitvity`

jsamoocha (approved):
Thanks for the improvements and explanation, @gitexel !

jsamoocha:
@lwasser I see now how pre-commit.ci is monitoring the whole repo and cannot be configured to only monitor specific branches for changes. However, I don't think we want it to make extra commits on PRs on master (yet). That's why we merged it into the `pydantic_model` feature branch.

Can we solve this pragmatically by adding an "empty" (but valid) .pre-commit-config.yaml to master? That way, CI will pass but will not make additional commits (yet). I'll make a new tiny PR to propose this.

PR #311, opened 2023-01-07
Prevents pre-commit CI from failing because of missing config file

### LINKED ISSUE #310: CI fails due to missing pre-commit config file
Pre-commit hooks were enabled in a feature branch for testing pre-commit.ci. However, pre-commit.ci now monitors the whole repository and fails the CI checks because it cannot find a config file.

jsamoocha:
closes #310 

## Description

This simply adds an empty pre-commit config file to prevent CI from failing.

## Type of change

## Does your PR include tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,4 @@
+# Placeholder for pre-commit config that will be added later.
+# Pre-commit CI needs this file to exist and will fail otherwise.
+
+repos: []


lwasser:
@jsamoocha i'm going to suggest another approach to this. I really think we should have pre-commit as a part of our dev workflow to enforce code format locally. i'm going to push my other PR but i'll turn off CI checks... we can still call the bot in pr's ... but really i think having commits in both the pydantic branch and here but not checks here for good code style isn't ideal.

jsamoocha:
Closing: #312 is a better solution.

PR #315, opened 2023-01-10
Model extension

jsamoocha:
## Description

This replaces implementations for the classes `Club`, `Gear`, `ActivityTotals`, `AthleteStats`, and `Athlete` by the generated Pydantic models. It adds essential (but not yet complete) backward compatibility with the legacy model/attribute approach.

These new classes can be tested by simply using the client, e.g. 

```python
client.get_club(CLUB_ID)
```

Although far from complete, I'd like to share this to get feedback about the approach before putting in the grunt work of replacing all the other classes. When testing, you'll notice that legacy classes that were not converted are now complaining about being unable to set as attribute objects from the new pydantic-based model.

### Supported Backward Compatibility
(only for the classes mentioned above)

- Legacy (de-)serialization methods of `BaseEntity` (`to_dict()`, `from_dict()`, `deserialize()`)
- Conversion of pydantic model enums to plain (`str`) values
- Native types for timestamp and time delta fields such as `elapsed_time`
- Unit fields

### Not (Yet) Supported Backward Compatibility

- Undocumented fields
- Bound client
- `BatchedResultsIterator`
- **Update** Nested entities (e.g., `ActivityTotals` as part of `AthleteStats`) are not yet of the correct (extended) type

### Breaking Changes

- Types such as `IdentifiableEntity`, `LoadableEntity`, `EntityCollection`, etc., are gone. Instance checks for these types will fail.
- Bike, Shoe as subtypes of Gear are gone. Instance checks will fail.
- repr() will use pydantic defaults and may differ from legacy classes
- and others that will probably pop up with more testing...

## Type of change

## Does your PR include tests

- [x] Yes


stravalib/client.py
```diff
@@ -263,7 +262,7 @@ def get_athlete(self):
         :rtype: :class:`stravalib.model.Athlete`
         """
         raw = self.protocol.get("/athlete")
-        return model.Athlete.deserialize(raw, bind_client=self)
+        return model.Athlete.parse_obj(raw)
```
jsamoocha, line 265: This is an example of how to deserialize the raw response dict into the new pydantic-based classes

stravalib/model.py
```diff
@@ -116,6 +125,80 @@ def __repr__(self):
         )
 
 
+class DeprecatedSerializableMixin(BaseModel):
+    """
+    Provides backward compatibility with legacy BaseEntity
+    """
+
+    @classmethod
+    def deserialize(cls, v):
```
jsamoocha, line 134: This will allow continued use of the legacy (de-)serialization methods, but it will raise a `DeprecationWarning`

stravalib/model.py
```diff
@@ -116,6 +125,80 @@ def __repr__(self):
...
+    * _unit_registry
+    """
+
+    def __getattribute__(self, attr):
+        value = object.__getattribute__(self, attr)
+        if attr in ["_field_conversions", "_unit_registry"]:
+            return value
+        try:
+            if attr in self._field_conversions:
+                value = self._field_conversions[attr](value)
```
jsamoocha, line 188: `_field_conversions` is a class attribute that maps field names to operations that change their type/value when requested

stravalib/model.py
```diff
@@ -116,6 +125,80 @@ def __repr__(self):
...
+        except AttributeError:
+            # Current model class has no field conversions defined
+            pass
+        try:
+            if attr in self._unit_registry:
+                # Return a Quantity
+                value = UnitConverter(self._unit_registry[attr])(value)
```
jsamoocha, line 195: `_unit_registry` is a class attribute that maps field names to units and uses the unit converter to transforms these fields to unit-full quantities when requested

stravalib/tests/unit/test_attributes.py
```diff
@@ -75,6 +77,7 @@ def test_with_incorrect_timezone(self):
         self.assertIsNone(timezone.unmarshal("(GMT+00:00) Factory"))
 
 
+@skip
```
jsamoocha, line 80: Skipping for now, but most of these tests will be removed, together with `attributes.py`

stravalib/unithelper.py
```diff
@@ -42,6 +42,10 @@ def unit(self):
         warn_units_deprecated()
         return str(self.q.units)
 
+    def __eq__(self, other):
```
jsamoocha, line 45: Added this to compare quantities in unit tests. Eventually this will be deprecated and we can use the rich comparison methods from Pint.

lwasser:
ok i'll play with this @jsamoocha ... sadly i haven't used those end points ever except for athlete. but i will test what's there now and then the new iterations locally... it might take me a day or two

`Club, Gear, ActivityTotals, AthleteStats, and Athlete` 

this is so much work!! so appreciative that you are here working on this too!!

lwasser:
@jsamoocha ok so far so good. if you don't mind, im going to ask some questions - some of which might be the way this worked before but i just don't understand some of the serialization and deserialization items. will ask inline.

stravalib/model.py
```diff
@@ -116,6 +125,80 @@ def __repr__(self):
...
+        """
+        warn_method_deprecation(
+            cls,
+            "deserialize()",
+            "parse_obj()",
+            "https://docs.pydantic.dev/usage/models/#helper-functions",
+        )
+        return cls.parse_obj(v)
+
+    def from_dict(self, v):
```
lwasser, line 146: sorry is this deprecated? where is from_dict being defined if it is deprecated here?

This workf locally but i find the variable name `v` to be confusing. I suspect v needs to be a `dict` with the club attributes - 

and is the intended workflow

to_dict --> update a few fields and then 
from_dict - > update the club metadata ?

then maybe push back to strava to update?

So sorry if these are basic questions but i'm trying to understand how this is used and hoping we can clarify what these methods do via better variable names, maybe some type hinting? and better docstrings (i can help too where needed with that part). you are the pydantic model guru here so i'm trying to better understand this shift while also thinking about usability and use cases.

stravalib/model.py
lwasser: ok last question and i think i will have enough to be a bit more helpful in testing. i know this is a private attribute. but what is an example of how this is used? 

is this a helper related to unit conversion meters to feet, etc? many thanks.

stravalib/model.py
jsamoocha: You can use (e.g.) ```Club.deserialize(some_dict)```, which will give you a Club object. That's all to it.

stravalib/model.py
jsamoocha: `_field_conversions` are used when an attribute's value from the pydantic model doesn't match the type or format of the legacy attribute. For example, the Strava API (and pydantic) models durations as simple integers representing seconds. However, the legacy API uses `timedelta` objects for durations. To preserve this behavior, we must convert certain attributes from the pydantic model. The `ActivityTotals` class gives an example of how this is used.

stravalib/model.py
jsamoocha: These (de-)serialization methods are mainly used internally by other (legacy) `model` and `client` code, but are defined as public methods. Simply discarding them may break existing functionality in projects dependent on stravalib.

The preference (which I think we should document) is to use pydantic's (de-)serialization API, which would look like:

```python
c = Club.parse_obj(some_club_dict)
club_as_dict = c.dict() # ... and push to document store
club_as_json = c.json() # ... and send to other API
```

See https://docs.pydantic.dev/usage/models/#helper-functions and https://docs.pydantic.dev/usage/exporting_models/

jsamoocha:
Nested entities (e.g., `ActivityTotals` as part of `AthleteStats`) are not yet of the correct (extended) type (see updated PR description). I'll address that in a next PR to keep the scope of this one as limited as possible.

stravalib/model.py
lwasser: oooh ok i understand now. the previous implementation was

`return model.Athlete.deserialize(raw, bind_client=self)` and it's now changed to `parse_obj(raw)`

i understand. i really don't see any issues with this approach. and am impressed that we can make such a big change to the API and still retain that backwards compatibility. I think you mentioned this early on in terms of raising warnings and such for the legacy methods. so essentially that is then what is happening here:

https://github.com/stravalib/stravalib/pull/315/files#diff-3f71efb739eb651ccb258a4bd554a784752bda3cd6c93cf9f01ddc0cdc1f5a36R139 

and we just need to develop a timeline for when we might drop legacy support? does that sound right? this all is looking great to me so far.

stravalib/model.py
lwasser: ok I will play with this a bit more. I totally remember that timedelta duration as it actually messed up activities that were >24 hours (100 milers) and it took me a minute to figure out that is why my data were off.

stravalib/model.py
lwasser: Oh i get it ... and that part of the API will be more expressive . 

so essentially the use case that i'm hearing here is this is for someone building an app, that might want to grab say club data and add it to a database or wherever they are storing app data via another API. (which might be what you're doing with the application that you're working on?)

This makes a lot of sense and the pydantic docs are really helpful. i'll spend some more time on them. 
So as far as i can tell, this works for me locally and seems seamless.

lwasser (approved):
@jsamoocha this looks good to me. approving. 

one last small question:
> Bike, Shoe as subtypes of Gear are gone. Instance checks will fail.

Where will those items end up ? 
i see it's in summary gear in the API? 
https://developers.strava.com/docs/reference/#api-models-SummaryGear

lwasser:
another question (i keep saying that 😆 )  - this is NOT related to your PR so feel free to merge...  but to the data. when i look at the activity year to date stats it is odd.

My data pulled down suggests i've done 5 runs total (Which is not correct ). on strava is says i have done 7 runs this year. that is closer to reality. Do you see some discrepancies between data pulled down from strava via the API and what's in your account stats? again i doubt this relates at all to code function. I am curious if you see different returned values. 
looking at my activities it should be 7. maybe the endpoint is doing something funky?

stravalib/model.py
jsamoocha: Yes, that's the use case I thought about (and often use myself). Something similar is getting activities from Strava and transforming them into a Pandas DataFrame, which would look like
```python
activity_df = pd.DataFrame([a.dict() for a in client.get_activities()])
```
once the `Activity` class is transformed.

jsamoocha:
> > Bike, Shoe as subtypes of Gear are gone. Instance checks will fail.
> 
> Where will those items end up ? i see it's in summary gear in the API? https://developers.strava.com/docs/reference/#api-models-SummaryGear

The legacy model has a very particular way of deserializing gear into either Bike or Shoe classes:
```python
class Gear(IdentifiableEntity):
    # ...
    @classmethod
    def deserialize(cls, v):
        """
        Creates a new object based on serialized (dict) struct.
        """
        if v is None:
            return None
        if cls == Gear and v.get('resource_state') == 3:
            if 'frame_type' in v:
                o = Bike()
            else:
                o = Shoe()
        else:
            o = cls()
        o.from_dict(v)
        return o
```

These are not (sub-)types in the official Strava API (which only defines `SummaryGear` and `DetailedGear`), and I haven't seen/understood how to hook into pydantic's deserialization to change the type itself of a model object. So, for the time being, I'll leave this as a breaking change.

jsamoocha:
> My data pulled down suggests i've done 5 runs total (Which is not correct ). on strava is says i have done 7 runs this year. that is closer to reality. Do you see some discrepancies between data pulled down from strava via the API and what's in your account stats? again i doubt this relates at all to code function. I am curious if you see different returned values. looking at my activities it should be 7. maybe the endpoint is doing something funky?

My guess would be that Strava maybe runs a weekly batch job to update our yearly stats?

PR #316, opened 2023-01-12
Nested model attributes

jsamoocha:
Adds support for nested model attributes. This ensures that entities within entities (e.g., `ActivityTotals` as part of `AthleteStats`) are also of the correct extended types and will be backward-compatible.

In addition to previous work on this branch, this PR also replaces the following legacy entities:

- ActivityComment
- ActivityPhotoPrimary
- ActivityPhotoMeta
- ActivityKudos

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1903,9 +1903,15 @@ class BatchedResultsIterator(object):
 
         entities = []
         for raw in raw_results:
-            entities.append(
-                self.entity.deserialize(raw, bind_client=self.bind_client)
-            )
+            try:
+                new_entity = self.entity.parse_obj(raw)
+                # TODO: try adding bind_client to entity
+            except AttributeError:
+                # entity doesn't have a parse_obj() method, so must be of a legacy type
+                new_entity = self.entity.deserialize(
+                    raw, bind_client=self.bind_client
+                )
+            entities.append(new_entity)
 
         self._buffer = collections.deque(entities)
 
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -442,6 +442,9 @@ class Client(object):
 
         https://developers.strava.com/docs/reference/#api-Athletes-getStats
 
+        Note that this will return the stats for _public_ activities only,
+        regardless of the scopes of the current access token.
+
         :return: A model containing the Stats
         :rtype: :py:class:`stravalib.model.AthleteStats`
         """
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -6,9 +6,9 @@ Entity classes for representing the various Strava datatypes.
 import abc
 import logging
 from collections.abc import Sequence
-from typing import Dict
+from typing import Dict, List
 
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
 
 from stravalib import exc
 from stravalib import unithelper as uh
@@ -305,11 +305,31 @@ class AthleteStats(
         "biggest_climb_elevation_gain": "meters",
     }
 
+    @validator(
+        "recent_ride_totals",
+        "recent_run_totals",
+        "recent_swim_totals",
+        "ytd_ride_totals",
+        "ytd_run_totals",
+        "ytd_swim_totals",
+        "all_ride_totals",
+        "all_run_totals",
+        "all_swim_totals",
+    )
+    def to_extended_totals(cls, raw_totals: Dict):
+        return ActivityTotals.parse_obj(raw_totals)
+
 
 class Athlete(
     DetailedAthlete, DeprecatedSerializableMixin, BackwardCompatibilityMixin
 ):
-    pass
+    @validator("clubs")
+    def to_extended_clubs(cls, raw_clubs: List[Dict]):
+        return [Club.parse_obj(c) for c in raw_clubs]
+
+    @validator("bikes", "shoes")
+    def to_extended_gear(cls, raw_gear: List[Dict]):
+        return [Gear.parse_obj(g) for g in raw_gear]
 
 
 class ActivityComment(LoadableEntity):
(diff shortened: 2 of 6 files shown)


stravalib/client.py
```diff
@@ -1903,9 +1903,15 @@ def _fill_buffer(self):
...
-            )
+            try:
+                new_entity = self.entity.parse_obj(raw)
+                # TODO: try adding bind_client to entity
+            except AttributeError:
+                # entity doesn't have a parse_obj() method, so must be of a legacy type
+                new_entity = self.entity.deserialize(
+                    raw, bind_client=self.bind_client
+                )
+            entities.append(new_entity)
```
jsamoocha, line 1914: This makes sure that the `BatchedResultsIterator` can parse the new pydantic-based entities (but has a temporary workaround to also handle legacy entities).

stravalib/model.py
```diff
@@ -305,64 +308,57 @@ class AthleteStats(
...
+        "recent_swim_totals",
+        "ytd_ride_totals",
+        "ytd_run_totals",
+        "ytd_swim_totals",
+        "all_ride_totals",
+        "all_run_totals",
+        "all_swim_totals",
+    )
+    def to_extended_totals(cls, raw_totals: Dict):
+        return ActivityTotals.parse_obj(raw_totals)
```
jsamoocha, line 323: Using pydantic's validator mechanism, we can intercept and modify the parsing of fields. In this case, we can make sure that nested entities will be of our backward-compatible extended types instead of the "plain" generated pydantic model type.

lwasser: @jsamoocha i'm working through all of these elements but when we grab athlete stats via the api call is it possible that it only grabs public activities? 

my numbers are still incorrect and i suspect it's because i don't always make things public. is that how ytd_ ... totals are supposed to pull data in? I looked at my scope and `read_all` should include public and private activities as i understand it.

So far everything is working well but the data i'm getting is just not right. i'm going to look at specific activities next.

stravalib/client.py
lwasser: Ok question about the activity results. 

i'm seeing a lot of information in the results that is empty 

```python
# Get activities
act = client.get_activities(limit=12)
for a in act:
    print(a)

a_dict = a.to_dict()
a_dict.keys()
a_dict["athlete"]
```
OUTPUT
```
{'id': None, 'city': None, 'country': None, 'created_at': None, 'firstname': None, 'lastname': None, 'premium': None, 'profile': None, 'profile_medium': None, 'resource_state': None, 'sex': None, 'state': None, 'summit': None, 'updated_at': None, 'bikes': None, 'clubs': None, 'follower_count': None, 'friend_count': None, 'ftp': None, 'measurement_preference': None, 'shoes': None, 'weight': None}
```

i'm just curious why some of this is in the return for activities as it's in the athlete data but that doesn't need to be returned for every activity as it would make the return object quite large.

The other odd thing is the return of a single activit also has this::
```
a.ALPINESKI                    a.embed_token
a.BACKCOUNTRYSKI               a.end_latlng
a.CANOEING                     a.expand(
a.CROSSCOUNTRYSKIING           a.external_id
a.CROSSFIT                     a.flagged
a.EBIKERIDE                    a.from_accepted_tag
a.ELLIPTICAL                   a.from_dict(
a.GOLF                         a.full_photos
```

if the activity has a single type why does it list all of the possible types in the return? i am just curious about what is in this return and why there is so much extra stuff associated with a single activity. (thinking about people processing thousands of activities).

otherwise the code is all working as expected i'm just getting some unexpected return VALUES.

lwasser:
this is great @jsamoocha please see comments above about data being returned. the code works. i'm not getting any errors. just odd return info. 

Also pre-commit is failing black and isort. can we try to run on this PR before merging? to test out our workflow? perhaps you don't have pre-commit running locally yet? but we can run it here! 

Many thanks for all of this work. this is a massive overhaul of our api!!

stravalib/client.py
jsamoocha: What you see for activities is still the legacy API, as I haven't replaced the `Activity` entity yet. That list of _possible_ activity types is probably some legacy way of supporting enum-like behavior (python didn't support that before 3.4). The generated pydantic model now has a proper enum, so these will not appear in an `Activity` object.

stravalib/model.py
jsamoocha: I just tested and can confirm that Strava only counts our public activities in these ytd_* totals. I'll add that to the documentation.

stravalib/client.py
jsamoocha: > i'm just curious why some of this is in the return for activities as it's in the athlete data but that doesn't need to be returned for every activity as it would make the return object quite large

Also this is just part of the legacy API, since it only supports one `Athlete` class (whereas the pydantic model you can see in `strava_model.py` has distinct types such as `MetaAthlete`, `SummaryAthlete` and `DetailedAthlete`. We can later add issues to extend these other types to limit the size of (nested) objects (I've noted this down).

jsamoocha:
pre-commit.ci autofix

stravalib/client.py
lwasser: Ok that all makes sense. I understand this is a very big work in progress!! 

I really like testing this in small pieces so let's add those as issues later if we need to limit redundancy of info in the returned nest data. Many thanks for clarifying.

stravalib/model.py
lwasser: ok perfect. thanks.

lwasser (approved):
Thank you for all of this work, @jsamoocha !! it looks good to me. 

There are some test gaps but I really think that's not worth looking at until the entire model has been rebuilt (and you have enough on your plate doing that!! ) so let's merge this.

looks like the bot is also working as we want it to here! 🎆

PR #318, opened 2023-01-13
feat(ActivityUploader): upload photo to activity

gitexel:
## Description

Upload and attach images to the Strava activities

> NOTE1: In order to upload a photo, the activity must be uploaded and processed.

> NOTE2: The ability to add photos to activity is currently limited to partner apps & devices such as Zwift,
        Peloton, Tempo Move, etc... Given that the ability isn't in the public API, neither are the docs.

After calling the POST https://www.strava.com/api/v3/uploads endpoint:

If the activity file was uploaded successfully, the response code will be 201 Created. Strava then processes your file data to create an activity, which usually takes about 3-8 seconds.

In order to ensure the activity has been processed before we upload the image, we need to call ActivityUploader `wait`.

The response of the above endpoint doesn’t contain the activity_id until the file has been processed which can take a few seconds. Instead, it contains an id that can be used to check the status of the upload request.

```
{
    "id": 1570175274,
    "external_id": "Lunch_Run.gpx",
    "error": null,
    "status": "Your activity is still being processed.",
    "activity_id": null
}
```

By requesting GET https://www.strava.com/api/v3/uploads/:id, we can see if the activity was processed and an activity_id eventually becomes available alongside a photo_metadata field containing information on how to upload an image:

```
{
    "id": 1570175274,
    "external_id": "Lunch_Run.gpx",
    "error": null,
    "status": "Your activity is ready.",
    "activity_id": 1456449319,
    "photo_metadata": [
      {
        "uri": "https://strava-photo-uploads-prod.s3.amazonaws.com/sU&AWSAccessKeyId=Key&Expires=1521148201&Signature=SIG",
        "header": {
            "Content-Type": "image/jpeg"
        },
        "method": "PUT",
        "max_size": 1600
      }
    ]
}
```

> Note: You can upload JPG or PNG images.

Example Request

  $ curl -X PUT "https://strava-photo-uploads-prod.s3.amazonaws.com/snsU&AWSAccessKeyId=Key&Expires=1521148201&Signature=SIG" \
        -H "Content-Type: image/jpeg" \
        --upload-file download.jpg

Example Usage

```python

client = Client()

activity_uploader = client.upload_activity(activity_file=file, data_type='fit')

# we need to wait until the activity is processed
activity_uploader.wait()

activity_uploader.upload_photo(photo=b'photo_data')


```

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
[...]

PR #322, opened 2023-01-21
Add support for lazy loading related entities

jsamoocha:
#### Bound Client

Adds support for "lazy" loading of related entity collections using a bound client, for example:

```python
class Club(
    DetailedClub,
    DeprecatedSerializableMixin,
    BackwardCompatibilityMixin,
    BoundClientEntity,  # <-- this mixin is needed ...
):
    @lazy_property  # <-- ... to make this decorator work
    def members(self):
        return self.bound_client.get_club_members(self.id)
```
The `@lazy_property` decorator checks for the id field and bound client.

#### Some Refactoring

Refactors existing validators for type extensions for nested entities to prevent boilerplate code and pydantic details. For example, the previously used construct in the class `Athlete`

```python
@validator("clubs")
    def to_extended_clubs(cls, raw_clubs: List[Dict]):
        return [Club.parse_obj(c) for c in raw_clubs]
```

now becomes

```python
_clubs_extension = extend_types('clubs', model_class=Club, as_collection=True)
```

#### New Model Classes

In addition to previous PRs on this branch, the following classes are now based on the pydantic model:

- Activity
- BaseEffort
- Segment
- SegmentEffort
- BestEffort
- Split
- Map
- Lap
- SegmentExplorerResult

This should allow traversal of the entire object graph related to `Activity` entities.

#### Client Cleanup

Removes some client methods related to segment leaderboard API endpoints, as these don't exist anymore at Strava.

## Type of change

- [x] Other (please describe)

Refactoring: replacement of legacy model classes by equivalent ones based on the generated pydantic model.

## Does your PR include tests

- [x] Yes

- [x] Check this when your PR has a green check.


.pre-commit-config.yaml
```diff
@@ -45,8 +45,8 @@ repos:
...
-  - repo: https://github.com/psf/black
-    rev: 22.12.0
-    hooks:
-     - id: black
-       language_version: python3.8
+#  - repo: https://github.com/psf/black
+#    rev: 22.12.0
+#    hooks:
+#     - id: black
+#       language_version: python3.8
```
jsamoocha, line 52: Disabling black to prevent merge conflicts with master

stravalib/client.py
```diff
@@ -508,7 +416,7 @@ def get_club(self, club_id):
         :rtype: :class:`stravalib.model.Club`
         """
         raw = self.protocol.get("/clubs/{id}", id=club_id)
-        return model.Club.parse_obj(raw)
+        return model.Club.parse_obj({**raw, **{"bound_client": self}})
```
jsamoocha, line 419: This is how we should add the bound client to an entity when deserializing it. When dropping python 3.8 support in the future, this can be expressed cleaner as `raw | {'bound_client: self}`

stravalib/client.py
```diff
@@ -1906,16 +1673,16 @@ def _fill_buffer(self):
...
         )
 
         entities = []
         for raw in raw_results:
             try:
-                new_entity = self.entity.parse_obj(raw)
-                # TODO: try adding bind_client to entity
+                new_entity = self.entity.parse_obj(
+                    {**raw, **{'bound_client': self.bind_client}}
+                )
```
jsamoocha, line 1685: Just to be sure, binds the client to any entity fetched in `BatchedResultIterator`

stravalib/field_conversions.py
```diff
@@ -35,8 +38,26 @@ def enum_values(enums: Sequence[Enum]) -> List:
...
+        # America/Los_Angeles
+        tzname = tz
+    try:
+        tz = pytz.timezone(tzname)
+    except UnknownTimeZoneError as e:
+        LOGGER.warning(
+            f"Encountered unknown time zone {tzname}, returning None"
+        )
+        tz = None
+    return tz
```
jsamoocha, line 63: For backward compatibility, converts timezone strings to timezone objects

stravalib/model.py
```diff
@@ -17,28 +22,105 @@
...
+    Split,
+    SummaryPRSegmentEffort,
+    SummarySegmentEffort,
 )
 from stravalib.unithelper import UnitConverter
 
+LOGGER = logging.getLogger(__name__)
+
+
+def lazy_property(fn):
```
jsamoocha, line 59: This decorator checks if an entity has a bound client, whether the entity has an `id` field, etc. All these checks were repeated in the legacy model.

stravalib/model.py
```diff
@@ -203,6 +285,13 @@ def __getattribute__(self, attr):
         return value
 
 
+class BoundClientEntity(BaseModel):
+    # Using Any as type here to prevent catch-22 between circular import and
+    # pydantic forward-referencing issues "resolved" by PEP-8 violations.
+    # See e.g. https://github.com/pydantic/pydantic/issues/1873
+    bound_client: Optional[Any] = None
```
jsamoocha, line 292: The "mixin" that allows binding a client object to entities

stravalib/model.py
```diff
@@ -323,27 +423,50 @@ class AthleteStats(
...
+        """
+        if not self.is_authenticated_athlete():
+            raise exc.NotAuthenticatedAthlete("Statistics are only available for the authenticated athlete")
+        return self.bound_client.get_athlete_stats(self.id)
 
-    @validator("bikes", "shoes")
-    def to_extended_gear(cls, raw_gear: List[Dict]):
-        return [Gear.parse_obj(g) for g in raw_gear]
 
+    def is_authenticated_athlete(self):
```
jsamoocha, line 452: This method in the legacy `Athlete` entity is added for backward compatibility

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
-    )  #: Max heartrate for lap
-    lap_index = Attribute(int, (SUMMARY, DETAILED))  #: Index of lap
-    device_watts = Attribute(
-        bool, (SUMMARY, DETAILED)
-    )  # true if the watts are from a power meter, false if estimated
-    pace_zone = Attribute(int, (SUMMARY, DETAILED))  #: (undocumented)
-    split = Attribute(int, (SUMMARY, DETAILED))  #: Split number
+    _naive_local = validator('start_date_local', allow_reuse=True)(naive_datetime)
 
+    _activity_extension = extend_types('activity', model_class='Activity')
```
jsamoocha, line 571: model_class sometimes needs to be given as str because the type is not defined (yet)

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
-        """
-        The :class:`stravalib.model.SegmentLeaderboard` object for this segment.
-        """
-        if self._leaderboard is None:
-            self.assert_bind_client()
-            if self.id is not None:
-                self._leaderboard = self.bind_client.get_segment_leaderboard(
-                    self.id
-                )
-        return self._leaderboard
```
jsamoocha, line 728: Segment leaderboards API was removed by Strava

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
+class SegmentEffortAchievement(BaseModel):
```
jsamoocha, line 673: This is a small example of how to add undocumented entities/fields to our model. Following PRs should add the undocumented fields from the legacy model.

stravalib/model.py
```diff
@@ -818,460 +725,61 @@ class SegmentEffort(BaseEffort):
...
-        """
-        if self._related is None:
-            if self.athlete_count - 1 > 0:
-                self.assert_bind_client()
-                self._related = self.bind_client.get_related_activities(
-                    self.id
-                )
-            else:
-                self._related = []
-        return self._related
```
jsamoocha, line 1221: The endpoint for related activities does not exist anymore at Strava

stravalib/tests/functional/test_client.py
```diff
@@ -343,38 +313,6 @@ def test_segment_explorer(self):
...
-        # Now will have to do a look up for the authenticated athlete and check the ids match
-        self.assertTrue(a.is_authenticated_athlete())
-
-    def test_not_authenticated_athlete_is_false(self):
-        CAV_ID = 1353775
-        a = self.client.get_athlete(CAV_ID)
-        self.assertEqual(a.resource_state, attributes.SUMMARY)
-        self.assertFalse(a.is_authenticated_athlete())
-
-
```
jsamoocha, line 377: This is move to the integration test, using the mock API

stravalib/tests/functional/test_client_write.py
```diff
@@ -40,7 +40,7 @@ def test_update_activity(self):
         now = datetime.now().replace(microsecond=0)
         a = self.client.create_activity(
             "test_update_activity#create",
-            activity_type=model.Activity.RIDE,
```
jsamoocha, line 43: These constants don't exist anymore. We may need to support these if any user code is highly dependent on them.

stravalib/model.py
lwasser: what is the PEP-8 part of this? is it the circular import?

lwasser: one other question. Can you explain when you say bind a client object to entities... what is an example of this?

stravalib/client.py
lwasser: please tell me if i understand this correctly. When you initially instantiate a Client object, it is empty and a workflow slowly adds content to Client attributes such as authentication. Once you download activities you get a BatchedResult Iterator. So when you say bind, does this just mean that the activity object (the iterator) is then "connected to" or a part of the Client object? i'm a bit confused because the `client.get_activities()` returns an iterator but it's not necessarily a part of the `client` object created. just trying to understand language if you don't mind!

.pre-commit-config.yaml
lwasser: we can run black once the big overhaul is done.  you can also add a `skip: ()` element to the config file but this is totally fine.

lwasser:
@jsamoocha this all looks good. I have some general questions about how lazy loading works. 

> ```
> class Club(
    DetailedClub,
    DeprecatedSerializableMixin,
    BackwardCompatibilityMixin,
    BoundClientEntity,  # <-- this mixin is needed ...):
    @lazy_property  # <-- ... to make this decorator work
    def members(self):
        return self.bound_client.get_club_members(self.id)
```
> The @lazy_property decorator checks for the id field and bound client.

I understand lazy loading in the context of data. However, here, the Club class has a method called members that returns club members based upon the club id. so lazy loading in theory would ensure that the method isn't run or the data isn't actually opened / collected until the user called it. 

Doesn't a method work as such that a user would have to call client.get_athlete_clubs() which would return a club object. and from there you could get club members. But you'd have to call `club_obj.members(id)` for the data to be actually loaded? i'm trying to wrap my head around where and when the actual data are downloaded via the api and stored and what lazy means in this instance. when you call `club_object.members()` for instance is that hitting the API again?  

Many thanks . this all looks good but there are some pieces I don't fully understand yet!

File: .pre-commit-config.yaml (hunk 1 above)
Comment (lwasser) [reply .pre-commit-config.yaml:52]: actually i think the skip might be for CI only. so this is good to just comment out fully until we are done with this big overhaul!

File: stravalib/model.py [hunk 14]
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
 
 
-class ActivityLap(LoadableEntity):
+class ActivityLap(Lap, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
+    _field_conversions = {
+        'elapsed_time': time_interval,
+        'moving_time': time_interval
+    }
 
-    name = Attribute(str, (SUMMARY, DETAILED))  #: Name of lap
```
lwasser, line 437: what is happening here? were these items removed from the API? but i see there's a backward compat mixin now on ActivityLab.

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
+    _athlete_extension = extend_types('athlete', model_class=Athlete)
 
-class Map(IdentifiableEntity):
-    id = Attribute(str, (SUMMARY, DETAILED))  #: Alpha-numeric identifier
-    polyline = Attribute(str, (SUMMARY, DETAILED))  #: Google polyline encoding
-    summary_polyline = Attribute(
-        str, (SUMMARY, DETAILED)
-    )  #: Google polyline encoding for summary shape
+
+class Map(PolylineMap):
```
lwasser, line 575: Going to keep looking here but is this the streaming data for activities? if so, why is map now an empty class? should i try to run / download the geo data from an activity?

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
-    moving_time = (
-        TimeIntervalAttribute()
-    )  #: :class:`datetime.timedelta` of moving time for split
-    average_heartrate = Attribute(float)  #: Average HR for split
-    split = Attribute(int)  #: Which split number
-    pace_zone = Attribute(int)  #: (undocumented)
-    average_speed = Attribute(float, units=uh.meters_per_second)
-    average_grade_adjusted_speed = Attribute(float, units=uh.meters_per_second)
+    _unit_registry = {
+        'distance': 'meters',
```
lwasser, line 586: similar to above - i see several items removed from this object such as moving time. were those items dropped from the API?

lwasser: same question for below - things seemed to be removed. many thanks

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
+        'elev_difference', 'meters',
+        'distance', 'meters'
+    }
 
-    @property
+    _latlng_extensions = extend_types('start_latlng', 'end_latlng', model_class='LatLon')
+
+    @lazy_property
     def segment(self):
         """Associated (full) :class:`stravalib.model.Segment` object."""
```
lwasser, line 613: so in the old API, there was a check for whether the ID existed. But here that is removed. and i see it's a lazily loaded element. is the ID being defined somewhere else now? i haven't play with the segment API endpoint yet so I will.

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
-        if self._segment is None:
-            self.assert_bind_client()
-            if self.id is not None:
-                self._segment = self.bind_client.get_segment(self.id)
-        return self._segment
+        return self.bound_client.get_segment(self.id)
 
 
-class AthleteSegmentStats(BaseEntity):
+class AthleteSegmentStats(SummarySegmentEffort, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
```
lwasser, line 617: it looks like segments was broken into stats and streams?
so count is no longer a part of the stats? 

https://developers.strava.com/docs/reference/#api-Streams-getSegmentEffortStreams

stravalib/model.py
lwasser: ok perfect. i asked this above. ignore my question there :)

stravalib/model.py
```diff
@@ -432,379 +553,165 @@ class ActivityKudos(Athlete):
...
-    average_cadence = Attribute(
-        float, (SUMMARY, DETAILED)
-    )  #: Average cadence during effort
-    start_index = Attribute(
-        int, (SUMMARY, DETAILED)
-    )  #: The activity stream index of the start of this effort
-    end_index = Attribute(
-        int, (SUMMARY, DETAILED)
-    )  #: The activity stream index of the end of this effort
+    _field_conversions = {
```
lwasser, line 702: another question (bare with me!). 
So here i see the backwardCompatibilityMixin is used and the deprecated mixin is used too. Does that mean that this endpoint is allowing for backwards compatibility. but if so how does backwards compat  work if all of the attributes that were there are removed? 

and why would it be deprecated if the api returns data? i may just not understand. 

is it deprecated because we are deprecating the WAY someone accesses the data via stravalib? but the api still returns this information.

stravalib/model.py
```diff
@@ -818,460 +725,61 @@ class SegmentEffort(BaseEffort):
...
-    photos = EntityAttribute(
-        ActivityPhotoMeta, (DETAILED,)
-    )  #: A new photo metadata structure.
-    instagram_primary_photo = Attribute(
-        str, (DETAILED,)
-    )  #: (undocumented) Appears to be the ref to first associated instagram photo
-
-    partner_logo_url = Attribute(str, (DETAILED,))  #: (undocumented)
-    partner_brand_tag = Attribute(str, (DETAILED,))  #: (undocumented)
+    _field_conversions = {
```
lwasser, line 736: ah... ok excited to try this locally. i think this might be why i was getting some odd data returned last time (in the last pr!)

lwasser (commented):
ok i have a few questions (inline) and then i'm also running into an error with members.
My approach:

```python
# All this works great!
ath = client.get_athlete()
clubs = ath.clubs
aclub = clubs[0]
clubid = aclub.id
# This returns an unable to fetch objects error
aclub.members
```


```python
>>> aclub.members()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/leahawasser/Documents/GitHub/stravalib/stravalib/model.py", line 269, in __getattribute__
    value = object.__getattribute__(self, attr)
  File "/Users/leahawasser/Documents/GitHub/stravalib/stravalib/model.py", line 72, in wrapper
    raise exc.UnboundEntity(
stravalib.exc.UnboundEntity: Unable to fetch objects for unbound <class 'stravalib.model.Club'> entity.
```

I haven't tried too much else yet but i wanted to submit this first to not overwhelm you. this is so much work!! i really appreciate all of the changes. and tell me if this error is an error on my end??

stravalib/client.py
jsamoocha: > i'm a bit confused because the client.get_activities() returns an iterator but it's not necessarily a part of the client object created

It's the other way around: the `client` object becomes a part of the entities (or iterators of entities) it returns.

In this case, when a `BatchedResultsIterator` is created (e.g., as part of `client.get_activities()`), the client attaches itself to this iterator as follows:

```python
return BatchedResultsIterator(
            entity=model.Activity,
            bind_client=self,  # <-- client "binds" itself to the iterator
            result_fetcher=result_fetcher,
            limit=limit,
        )
```

This is not hitting the API yet. Now, when we request the next batch of results from this iterator, e.g., by `next()` or `list()`, it internally will hit the Strava API to fetch entities (see `_fill_buffer(self)`) and attach its client reference (pointed to in the snippet above) to each of these entities.

By making a `client` object part of an `Activity` object, we can now use constructs like these:

```python
@lazy_property
    def comments(self):
        return (
            self
            .bound_client  # <-- this is the same "bind_client=self" object from the snippet above
            .get_activity_comments(self.id)
        )
```

which makes it appear to the outside world that `comments` is a plain attribute of an `Activity` object, while in fact, referring to `some_activity.comments` will hit the API through the bound client.

stravalib/model.py
jsamoocha: > what is the PEP-8 part of this? is it the circular import?

Circular imports will prevent this from running at all. As I understood from the linked discussion, circular imports could be prevented by putting some import statements at the end of the module, which is against PEP-8.

jsamoocha:
> and tell me if this error is an error on my end??

It is not. I overlooked binding the client to nested entities (or collections thereof). Will fix that soon. Thanks for testing that!

stravalib/client.py
```diff
@@ -351,67 +318,6 @@ def update_athlete(
         raw_athlete = self.protocol.put("/athlete", **params)
         return model.Athlete.parse_obj(raw_athlete)
 
-    def get_athlete_followers(self, athlete_id=None, limit=None):
```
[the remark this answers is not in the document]
jsamoocha: Yes, athlete friends/followers endpoints were removed.

stravalib/client.py
```diff
@@ -351,67 +318,6 @@ def update_athlete(
...
-        # result_fetcher = functools.partial(self.protocol.get,
-        #                                    '/athletes/{id}/both-following',
-        #                                    id=athlete_id)
-        #
-        # return BatchedResultsIterator(entity=model.Athlete,
-        #                               bind_client=self,
-        #                               result_fetcher=result_fetcher,
-        #                               limit=limit)
 
     # TODO: Can't find this in the api documentation either. Does it still work?
```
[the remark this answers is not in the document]
jsamoocha: `/athletes/{id}/koms` is undocumented but still seems to work. We may want to add a user warning here. It may be a good idea to make a separate issue for systematically checking which client methods are still documented, just silently supported, or have stopped working altogether.

stravalib/client.py
```diff
@@ -1989,7 +1757,7 @@ def photo_metadata(self):
         Note2: available only for partner apps (not for regular account)
```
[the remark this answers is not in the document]
jsamoocha: I have no idea, this was introduced in the latest PR #318.

stravalib/model.py
jsamoocha: No, nothing was removed. `ActivityLap` inherits from the `Lap` class in the `strava_model` module (which, in turn, is generated by the bot that checks Strava's API). In the `Lap` class, you'll find all the documented attributes such as `name`, `distance`, etc.

The difference is that we don't need to maintain all these attributes anymore; the bot will do that for us. The only thing we need to do in our extended classes is to customize some behavior and maintain the undocumented attributes (adding these still needs to be done).

stravalib/model.py
jsamoocha: This is not the stream data (I haven't processed that entity yet). As mentioned above, this class is not empty: it inherits from `PolylineMap`. By subclassing it, we can use the name `Map` for this type to preserve backward compatibility.

stravalib/model.py
jsamoocha: I hope the above answers about inheritance from the generated model explain enough. If not, please let me know.

stravalib/model.py
jsamoocha: To prevent code repetition, the checks for self.id and self.bound_client have moved to this `@lazy_property` decorator (you can see its full definition in `model.py`):

```python
            if obj.bound_client is None:
                raise exc.UnboundEntity(
                    f"Unable to fetch objects for unbound {obj.__class__} entity."
                )
            if obj.id is None:
                LOGGER.warning(f'Cannot retrieve {obj.__class__}.{fn.__name__}, self.id is None')
                return None
```

stravalib/model.py
jsamoocha: I haven't touched streams yet, so I don't know.

stravalib/model.py
jsamoocha: > but if so how does backwards compat work if all of the attributes that were there are removed?

See the [previous conversations](https://github.com/stravalib/stravalib/pull/322#discussion_r1083843237). The attributes are inherited from the generated model.

The `DeprecatedSerializableMixin` allows us to deprecate (but still support for the time being) the legacy (de-)serialization methods `deserialize()`, `from_dict()`, and `to_dict()`. It is not deprecating the entity itself or any endpoint returning these entities, but just these three (de-)serialization methods.

stravalib/client.py
lwasser: ok got it. thank you for taking the time to help me understand the way these objects interact!!

stravalib/model.py
lwasser: yes aboslutely - i understand better how the "binding" works and the objects interact. and yes that would be really weird to add imports to the bottom of a module and in conflict with pep 8! so i understand that element too now! thank you. thanks!

stravalib/client.py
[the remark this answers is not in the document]
lwasser: thanks!

stravalib/client.py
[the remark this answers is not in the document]
lwasser: AHHH ok that sounds good. do you want me to open issues to atleast help a little bit here ? i can atleast do that from this PR so you can keep moving on code. 

i can spend a bit more time working through running the various attributes and such later this week as well but we could still merge and i can open issues if i find other code related issues while testing methods and attrs.

stravalib/model.py
lwasser: Ahhh ok thank you again. i appreciate the time youre taking here to explain this all.

stravalib/model.py
lwasser: ok got it. thank you again!

stravalib/model.py
lwasser: it does. i see how i was really asking the same question over and over! appreciate your patience!!

stravalib/model.py
lwasser: ok no problem. !

stravalib/model.py
lwasser: got it. and yes again i see i'm asking similar questions :)

stravalib/client.py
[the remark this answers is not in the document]
jsamoocha: See #323 . I'm in favor of letting that wait until this work is finished - I'm a very bad multitasker ¯\_(ツ)_/¯

lwasser:
All is working well here. one note / comment. i noticed that when running to_dict it also returns the bound client. i wasn't sure how that returned object would be used or what the use case was for it. 

but this shouldn't stop us from merging this pr i think it's more of an observation. otherwise methods that i've tested (including members) are now working well. 

```python 
d = aclub.to_dict()
>>> d
{'bound_client': <stravalib.client.Client object at 0x11e6c4910>, 'id': 2943, 'name': 'Boulder Trail Runners', 'resource_state': 2, 'activity_types': [<ActivityType.Run: 'Run'>, <ActivityType.VirtualRun: 'VirtualRun'>, <ActivityType.Wheelchair: 'Wheelchair'>]....}
```

lwasser:
Ok two somewhat silly questions. 

1. are the functions in field_conversions just meant to be run independently on the returned activity data values?  
2. when i run `get_activities` there is a weird map element that i'm not sure what to do with. its a summary polyline which i would expect to contain x,y coords. but see what i get below:

```python
act = client.get_activities(limit=1)
for stuff in act:
    print(stuff)

stuff.map
```
Returns
```
Map(id='a8455294762', polyline=None, summary_polyline='es|rFtomaSXXl@jACx@@`@FRLNHP?\\AGC@DBBCEV?v@Lz@CLKHDF?@CEBNIMEHR|@Xh@H?VFr@j@v@EJ@l@dBOW]ALAEB?BLAJ@LRLZHNXtA?VP^RbABZDVATGLGBg@?MEa@U_@ESMSAMEWCEIEA{AHUKm@IIBQLODGE[a@QEg@GE?OJg@CW@OEQ@GAKG]g@YOUWWOOAGJWzBAvCCJi@|@ERKxB@h@GbAGRORSb@w@^CJ@f@d@hALJLTH^Dh@h@~@P`@`AdANDNNDd@E`BJLJd@^b@NHLCJG\\u@Z{@NYTUfAo@`@Ib@EtA?JDFDV`@vAvAtAx@fBvAFJZRPZ`@`BJTHJH@HAd@a@JSBUHWTSZm@PSRSPCJBDHBL@bBH\\\\n@HZFFNAh@QHKFU?a@Kc@AOHiA?cAFm@DKFGLGl@MZMHYBYG_@MYa@]_@g@MYQoA@c@Gs@@s@Ci@WgCEOa@e@GUCY?]DORWFMHy@Is@Qa@Aa@BO@SCWIOs@E`@B]@CCIJCFKHKCMGOWk@}AIe@GmAF_EAoA@{BA}ABuBC[Wm@OY]Q[Ea@BsBCU@IDGA?ECD[EYA}AFi@CW@{@?@ACFc@cAQU_@s@QKQw@')

```

Is this some sort of representation of map / spatial data that i don't understand? is this an odd api return? is this simply because you haven't worked on the streaming data yet and the summary polyline is some compressed version of that? any feedback would be great!!

I think we can generally merge this PR . you have done so much work. but if you can help me understand the above things that would be great!!

lwasser (approved):

jsamoocha:
Thanks for reviewing, that was quite a chunk of changes!

About the bound client being serialized: great catch! I'll address that in the next PR.

About the polyline map: I had no idea what it represented. Its contents are the same as when using the legacy model/client. It seems to be a compressed encoding of an array of lat/lon coordinates. (https://developers.google.com/maps/documentation/utilities/polylinealgorithm)

lwasser:
it's a huge chunk of changes!!  i haven't looked at the most recent PR but will this weekend! phew!

So i looked into that encoding and there is a nice polyline tool on PyPI that decodes. It worked well and i was able to turn it in a geodataframe object. i am now fighting with VSCode and ipython so i can plot. I'm new to using vscode but like it aside from the fact that plotting is now different! 

Anyway we can add that to the docs (i will add once i get time to work on the tutorial again!

`pip install polyline` (also on conda)
then 
```python
import polyline
decoded_poly = polyline.decode(polyobj)
```

stravalib/client.py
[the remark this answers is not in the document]
gitexel: There are some useful documents explaining partner apps.
https://www.strava.com/settings/partner_integrations
https://blog.strava.com/gym-studio-sync/

PR #329, opened 2023-02-05
Fix: Corrects attribute lookup for enum values

### LINKED ISSUE #330: Potential issue with backwards compatibility of returned objects
@jsamoocha 
I just delete the version things then test.

there is a break change in my repo.

check 
https://github.com/yihong0618/running_page/blob/master/scripts/generator/__init__.py#L55

old return str like "Run"

now return a model --> ActivityType.Run

image

_Originally posted by @yihong0618 in https://github.com/stravalib/stravalib/issues/328#issuecomment-1418358205_

jsamoocha:
Closes #330

## Description

Fixes enum lookup behavior. Enums in the model should be converted to their basic value for backward compatibility. Pydantic should support this out of the box by providing config, but unfortunately, it doesn't work for collections of enums in model subclasses (see https://github.com/pydantic/pydantic/issues/5005).

This fixes the order of attribute lookup because enums do not raise `AttributeError` when being assigned new attributes.

## Type of change

## Does your PR include tests

- [x] Yes

NA, will happen when merging into master

### DIFF
--- a/stravalib/field_conversions.py
+++ b/stravalib/field_conversions.py
@@ -36,6 +36,7 @@ def enum_value(v: Enum) -> Any:
 def enum_values(enums: Sequence[Enum]) -> List:
     # Pydantic (1.x) has config for using enum values, but unfortunately
     # it doesn't work for lists of enums.
+    # See https://github.com/pydantic/pydantic/issues/5005
     return [enum_value(e) for e in enums]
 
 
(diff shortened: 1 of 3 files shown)


lwasser:
@jsamoocha what code triggered the error here? i can test locally to see if i see the same behavior but otherwise we can just approve if this fixes an error that you found.

jsamoocha:
If you want to test, the only thing to look out for is what also changed in the unit test: `Activity.type` (or `sport_type`) should return the string `Run` (or `Ride` or whatever else) instead of the full enum object. Same for the list `Club.activity_types`.

lwasser:
ok looks good!!

Was:
returns - <SportType.TrailRun: 'TrailRun'>

with this pr i get:
'TrailRun'

lwasser (approved):

PR #333, opened 2023-02-07
Adds some Strava API changes, and datamodel-code-generator bug fix

jsamoocha:
## Description

Strava introduces ClubActivity and ClubAthlete types that replace the existing SummaryActivity and SummaryAthlete types that are part of a club.

A bug fix in the datamodel-code-generator package now ensures that no enum types are created. This causes some slight changes in enum handling.

## Type of change

- [x] Other (please describe)

Strava API change

## Does your PR include tests

- [x] Yes

### DIFF
--- a/stravalib/field_conversions.py
+++ b/stravalib/field_conversions.py
@@ -1,11 +1,12 @@
 import logging
 from datetime import timedelta
-from enum import Enum
 from functools import wraps
-from typing import Any, Callable, List, Optional, Sequence
+from typing import Any, Callable, List, Optional, Sequence, Union
 
 import pytz
 from pytz.exceptions import UnknownTimeZoneError
+
+from stravalib.strava_model import ActivityType, SportType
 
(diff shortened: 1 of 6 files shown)


jsamoocha:
Added one bug fix that would otherwise be gone by merging `pydantic_model` into master, see #334.

lwasser (approved):
I think we can merge this - thank you @jsamoocha !!

lwasser:
gosh do you find the numerous codecov comments helpful? i think i would for test only pr's but i find it also adds noise to a pr where i just want to look at the code!

jsamoocha:
> gosh do you find the numerous codecov comments helpful? i think i would for test only pr's but i find it also adds noise to a pr where i just want to look at the code!

While reviewing, you can disable them file-by-file by choosing not to view annotations. There are many more options that we could potentially configure, see https://docs.codecov.com/docs/github-checks.

PR #334, opened 2023-02-11
fix LatLon unmarshal: deal with string type

oliverkurth:
*The text above will ensure the issue will be closed when this PR is merged*

## Description

The `LatLon` object will be written to file as a string in the format `"{lat},{lon}"`. However, on `unmarshal()`, it is expected to be a list. This results in the value being read wrong. For example, `"start_latlng": "47.441764,-121.672241"` will be result in a a `LatLong` object with `lat` = `4` and `lon` = `7`.

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 * Add: Support uploading `activity_file` object with type `bytes` (@gitexel, #308)
 * Add: Pre-commit hook + instructions and configure precommit.ci bot (@lwasser, #293)
 * Fix: Setuptools_scm bug when installing stravalib remotely via GitHub (@lwasser, #331)
+* Fix: fix LatLon unmarshal from string type (@oliverkurth, #334)
 
 ## v1.1.0
 
--- a/stravalib/attributes.py
+++ b/stravalib/attributes.py
@@ -182,7 +182,11 @@ class LocationAttribute(Attribute):
         """
         """
         if not isinstance(v, LatLon):
-            v = LatLon(lat=v[0], lon=v[1]) if v else None
+            if type(v) == str:
+                lat, lon = v.split(',')
+                v = LatLon(lat=lat, lon=lon)
+            else:
+                v = LatLon(lat=v[0], lon=v[1]) if v else None
         return v
 
 


jsamoocha:
Hi, thanks for your contribution! I see the inconsistency that it fixes.

However, there's a big change coming soon to this package, where (a.o.) all this manual (un-) marshaling logic will be replaced by a pydantic model (see the branch `pydantic_model`). The `attributes` module will not exist anymore in the following release. For that reason, I don't think it makes sense to merge this change into `master`.

In case you (and potentially several others) have already serialized `LatLon` objects as `str`, I added the custom `str` deserialization to that new branch (see #333, and https://github.com/stravalib/stravalib/pull/333/commits/250680bc66e5e9ec1d0a7a1747645c6bda0a7b20).

If this is ok with you, I'll close this PR. If not, please let me know!

lwasser:
One quick thought before we close!! And wow thank you so much Oliver for
this contribution!

I was wondering if this fixes a bug encountered in an app or tool you are
working on. Jonatan from a user perspective when do you think we can merge
that branch here? I'm asking because if we wanted to wait a bit longer and
ask the community to try it out it could be worth merging small but
important fixes like this so applications still run smoothly. We could
create another minor release here and then plan to merge pydantic (unless
of course said fix would cause you more work)

This is just a thought. I'd love to do what's best for both us as
maintainers and our users ! But what is best here?

Cheers!!

oliverkurth:
Thank you for the reply!

I would appreciate it it this fix would be merged, no matter if this gets deprecated or not. There may be a breaking change when you switch to pydantic and then it would be good to have something to fall back on. And I think this change has very low risk to break anything.

I am in a similar situation right now with another big transition in stravalib- my code (https://github.com/oliverkurth/runquery) needs to branch of stravalib (temporarily) from the point just before the switch from `units` to `pint`, plus this fix. Users in the same situation could just cherry-pick this change.

Also, as additional info, my code used the fields `start_longitude` and `start_latitude` to work around this bug. But looks like these aren't filled in by strava any more (they are `null`), so now I need the `start_latlng` field.

jsamoocha:
> I would appreciate if this fix would be merged, no matter if this gets deprecated or not.

That's great; I'll merge it. I just wasn't sure about the urgency. In any case, the `pydantic_model` branch already has this fix, thanks to your input.

> ... needs to branch of stravalib (temporarily) from the point just before the switch from units to pint

Can you maybe tell me more about how that change broke your code? We may be able to fix that, so you don't need to cherry-pick anymore.

@lwasser Yes, it sounds like a good idea to make a (minor) release with the changes so far before we merge `pydantic_model` into `master`.

jsamoocha (approved):

oliverkurth:
> That's great; I'll merge it. I just wasn't sure about the urgency. In any case, the pydantic_model branch already has this fix, thanks to your input.

Thank you very much!

> Can you maybe tell me more about how that change broke your code? We may be able to fix that, so you don't need to cherry-pick anymore.

I copied the error message and some context to this issue on my project (I don't want to derail this thread): https://github.com/oliverkurth/runquery/issues/6 . It may be not a big deal to fix it, not sure. But yesterday I just wanted to get it working.

lwasser:
Yay!! I can create a minor release in the next day or two!

PR #351, opened 2023-03-05
Fix: pin pydata sphinx theme to avoid icon error issue

lwasser:
This just fixes the current PR where docs are not building due to a bug in the [pydata sphinx theme ](https://github.com/pydata/pydata-sphinx-theme/issues/1226)

docs work with version 0.12 so i'm pinning to that version until i see the bug is fixed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,8 @@ pytest
 pytest-cov
 
 # Docs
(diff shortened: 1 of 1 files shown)


lwasser:
feel free to merge this if you are ok with it @jsamoocha ! it should fix your PR.

Issue #326, opened 2023-01-29
DOC: update / fix docstrings throughout to be numpy style


lwasser:
Currently there are a few issues thorughout our API related to documentation. 

- [x] Docs have a mixture of numpy style and rst style . We should move to numpy style to facilitate consistent documentation
- [x] Pydantic makes autodoc a bit more complex. however we can easily using autodoc-pydantic extension to create cleaner documented objects that use / inherit from pydantic objects. 
- [x] move to pydata sphinx theme to fix the autodoc toctree issues.

I am going to work on this in pieces - starting first with simply cleaning up the docs and fixing docstrings that don't render. 

Then in a second push we can begin to document each module a bit more carefully adding examples where needed. 

## Files to fix

- [x] client.py
- [x] model.py

lwasser:
@jsamoocha i do want to be careful with this as you're adding commits. this is why i thought it might be nice to merge that current PR #324 and then i can add on top of it a few cleanup docs focused pr's! this will take a bit of the onus off of you to cleanup docstrings but you can review my pr's and suggest changes as needed. i think examples will be really helpful to add and we can do it in pieces :) .

lwasser:
i also want to move us to pydata sphinx theme 

essentially using furo i'm running into TOC issues where each object that is documented is listed in the left side bar. this is a default for furo. But i think moving to pydata we can customize. i'll wait to do this once we merge the pydantic branch. 

<img width="1202" alt="Screen Shot 2023-02-03 at 8 21 24 PM" src="https://user-images.githubusercontent.com/7649194/216745573-e6404cc1-957f-47e1-a40d-c0923e40810a.png">

Issue #323, opened 2023-01-24
META: check client methods vs. Strava API consistency


jsamoocha:
Methods from the `client` module can have one of the following states compared with the Strava API:

1. Officially documented
2. Officially deprecated
3. Undocumented but somehow working
4. Undocumented and unsupported (returning 4XX responses)

We need to systematically check the methods in `client`, and for states 2/3/4, we need to investigate possible solutions and create new issues where necessary.

futureshape:
Issue #238 has an example of a method removed from the Strava API but still in the library

lwasser:
@jsamoocha is there anything left todo on this task? i think the[ one issue](https://github.com/stravalib/stravalib/issues/528) around checking undocumented attributes is a good one for us to implement! But i think we've gone through most if not all of the API endpoints at this point to check for what's documented and what's returned. 

Can we close this? ✨

jsamoocha:
Given our limited availability, it may be better to deal with inconsistencies between Strava API and our client methods on a case-by-case basis (i.e., when a user reports something is broken or doesn't work as expected), instead of the proactive approach suggested in this issue. Closing this now.

PR #356, opened 2023-03-28
Fix: Move all code to numpy style docs

lwasser:
begins to address #326 

## Description
I can't seem to find the original discussion but we had discussed moving all code to numpy style docstrings. its currently a mix of both. this is a single file - client.py that we can look at together and decide how we wish to proceed. i started to add a but of typing to this but wasn't sure how far we wanted to go with that...  

Following this guide here: https://numpydoc.readthedocs.io/en/latest/format.html

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

i haven't added a changelog entry for this yet - just wanted to open this to see if we are all happy with moving in this direction to clean up our docs a bit.


stravalib/client.py
```diff
@@ -32,12 +32,17 @@
...
 
 class Client(object):
-    """
-    Main client class for interacting with the exposed Strava v3 API methods.
+    """Main client class for interacting with the exposed Strava v3 API methods.
 
     This class can be instantiated without an access_token when performing authentication;
     however, most methods will require a valid access token.
 
+    Parameters
```
lwasser, line 40: i need to do a bit of digging. if the class is just an object being created we may not need these two empty items. they seem pointless.

stravalib/client.py
```diff
@@ -85,15 +90,21 @@ def __init__(
...
 
     @access_token.setter
     def access_token(self, v):
-        """
-        Set the currently configured authorization token.
+        """Set the currently configured authorization token.
+
+        Parameters
+        ----------
+        v :
```
lwasser, line 102:
```suggestion
        v : int
             User's access token for authentication.
```

stravalib/client.py
```diff
@@ -177,8 +186,8 @@ def refresh_access_token(self, client_id, client_secret, refresh_token):
 
         Returns
         -------
-        Dictionary containing the access_token, refresh_token
-        and expires_at (number of seconds since Epoch when the provided access token will expire)
+
```
lwasser, line 189:
```suggestion
Dictionary containing the access_token, refresh_token
        and expires_at (number of seconds since Epoch when the provided access token will expire)
```

stravalib/client.py
```diff
@@ -380,13 +406,18 @@ def get_athlete_stats(self, athlete_id=None):
...
-        """
-        List the clubs for the currently authenticated athlete.
+        """List the clubs for the currently authenticated athlete.
 
         https://developers.strava.com/docs/reference/#api-Clubs-getLoggedInAthleteClubs
 
-        :return: A list of :class:`stravalib.model.Club`
-        :rtype: :py:class:`list`
+        Parameters
+        ----------
```
lwasser, line 414:
```suggestion
```

stravalib/client.py
```diff
@@ -532,39 +585,39 @@ def get_friend_activities(self, limit=None):
 
     def create_activity(
         self,
-        name,
-        activity_type,
-        start_date_local,
-        elapsed_time,
-        description=None,
+        name:str,
```
lwasser, line 588: Here i started to add typing to the definitions. do we want to proceed with this approach? it will take a bit of time to do but could be worthwhile.

stravalib/client.py
```diff
@@ -967,14 +1036,20 @@ def get_activity_laps(self, activity_id):
...
+        """Deprecated. This endpoint was removed by strava in Jan 2018.
 
-        :param activity_id: The activity for which to fetch related activities.
-        :type activity_id: int
+        Parameters
+        ----------
+        activity_id : int
+            The activity for which to fetch related activities.
+        limit :
+             (Default value = None)
```
lwasser, line 1046:
```suggestion
        limit : int, default=None
             Rate limit value for getting activities
```

stravalib/client.py
```diff
@@ -32,12 +32,17 @@
...
+    """Main client class for interacting with the exposed Strava v3 API methods.
 
     This class can be instantiated without an access_token when performing authentication;
     however, most methods will require a valid access token.
 
+    Parameters
+    ----------
+
+    Returns
+    -------
```
lwasser, line 44:
```suggestion
```

stravalib/client.py
```diff
@@ -154,16 +160,15 @@ def exchange_code_for_token(self, client_id, client_secret, code):
 
         Returns
         -------
-        Dictionary containing the access_token, refresh_token
-        and expires_at (number of seconds since Epoch when the provided access token will expire)
+
+
```
lwasser, line 164:
```suggestion
        Dictionary containing the access_token, refresh_token
        and expires_at (number of seconds since Epoch when the provided access token will expire)
```

stravalib/client.py
jsamoocha: Does this prevent us from adding types in the docstrings as well? If not, adding typing everywhere is an awful lot of work.

lwasser: @jsamoocha this is a good point. we could start by only adding docstring type info / cleaning up docstrings. 

Then, there are a few tools that can help automate the typing piece that i was going to look into such as:

https://github.com/Instagram/MonkeyType that was just shared with me today.  

perhaps that is a more efficient way to move forward?

stravalib/client.py
```diff
@@ -85,15 +90,22 @@ def __init__(
...
-        The currently configured authorization token.
-        """
+        """The currently configured authorization token."""
         return self.protocol.access_token
 
     @access_token.setter
     def access_token(self, v):
-        """
-        Set the currently configured authorization token.
+        """Set the currently configured authorization token.
```
lwasser, line 98: @jsamoocha one other thought i had ... 

there are some instances of a variable like this - i suspect v == value. i thought about making the variables more expressive to changing v -> token_value or token. i suppose if some people have hard coded v then it could break things in multi variable input methods .

stravalib/client.py
```diff
@@ -510,15 +566,20 @@ def get_activity(self, activity_id, include_all_efforts=False):
         )
         return model.Activity.parse_obj({**raw, **{"bound_client": self}})
 
-    def get_friend_activities(self, limit=None):
-        """
-        DEPRECATED This endpoint was removed by Strava in Jan 2018.
+    # TODO: can we remove commented out code in this function?
```
lwasser, line 569: @jsamoocha there is commented code in this method and in another one in this module. do you think we can just delete it or is it a reminder for something else that could be converted to a TODO ?

stravalib/client.py
```diff
@@ -966,15 +1036,22 @@ def get_activity_laps(self, activity_id):
             result_fetcher=result_fetcher,
         )
 
+    # TODO this also has commented code
```
lwasser, line 1039: this is the other method with commented out code @jsamoocha . just cleaning up a bit :)

lwasser:
i ran black on just these two files as well and updated our pre-commit config to the most recent version. i am not sure why precommit is not updating on it's own .

stravalib/client.py
jsamoocha: Yeah, let's try to do one thing at a time (i.e. add the typing in a later PR)

stravalib/client.py
jsamoocha: > i suppose if some people have hard coded v then it could break things in multi variable input methods .

That's a valid concern, but in this case, we're dealing with a single positional argument of a property setter. This will be called as `client.access_token = SOME_TOKEN`, so the risk of breaking stuff is minimal. It makes sense to use a more descriptive variable name here.

stravalib/client.py
jsamoocha: This whole method should be removed as it is no longer part of the Strava API. We should systematically check this sometime; see #323.

So, yes, the commented-out code can be deleted.

jsamoocha:
pre-commit.ci autofix

jsamoocha:
Oops, now I ran pre-commit on the whole code base. I assumed it would only apply changes within the scope of this PR 🙈. Should we revert these changes?

stravalib/client.py
lwasser: oh that's right. i didn't pay attention to the decorator i was just in fix everything mode. ok i'll update.

stravalib/client.py
lwasser: ok i'll remove commented code here and leave a todo that this entire method can be removed. i think given this pr will get big with the black fix, and this is more of a cleanup pr, let's implement the removal of methods separately in another pr.

stravalib/client.py
lwasser: removed and left a todo to remove the method in a future pr

stravalib/client.py
```diff
@@ -1760,53 +1902,79 @@ def next(self):
...
+        * This metadata is only available for partner apps. If you have a
+        regular / non partner related Strava app / account it will not work.
+
         """
 
         warn_attribute_unofficial("photo_metadata")
 
         return self._photo_metadata
 
+    # TODO: what is value here? perhaps an int representing a photo id?
```
lwasser, line 1942: @jsamoocha i had a question about this as well. i suspect photo_metadata would be an `int` representing the photo id value. but i haven't worked with photos at all yet to know for sure so i wanted to ask before updating.

lwasser:
ok! i've made a few more cleanup fixes AND i've removed the typing i started to add. If this looks good we can run pre-commit before merging to implement black throughout the code base. I can then start working on other modules.

stravalib/client.py
lwasser: great - i've removed the typing in there. better to keep each update simple.

stravalib/client.py
```diff
@@ -1566,41 +1687,59 @@ def create_subscription(
...
 
-        :return: The JSON response expected by Strava to the challenge request.
-        :rtype: Dict[str, str]
         """
         callback = model.SubscriptionCallback.deserialize(raw)
         callback.validate(verify_token)
         response_raw = {"hub.challenge": callback.hub_challenge}
         return response_raw
 
+    # TODO: i'm not sure what raw's "type" is here
```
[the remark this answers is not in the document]
jsamoocha: Yes, `raw` is a dict representation of the JSON request. It's good you found this, because it shows a usage of the deprecated `deserialize()` method. I just created #358 as a reminder to get rid of those.

stravalib/client.py
jsamoocha: Here's the code where it's set (which was outside of `__init__()`, so not straightforward to find its type):

```python
# undocumented field, it contains pre-signed uri to upload photo to
self._photo_metadata: Optional[List[Dict]] = response.get(
    "photo_metadata"
)
```

jsamoocha:
One thing I don't understand is why pre-commit.ci runs black on the whole codebase, while pre-commit locally runs its stuff only on the commit.

If pre-commit.ci can only run on the entire codebase, let's apply black on everything, as you suggested.

stravalib/client.py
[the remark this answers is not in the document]
lwasser: ok great! i'll leave a TODO: to mark this! and will update the docstring now.

stravalib/client.py
lwasser: ok i looked at that too quickly. I see the Optional is a typing object now. so i have list of dictionaries or `None` as the type here.

lwasser:
> One thing I don't understand is why pre-commit.ci runs black on the whole codebase, while pre-commit locally runs its stuff only on the commit.
> 
> If pre-commit.ci can only run on the entire codebase, let's apply black on everything, as you suggested.

I hear you. i'm actually looking into this as i have the same question / frustration. There is another tool called [darker](https://github.com/akaihola/darker) that i'm also looking into. But it does seem to be something the bot is doing specifically as you point out. Let me dig a bit more into this to see if we can find a better solution.

lwasser:
Can i run black on everything here and merge this PR? Then i'll continue to look for a better approach separately. 

In theory once we implement black and isort here, we shouldn't run into this issue again with future prs. We can test when I submit the next PR with more docstring updates that builds on top of this pr (assuming your comfortable with all of the docstring updates so far!).

jsamoocha:
> Can i run black on everything here and merge this PR?

Sounds like a great plan. Thanks for the work on these docstrings!

lwasser:
pre-commit.ci autofix

.pre-commit-config.yaml
```diff
@@ -45,8 +48,10 @@ repos:
...
 
   # Black for auto code formatting
-#  - repo: https://github.com/psf/black
-#    rev: 22.12.0
-#    hooks:
-#     - id: black
-#       language_version: python3.8
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
+    stages: [commit]
```
lwasser, line 53:
```suggestion
    rev: 23.3.0
```

lwasser:
pre-commit.ci autofix

lwasser:
pre-commit.ci autofix

lwasser:
pre-commit.ci autofix

lwasser:
I am having a troubles with black and isort overwriting each other. even with the black profile specified for isort. i'm not sure how to fix this other than skipping isort for the time being ?? it's only with model.py

lwasser:
pre-commit.ci autofix

lwasser:
ok many thanks to @pradyun as the issue was black needed to be bumped backwards to work. black 23.3 is new released just a few days ago. isort hasn't had a release since january. I used black 23.1 to sync with the jan release of isort and it's working now. just noting this for any future sync issues that we might have. 

[I have reported the bug here:](https://github.com/PyCQA/isort/issues/2116)

PR #362, opened 2023-04-20
implemented sport_type for client.update_activity

### LINKED ISSUE #360: can I access or set the sub type of an activity?
Hi,

in strava I can set sub types like Mountain Bike (MTB) as sub type of Bike for activities.

In stravalib, activities seem to have type == 'Ride' for example but I didn't find the possibility to access or set sub types like MTB.

Is this missing or did I just not find it?

think-nice-things:
updates:
- [github.com/psf/black: 23.1.0 → 23.3.0](https://github.com/psf/black/compare/23.1.0...23.3.0)

Co-authored-by: pre-commit-ci[bot] <66853113+pre-commit-ci[bot]@users.noreply.github.com>

closes #360

## Description

support sport_type in client.update_activitiy().

From the strava api UpdatableActivity it seems to me that type is deprecated anyway in favour of sport_type.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

I didn't add tests since I'm not familiar with it. I tested and verified the PR with modifiying some sport_type's of my personal strava records though.


jsamoocha (commented):
Thanks, this looks good! Two more things need to happen:

- [x] Modify the test for this client method
- [x] Show a deprecation warning when the old `activity_type` param is used

You indicated you're unfamiliar with the test suite, so I'll do the above.

think-nice-things:
@jsamoocha I added a deprecation warning if using activity_type.
Since both `warn_method_deprecation()` and `warn_param_deprecation()` didn't really match my needs  I created a new one `warn_param_deprecation_in_favor_of()`.

lwasser:
@think-nice-things thank you so much for this pr!!
Can you please help us understand why the existing deprecation methods won't work? There is the option to pass the alternative str to suggest an alternative option (`sport_type` vs `type`) rather than creating a new method that is specific to sport types. Many thanks for clarifying!

```
warnings.warn(
        f'The method "{method_name}" of class "{klass}" is deprecated and '
        f"will be removed in the "
        f'future. Instead, you can use "{alternative}".{alt_support_msg}',
        DeprecationWarning,
        stacklevel=3,
    )
```

think-nice-things:
> @think-nice-things thank you so much for this pr!! Can you please help us understand why the existing deprecation methods won't work? There is the option to pass the alternative str to suggest an alternative option (`sport_type` vs `type`) rather than creating a new method that is specific to sport types. Many thanks for clarifying!
> 
> ```
> warnings.warn(
>         f'The method "{method_name}" of class "{klass}" is deprecated and '
>         f"will be removed in the "
>         f'future. Instead, you can use "{alternative}".{alt_support_msg}',
>         DeprecationWarning,
>         stacklevel=3,
>     )
> ```
> 
> is it because it says "method" and class? i'm just wondering if perhaps we can make the existing deprecation method more flexible in this case vs creating something new. curious what @jsamoocha thinks.

right. the existing `warning_method_deprecation()` talks about a method and class, but sport_type is a parameter and I'm also not sure which class we should talk about here. The message of the other existing `warn_param_deprecation()` is also not applicable here because activity_type (or simply type) is still supported but deprecated.

jsamoocha:
The addition of `warn_param_deprecation_in_favor_of()` is fine. I just renamed it to `warn_param_deprecation()` to have more consistent naming. The previously existing `warn_param_deprecation()` was actually used to warn about unsupported params, so I renamed that as well. I hope the result is less confusing.

@lwasser , @think-nice-things let's merge this if you're ok with the latest changes.

think-nice-things:
fine for me, thanks

lwasser:
Sounds good to me too!! Merge and then let's release 1.3?

lwasser:
Also thank you again for this pr and for the thought that went into it
@think-nice-things !!

PR #378, opened 2023-06-30
Fix: Flake 8 and numpy docstring updates addressing #326

lwasser:
## Description

This pr begins to (finally) address #326 which i started a while back. There are a few small questions related to pep compliance that i'll comment on. we can decide to ignore E731 in the flake8 settings if we wish. I did not update any docstrings i just restructured and cleaned up PEP 8 issues.

## Type of change

- [x] This is a (cleanup) documentation update
- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/stravalib/protocol.py
+++ b/stravalib/protocol.py
@@ -79,15 +79,7 @@ class ApiV3(metaclass=abc.ABCMeta):
         else:
             self.rsession = requests.Session()
 
-        if rate_limiter is None:
-            # Make it a dummy function, so we don't have to check if it's
-            # defined before calling it later
-
-            # TODO: This violates E731 in PEP 8 - are we ok with that?
-            # The preferred syntax here is a def()
-            rate_limiter = lambda x: None
-
-        self.rate_limiter = rate_limiter
+        self.rate_limiter = rate_limiter or (lambda _: None)
 
     def authorization_url(
         self,
--- a/stravalib/protocol.py
+++ b/stravalib/protocol.py
@@ -157,7 +157,7 @@ class ApiV3(metaclass=abc.ABCMeta):
         self, client_id: int, client_secret: str, code: str
     ) -> AccessInfo:
         """Exchange the temporary authorization code (returned with redirect
-        from strava authorization URL) for a short-lived access token and a
+        from Strava authorization URL) for a short-lived access token and a
         refresh token (used to obtain the next access token later on).
 
         Parameters
@@ -237,6 +237,18 @@ class ApiV3(metaclass=abc.ABCMeta):
         return access_info
 
     def resolve_url(self, url: str) -> str:
+        """
+
+        Parameters
+        ----------
+        url : str
+            url string to be be accessed / resolved
+
+        Returns
+        -------
+        str
+            A string representing the full properly formatted (https) url.
+        """
         if not url.startswith("http"):
             url = urljoin(
                 "https://{0}".format(self.server),
@@ -402,7 +414,21 @@ class ApiV3(metaclass=abc.ABCMeta):
         self, url: str, check_for_errors: bool = True, **kwargs: Any
     ) -> Any:
         """Performs a generic GET request for specified params, returning the
-        response."""
+        response.
+
+        Parameters
+        ----------
+        url : str
+            String representing the url to retrieve
+        check-for_errors: bool (default = True)
+            Flag used to raise an error (or not)
+
+        Returns
+        -------
+        dict
+            Performs the request and returns a JSON object deserialized as dict
+
+        """
         referenced = self._extract_referenced_vars(url)
         url = url.format(**kwargs)
         params = dict(
@@ -420,7 +446,22 @@ class ApiV3(metaclass=abc.ABCMeta):
         **kwargs: Any,
     ) -> Any:
         """Performs a generic POST request for specified params, returning the
-        response."""
+        response.
+
+        Parameters
+        ----------
+        url : str
+            Url string to be requested.
+        files: dict
+            Dictionary of file name to file-like objects. Used by _requests
+        check_for_errors: bool
+            Whether to raise an error (or not)
+
+        Returns
+        -------
+            Deserialized request output.
+
+        """
         referenced = self._extract_referenced_vars(url)
         url = url.format(**kwargs)
         params = dict(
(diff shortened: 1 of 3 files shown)


stravalib/protocol.py
```diff
@@ -421,13 +445,12 @@ def post(
...
-        """
-        Performs a generic PUT request for specified params, returning the response.
-        """
+        """Performs a generic PUT request for specified params, returning the
+        response."""
         referenced = self._extract_referenced_vars(url)
         url = url.format(**kwargs)
         params = dict(
-            [(k, v) for k, v in kwargs.items() if not k in referenced]
+            [(k, v) for k, v in kwargs.items() if k not in referenced]
```
lwasser, line 453: pep 8 prefers to use `not in` rather than `if not k in`

lwasser:
this is ready for review! i did PIN pydantic to allow tests and such to run and opened #379 about the new found issues with the 2.0 release

stravalib/protocol.py
```diff
@@ -80,8 +80,11 @@ def __init__(
             self.rsession = requests.Session()
 
         if rate_limiter is None:
-            # Make it a dummy function, so we don't have to check if it's defined before
-            # calling it later
+            # Make it a dummy function, so we don't have to check if it's
+            # defined before calling it later
+
+            # TODO: This violates E731 in PEP 8 - are we ok with that?
```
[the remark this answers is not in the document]
jsamoocha: I added a small commit to address this

stravalib/util/limiter.py
```diff
@@ -418,6 +455,7 @@ def __init__(self) -> None:
             )
         )
 
+        # TODO: can we delete this code given it's commented out?
```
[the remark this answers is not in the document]
jsamoocha: Maybe these lines should be added to the documentation later as examples in a TBD section on rate limiting. Let's keep them for the moment.

jsamoocha (commented):
Thanks, this looks good! I saw at least one docstring (`RateLimitRule.__init__()`) missing types in the docstring, and one method (`RateLimitRule.__call__()`) where the docstring wasn't converted to numpy style.

stravalib/util/limiter.py
```diff
@@ -292,6 +323,9 @@ def _get_wait_time(
         elif self.priority == "low":
             return seconds_until_long_limit / (self.long_limit - long_usage)
 
+    # TODO: pyright returning - Argument of type "float | None" cannot be assigned to parameter "secs" of type "float" in function "sleep"
+    # Type "float | None" cannot be assigned to type "float"
+    # Type "None" cannot be assigned to type
```
lwasser, line 328: @enadeau question about typing here. for me pyright is flagging this suggesting Non can't be assigned. mypy does not pick up on this. Should we address this? I have heard that pyright is a bit noisier than mypy and i could also migrate to mypy in my editor as well.

stravalib/util/limiter.py
```diff
@@ -277,7 +308,7 @@ def _get_wait_time(
         long_usage: int,
         seconds_until_short_limit: int,
         seconds_until_long_limit: int,
-    ) -> float:
+    ) -> float | None:
```
lwasser, line 311: @enadeau i adjusted the return here to add an OR in case it was none. Pyright was again flagging this where mypy wasn't. i'm curious about the difference between these two type checkers and why mypy was ok with this whereas pyright was not. looking at the code it feels unlikely to return None  but perhaps ?

stravalib/protocol.py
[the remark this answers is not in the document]
lwasser: perfect -thanks! Jonatan -your fix looks like this

`self.rate_limiter = rate_limiter or (lambda _: None)` 
which also allowed you to remove another line of code. 

BUT why did you define the empty lamba function as `_` vs just using `(lambda: None)` ?
i'm just curious. i don't use lambda functions a lot but do understand it allows us to create an empty object here . and the or would return whatever has some value in this case OR the empty function.

many thanks!

stravalib/util/limiter.py
enadeau: I've not worked a lot with pyright but it seems like here mypy is outsmarting it because `self.priority` is typed as `Literal["low", "medium", "high"]` mypy know that  you will always return in one of the three blocks of the if statement. Pyrigth do not realize that as it only consider self.priority as being an `str`. Since by default a function return None it says that the return type should be `float | None`

I think we definitely don't want to change the return type to `str | None`. As you noticed, it actually causes problem later down the road when we try to use the result in a function that expect a float.

If you want to silence pyright on this one you can raise a relevant exception after the if statement

stravalib/util/limiter.py
enadeau: This warning is caused by modification you made on the _get_wait_time method

stravalib/util/limiter.py
lwasser: ahhh you are right. i just was listening to the type checker and not closely evaluating the return possibilities. i'm switching to mypy only in my gui. thank you!

stravalib/util/limiter.py
lwasser: ok great - yup i removed that other return type and it looks good now. thanks!!

lwasser:
ok this is ready to go again. a few notes
I did NOT work on parameter docstrings for helpers. i left just a simple description. If we decide we want to fully document helpers I can - but generally if we don't expect users to directly call helpers / private methods i think a short description with type annotations is sufficient for development. let me know if that sounds reasonable!

CI is failing only because of code coverage here! but there are no actual code modifications only docstring updates / cleanup.

I couldn't find a docstring that still uses REST but if you can point out where i missed something i'm happy to fix it!

stravalib/protocol.py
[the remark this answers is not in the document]
jsamoocha: The `_` allows us to assign as `rate_limiter` an (anonymous) function that accepts a single (anonymous) positional argument. Without that argument (i.e., using `lambda : None`), the code using this will try to call that lambda _with_ an argument and cause a TypeError (or fail static type checks otherwise).

To me, the `_` signifies that the argument isn't used. We can also write `lambda x: None` (as in the original code) if that makes more sense to you (although that may raise a PEP warning about declared but unused variables).

Issue #383, opened 2023-07-01
Method to get stream kinda broken.


enadeau:
The method to get the streams (`get_activity_streams`, `get_effort_streams`, `get_segment_streams`) seems to not match the schema from strava https://developers.strava.com/playground/#/Streams/getActivityStreams

The types are supposed to be passed as a query argument and not as part of the path. Also, this argument which they call `key` in the schema is required.

I suggest we remove the default of None for those functions and fix the way we build the request. I can take care of it if you agree with the proposed changes.

enadeau:
Actually, most of the function argument seems to not be relevant anymore. It was also pointed out in #367 maybe I can tackle both at once.

Would you prefer removing the argument or deprecating them?

jsamoocha:
Thanks, this sounds good and relates indeed to #367. For backward compatibility (as the existing undocumented arguments still seem to be supported by Strava), can you please add deprecation warnings when using these arguments (you can probably reuse one of the `warn_...` functions in the `exc` module). It will add some additional branching and may require a few more integration tests; please let me know if you need some help with these.

enadeau:
Hi I've started working on it, but I'm confused on how to write tests

My understanding is the strava api mock uses the example from the open api specification to generate a response. Is that right?

My issue might come from some incoherence in strava's spec. For the get activity stream route. The example from the spec is a list of dict while the model is a dict with stream type as key. I also get something matching the model when I make an actual call. https://developers.strava.com/playground/#/Streams/getActivityStreams

The mock is currently returning a list. What would be the best way to get the actual right format out of the mock?

enadeau:
I've pushed the code I've wrote in case it helps you. Should I use response_update to provide a response that matches strava model?

jsamoocha:
> My understanding is the strava api mock uses the example from the open api specification to generate a response. Is that right?

Yes, that is the default. With `response_update` you can update the example from the OpenAPI spec (just as you update a plain Python dict). Alternatively, you can use the `json` or `body` keywords to totally ignore and overwrite the example from the OpenAPI spec.

jsamoocha:
> My issue might come from some incoherence in strava's spec. For the get activity stream route. The example from the spec is a list of dict while the model is a dict with stream type as key. I also get something matching the model when I make an actual call.

I understand your confusion :)

We deviated from the "official" model for reasons of backward compatibility. This final snippet of `get_activity_streams()` is the culprit:

```python
# Pack streams into dictionary
try:
[...]

Issue #373, opened 2023-06-24
Working towards better type annotation


enadeau:
Hi I've been using the package and I quite like it.

I'd be interested in contributing to the library by adding up type hints because it's always nice to have good more confidence in code correctness and more autocomplete for the users.

If that's something you think is a good idea, I'd try to work in small chunks. I think the best way to do would be to start with a first PR that sets up configure mypy and setup CI. The goal would be to get that in a working state with minimal changes.  Probably needs to fix a few types and maybe add typing to one file or two. Then I could incrementally add types and check more files in subsequent PRs.

jsamoocha:
Hi, this sounds like a great idea! Looking forward to PRs.

lwasser:
I also think this is a wonderful idea!! Welcome to stravalib!!

I'd suggest we discuss mypy prior to setting it up. Essentially with the
precommit setup that we have and the bot it will potentially fail all prs.

If you ignore it in a ci space then it's easy to always ignore typing
issues.

I suggest starting w small prs that address typing issues and work towards
the 100% that mypy might want across our code base.

There is a bit of (older) discussion about this below. But I think just
running mypy independently at first might be best as we build out typing
throughout . Strict typing for an entire code base will take a good bit of
time to implement!

https://stackoverflow.com/questions/59730765/is-it-possible-to-run-mypy-pre-commit-without-making-it-fail

What do you think about that approach?

Leah

enadeau:
I've read that getting mypy to working pre-commit is problematic. The best approach is probably to have it as another workflow that is not part of pre-commit.

It could eventually be it's on github action/workflow but it doesn't need to happen right now.

The progressive approach seems right to me. I'd try to add type hints it one file at a time, activate mypy on the files that I've fixed.

enadeau:
I was thinking about all the type annotation/description that are present in the docstring. They are quite useful to write the type annotation, but they become kind of redundant once the code is annotated and can easily be out of sync. I see mostly three options.

1. Remove them as the funcitons get annotated
2. Remove them once the whole code base is annotated
3. Just keep them

I think I'd go for 1 or 2. Removing them immediately has the advantage to not need to make another pass through the file while keeping while we complete the typing will make it easier to refer back as needed

lwasser:
let me dig into that a hit more. this issue becomes we use autodoc that filters into sphinx. so i need to understand how that would impact our documentation. 

[...]

PR #384, opened 2023-07-02
Typing client file

enadeau:
## Description

Add type annotations to the client file. See #373

Still need to fix a couple small typing error and would be best to address #383
first as it relates to the test that are now failling.

## Type of change

- [x] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -296,7 +296,7 @@ class Client:
 
         before_epoch = self._utc_datetime_to_epoch(before) if before else None
         after_epoch = self._utc_datetime_to_epoch(after) if after else None
-        params = dict(before=before, after=after)
+        params = dict(before=before_epoch, after=after_epoch)
         result_fetcher = functools.partial(
             self.protocol.get, "/athlete/activities", **params
         )
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1570,7 +1570,7 @@ class Client:
         effort_id: int,
         types: list[StreamType] | None = None,
         resolution: Literal["low", "medium", "high"] | None = None,
-        series_type: Literal["time", "distance"] = "distance",
+        series_type: Literal["time", "distance"] | None = None,
     ) -> dict[StreamType, model.Stream]:
         """Returns an streams for an effort.
 
@@ -1624,7 +1624,7 @@ class Client:
         segment_id: int,
         types: list[StreamType] | None = None,
         resolution: Literal["low", "medium", "high"] | None = None,
-        series_type: Literal["time", "distance"] = "distance",
+        series_type: Literal["time", "distance"] | None = None,
     ) -> dict[StreamType, model.Stream]:
         """Returns an streams for a segment.
 
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -17,17 +17,12 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Deque,
-    Dict,
     Generic,
     Iterable,
-    List,
     Literal,
     NoReturn,
-    Optional,
     Protocol,
-    Type,
     TypeVar,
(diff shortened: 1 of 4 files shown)


enadeau:
Back from summer holiday! Just finished typing the client file. Sorry for the big diff but at least after that we will have most of the outward facing method typed I think.

stravalib/client.py
```diff
@@ -263,12 +294,8 @@ def get_activities(self, before=None, after=None, limit=None):
...
 
-        if before:
-            before = self._utc_datetime_to_epoch(before)
-
-        if after:
-            after = self._utc_datetime_to_epoch(after)
-
+        before_epoch = self._utc_datetime_to_epoch(before) if before else None
+        after_epoch = self._utc_datetime_to_epoch(after) if after else None
         params = dict(before=before, after=after)
```
jsamoocha, line 299: I don't see where `before_epoch` or `after_epoc` are used. Can it be that these were meant to be set as part of the `params` dict?

stravalib/client.py
```diff
@@ -1305,15 +1370,20 @@ def get_segment_efforts(
         )
 
     def explore_segments(
-        self, bounds, activity_type=None, min_cat=None, max_cat=None
-    ):
+        self,
+        bounds: tuple[float, float, float, float]
```
jsamoocha, line 1374: I'm curious why this must be a tuple (instead of any indexable sequence (which I wouldn't know how to type, BTW)).

stravalib/client.py
```diff
@@ -1909,23 +2022,23 @@ def _fill_buffer(self):
...
         return self.next()
 
-    def next(self):
-        """ """
+    def next(self) -> T:
         if self.limit and self._counter >= self.limit:
             self._eof()
         if not self._buffer:
             self._fill_buffer()
+        assert isinstance(self._buffer, collections.deque)
```
jsamoocha, line 2041: Again, out of curiosity: why did you add this assertion?

stravalib/client.py
```diff
@@ -1815,7 +1920,15 @@ def delete_subscription(self, subscription_id, client_id, client_secret):
...
-class BatchedResultsIterator(object):
+T = TypeVar("T", bound=BaseModel)
+
+
+class ResultFetcher(Protocol):
+    def __call__(self, *, page: int, per_page: int) -> Iterable[Any]:
+        ...
+
+
+class BatchedResultsIterator(Generic[T]):
```
jsamoocha, line 1931: It's great that you turned this into a generic class 👍

jsamoocha (commented):
Thanks a lot for this PR! Before merging, I'd like to be sure we're not changing any behavior regarding the `before` and `after` parameters I mentioned in a specific comment.

stravalib/client.py
enadeau: Could catch they should be passed as param. I'll fix that

stravalib/client.py
enadeau: The tuple allows to incorporate the fact that there is exactly for float in the sequence, so that's why I chose it. Note that the function with still work if call with the list but the mypy will complain.

For a general indexable sequence you can use `typing.Sequence[float]`

stravalib/client.py
enadeau: That really helps get good typing of the method that uses it 🙂

stravalib/client.py
enadeau: It tells mypy that here buffer cannot be None. I've actually changed to assert it is not None to make it clearer.

enadeau:
I think it is now ready to go :100:

jsamoocha (approved):
Awesome, thanks for the explanations and fixes!

PR #385, opened 2023-07-02
Fix stream methods

### LINKED ISSUE #367: Doc Update required for get_activity_streams()
The documentation for get_activity_streams() at [https://stravalib.readthedocs.io/en/latest/reference/api/stravalib.client.Client.get_activity_streams.html](https://stravalib.readthedocs.io/en/latest/reference/api/stravalib.client.Client.get_activity_streams.html) seems not accurate any longer. The resolution 'all' seems not to exist any longer, using it explicitly results in a Bad Request. In contrast, 'high' seems to have a higher resolution than 10.000 points as I received a response with 11.500 points having a resolution 'high'.

### LINKED ISSUE #383: Method to get stream kinda broken.

enadeau:
closes #383 and closes #367

## Description

Update the client's methods to get stream to match the current strata API.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

### DIFF
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -18,7 +18,7 @@ from typing import Dict, List, Optional
 import arrow
 import pytz
 
-from stravalib import exc, model, unithelper
+from stravalib import exc, model, strava_model, unithelper
 from stravalib.exc import (
     ActivityPhotoUploadNotSupported,
     warn_attribute_unofficial,
@@ -30,20 +30,6 @@ from stravalib.exc import (
 from stravalib.protocol import ApiV3
 from stravalib.unithelper import is_quantity_type
 from stravalib.util import limiter
-
-VALID_STREAM_TYPE = [
-    "time",
-    "latlng",
-    "distance",
-    "altitude",
-    "velocity_smooth",
-    "heartrate",
-    "cadence",
-    "watts",
-    "temp",
-    "moving",
-    "grade_smooth",
-]
 
 
 class Client(object):
@@ -1398,7 +1384,7 @@ class Client(object):
         activity_id : int
             The ID of activity.
         types : list[str], optional, default=None
-            A list of the types of streams to fetch. If not is specified all
+            A list of the types of streams to fetch.
         resolution :
             .. deprecated
                 This param is not supported by the Strava API and may be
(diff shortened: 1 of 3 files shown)


stravalib/client.py
```diff
@@ -1383,54 +1397,46 @@ def get_activity_streams(
...
-        # Pack streams into dictionary
-        try:
-            return {i.type: i for i in streams}
-        except exc.ObjectNotFound:
-            return None  # just to be explicit.
+        print(response)
+        return {
+            model.Stream.parse_obj(stream)
+            for stream_type, stream in response.json().items()
+        }
```
jsamoocha, line 1439: This is much cleaner than the existing approach but is a breaking change.

stravalib/client.py
```diff
@@ -31,6 +31,20 @@
...
+    "altitude",
+    "velocity_smooth",
+    "heartrate",
+    "cadence",
+    "watts",
+    "temp",
+    "moving",
+    "grade_smooth",
+]
+
```
jsamoocha, line 47: We can use `strava_model.StreamType.schema()['enum']` to get these values from the OpenAPI spec.

stravalib/client.py
enadeau: My bad I intended to do
```py
return {
    stream_type: model.Stream.parse_obj(stream)
    for stream_type, stream in response.json().items()
}
```
but I should probably also get the None back if an exception is raised

stravalib/client.py
enadeau: Oh that's really nice.

enadeau:
I've written some test to bring the coverage to 100% on the method.

I think it is ready for review

The `get_effrot_streams` and `get_segments_streams` still need to be modified, but the change would be identical to the one I did for `get_activity_streams` so best to get the review in first.

jsamoocha (commented):
Thanks for this fix and the added test coverage! Because Strava still supports these undocumented parameters, we cannot simply remove them, as that would change the behavior of our API.

Besides that, your overall approach of using the documented query params for the stream types instead of the path param and getting rid of the unnecessary `BatchedResultsIterator` is good and can be applied for other endpoints involving streams.

enadeau:
Applied your suggestions and also did the effort and segment method. 

I've refactored the code to avoid having the same logic 3 time.

Ready for review

jsamoocha (approved):
Thanks a lot, the refactoring and added test coverage are great improvements!

PR #387, opened 2023-07-04
Fix:  model.py docstrings and typing

### LINKED ISSUE #375: Warning when building the documentation
When trying to build the documentation, three warnings are raised and they all translate to minore issue in the doc display.

- [x] [XRateLimiteRule `__init__` function](https://stravalib.readthedocs.io/en/latest/reference/api/stravalib.util.limiter.XRateLimitRule.html) the description of the force limits param ends up on an other line

- [x] [Client.upload_activity](https://stravalib.readthedocs.io/en/latest/reference/api/stravalib.client.Client.upload_activity.html) the display message of description for the deprecated param is not in the deprecated box
- [ ] [BaseEffort](https://stravalib.readthedocs.io/en/latest/reference/model.html#stravalib.model.BaseEffort) model json schema is not computed correctly for the activity field.

I have a pull request ready for first and second error but I'm not quite sure about 3 so I decided to open an issue about it.

lwasser:
closes #375 

This should fix 375 - the warning when building docs

## Description

This is a small issue
1. if docs are missing docstrings they won't appear at all in the documentation.
2. If a class inherits from another class but is missing a pass, it also can't render properly making it difficult for pydantic-autodoc to properly render the expected JSON.

This fixes that issue in model.py and also addresses a few other docstring related items.

## Type of change

- [x ] This is a documentation update
- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,7 +6,8 @@
 - Update client's stream method to warn when using unofficial parameters (@enadeau, #385)
 - Type annotation to client file (@enadeau, #384)
 - Fix docstring in SleepingRateLimitRule (@enadeau)
-- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #394)
+- Fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #394)
+- Fix: docstrings in model.py, documentation errors, findfonts warning suppression by removing opengraph (temporarily), typing (@lwasser)
 
 ## v1.3.3
 
--- a/changelog.md
+++ b/changelog.md
@@ -7,7 +7,7 @@
 - Type annotation to client file (@enadeau, #384)
 - Fix docstring in SleepingRateLimitRule (@enadeau)
 - Fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #394)
-- Fix: docstrings in model.py, documentation errors, findfonts warning suppression by removing opengraph (temporarily), typing (@lwasser)
+- Fix: docstrings in model.py, documentation errors, findfonts warning suppression by removing opengraph (temporarily), typing updates (@lwasser, #387)
 
 ## v1.3.3
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -113,7 +113,7 @@ html_theme_options = {
     "show_toc_level": 1,
     # "navbar_align": "left",  # [left, content, right] For testing that the navbar items align properly
     "github_url": "https://github.com/stravalib/stravalib",
-    "footer_items": ["copyright"],
+    "footer_start": ["copyright"],
 }
 
 html_context = {
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,7 +63,7 @@ extensions = [
     "sphinx.ext.autosummary",
     "myst_nb",
     "sphinx_design",
-    "sphinxext.opengraph",
+    # "sphinxext.opengraph",
     "sphinx_inline_tabs",
 ]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,6 +63,8 @@ extensions = [
     "sphinx.ext.autosummary",
     "myst_nb",
     "sphinx_design",
+    # Commented out because matplotlib raises a findfonts warning over and over
+    # TODO: fix is to install fonts in CI or remove opengraph
     # "sphinxext.opengraph",
     "sphinx_inline_tabs",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,7 @@ files = [
     "stravalib/protocol.py",
     "stravalib/exc.py",
     "stravalib/client.py",
+    "stravalib/model.py",
 ]
 
 [tool.pydantic-mypy]
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -21,6 +21,7 @@ from typing import (
     Iterable,
     Literal,
     NoReturn,
+    Optional,
     Protocol,
     TypeVar,
     cast,
@@ -1822,7 +1823,7 @@ class Client:
         self,
         raw: dict[str, Any],
         verify_token: str = model.Subscription.VERIFY_TOKEN_DEFAULT,
-    ) -> dict[str, str]:
+    ) -> dict[str, Optional[str]]:
         """Validate callback request and return valid response with challenge.
 
         Parameters
@@ -1833,7 +1834,7 @@ class Client:
 
         Returns
         -------
-        Dict[str, str]
+        dict[str, str]
             The JSON response expected by Strava to the challenge request.
 
         """
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -21,7 +21,6 @@ from typing import (
     Iterable,
     Literal,
     NoReturn,
-    Optional,
     Protocol,
     TypeVar,
     cast,
@@ -1823,7 +1822,7 @@ class Client:
         self,
         raw: dict[str, Any],
         verify_token: str = model.Subscription.VERIFY_TOKEN_DEFAULT,
-    ) -> dict[str, Optional[str]]:
+    ) -> dict[str, str]:
         """Validate callback request and return valid response with challenge.
 
         Parameters
@@ -1840,10 +1839,11 @@ class Client:
         """
         callback = model.SubscriptionCallback.deserialize(raw)
         callback.validate_token(verify_token)
+
+        assert callback.hub_challenge is not None
         response_raw = {"hub.challenge": callback.hub_challenge}
         return response_raw
 
-    # TODO: i'm not sure what raw's "type" is here
     def handle_subscription_update(
         self, raw: dict[str, Any]
     ) -> model.SubscriptionUpdate:
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1471,13 +1471,13 @@ class Client:
             Indicates desired number of data points. 'low' (100), 'medium'
             (1000) or 'high' (10000).
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
+                This param is not officially supported by the Strava API and may be
                 removed in the future.
         series_type : str, optional
             Relevant only if using resolution either 'time' or 'distance'.
             Used to index the streams if the stream is being reduced.
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
+                This param is not officially supported by the Strava API and may be
                 removed in the future.
 
         Returns
@@ -1532,7 +1532,7 @@ class Client:
         results.
 
         Streams types are: time, latlng, distance, altitude, velocity_smooth,
-                           heartrate, cadence, watts, temp, moving, grade_smooth
+                            heartrate, cadence, watts, temp, moving, grade_smooth
 
         Parameters
         ----------
@@ -1544,19 +1544,19 @@ class Client:
             Indicates desired number of data points. 'low' (100), 'medium'
             (1000) or 'high' (10000).
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
-                removed in the future.
+                This param is not officially supported by the Strava API and
+                may be removed in the future.
         series_type : str, optional
             Relevant only if using resolution either 'time' or 'distance'.
             Used to index the streams if the stream is being reduced.
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
+                This param is not officially supported by the Strava API and may be
                 removed in the future.
 
         Returns
         -------
         py:class:`dict`
-            An dictionary of :class:`stravalib.model.Stream` from the activity
+            A dictionary of :class:`stravalib.model.Stream` from the activity
         """
         return self._get_streams(
             f"/activities/{activity_id}/streams",
@@ -1597,14 +1597,17 @@ class Client:
             Indicates desired number of data points. 'low' (100), 'medium'
             (1000) or 'high' (10000).
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
-                removed in the future.
+
+                This param is not officially supported by the Strava API and
+                may be removed in the future.
+
         series_type : str, optional
             Relevant only if using resolution either 'time' or 'distance'.
             Used to index the streams if the stream is being reduced.
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
-                removed in the future.
+
+                This param is not officially supported by the Strava API and
+                may be removed in the future.
 
         Returns
         -------
@@ -1639,7 +1642,7 @@ class Client:
         results.
 
         Streams types are: time, latlng, distance, altitude, velocity_smooth,
-                           heartrate, cadence, watts, temp, moving, grade_smooth
+        heartrate, cadence, watts, temp, moving, grade_smooth
 
         Parameters
         ----------
@@ -1651,13 +1654,13 @@ class Client:
             Indicates desired number of data points. 'low' (100), 'medium'
             (1000) or 'high' (10000).
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
+                This param is not officially supported by the Strava API and may be
                 removed in the future.
         series_type : str, optional
             Relevant only if using resolution either 'time' or 'distance'.
             Used to index the streams if the stream is being reduced.
             .. deprecated::
-                This param is not officialy supported by the Strava API and may be
+                This param is not officially supported by the Strava API and may be
                 removed in the future.
 
         Returns
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -19,7 +19,6 @@ from typing import (
     Deque,
     Generic,
     Iterable,
-    List,
     Literal,
     NoReturn,
     Protocol,
@@ -960,7 +959,7 @@ class Client:
 
     def get_activity_zones(
         self, activity_id: int
-    ) -> List[model.BaseActivityZone]:
+    ) -> list[model.BaseActivityZone]:
         """Gets zones for activity.
 
         Requires premium account.
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -19,6 +19,7 @@ from typing import (
     Deque,
     Generic,
     Iterable,
+    List,
     Literal,
     NoReturn,
     Protocol,
@@ -959,7 +960,7 @@ class Client:
 
     def get_activity_zones(
         self, activity_id: int
-    ) -> list[model.BaseActivityZone]:
+    ) -> List[model.BaseActivityZone]:
         """Gets zones for activity.
 
         Requires premium account.
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -1051,6 +1051,7 @@ class Activity(
     splits_metric: Optional[List[Split]] = None
     splits_standard: Optional[List[Split]] = None
     photos: Optional[ActivityPhotoMeta] = None
+    # TODO Detailed activity has this defined as - Optional[List[Lap]]
     laps: Optional[List[ActivityLap]] = None
 
     # Added for backward compatibility
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -251,7 +251,7 @@ class DeprecatedSerializableMixin(BaseModel):
         """
         # TODO: Argument 1 to "warn_method_deprecation" has incompatible type "str"; expected "type
         exc.warn_method_deprecation(
-            self.__class__.__name__,
+            self.__class__
             "from_dict()",
             "parse_obj()",
             "https://docs.pydantic.dev/usage/models/#helper-functions",
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -16,6 +16,7 @@ import logging
 from datetime import date, datetime
 from functools import wraps
 from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     ClassVar,
@@ -33,7 +34,6 @@ from pydantic.datetime_parse import parse_datetime
 
 from stravalib import exc
 from stravalib import unithelper as uh
-from stravalib.client import BatchedResultsIterator
 from stravalib.field_conversions import enum_value, enum_values, time_interval, timezone
 from stravalib.strava_model import (
     ActivityStats,
@@ -64,6 +64,9 @@ from stravalib.strava_model import (
     SummarySegmentEffort,
     TimedZoneRange,
 )
+
+if TYPE_CHECKING:
+    from stravalib.client import BatchedResultsIterator
 
 LOGGER = logging.getLogger(__name__)
 
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -69,8 +69,8 @@ if TYPE_CHECKING:
 LOGGER = logging.getLogger(__name__)
 
 
-# TODO: stravalib/model.py:71: error: Please use "Callable[[<parameters>], <return type>]"  [misc]
-def lazy_property(fn: Callable[Any]) -> property:
+# TODO: how do you properly type this with more precision than Any?
+def lazy_property(fn: Callable[Any], Any) -> property:
     """
     Should be used to decorate the functions that return a lazily loaded
     entity (collection), e.g., the members of a club.
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -156,6 +156,7 @@ def check_valid_location(
 # Create alias for this type so docs are more user-readable
 AllDateTypes = NewType("AllDateTypes", Union[datetime, str, bytes, int, float])
 
+
 def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
     """Utility helper that parses a datetime value provided in
     JSON, string, int or other formats and returns a datetime.datetime
@@ -258,13 +259,13 @@ class DeprecatedSerializableMixin(BaseModel):
         ----------
             The `to_dict()` method is deprecated in favor of `dict()` method.
             For more details, refer to the Pydantic documentation:
-            https://docs.pydantic.dev/usage/exporting_models/
+            https://docs.pydantic.dev/1.10/usage/exporting_models/
         """
         warn_method_deprecation(
             self.__class__.__name__,
             "to_dict()",
             "dict()",
-            "https://docs.pydantic.dev/usage/exporting_models/",
+            "https://docs.pydantic.dev/1.10/usage/exporting_models/",
         )
         return self.dict()
 
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -13,6 +13,7 @@ related entities from the API.
 from __future__ import annotations
 
 import logging
+import warnings
 from datetime import date, datetime
 from functools import wraps
 from typing import (
@@ -358,9 +359,7 @@ class LatLon(LatLng, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     # this error doesn't make sense as the types are correct based on what the
     # error says it wants to see. so why is it throwing the error?
     @root_validator
-    def check_valid_latlng(
-        cls, values: List[Optional[float]]
-    ) -> Optional[List[float]]:
+    def check_valid_latlng(cls, values: List[float]) -> Optional[List[float]]:
         """Validate that Strava returned an actual lat/lon rather than an empty
         list. If list is empty, return None
 
@@ -601,10 +600,12 @@ class Athlete(
             The string representation of the athlete type.
         """
 
-        athlete_type = {0: "cyclist", 1: "runner"}
-        if raw_type in [0, 1]:
-            return athlete_type.get(raw_type)
+        if raw_type == 0:
+            return "cyclist"
+        elif raw_type == 1:
+            return "runner"
         else:
+            warnings.warn(f"Unknown athlete type value: {raw_type}")
             return None
 
     @lazy_property
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -613,12 +613,12 @@ class Athlete(
     @lazy_property
     def authenticated_athlete(self) -> Athlete:
         """
-        Returns an `stravalib.Client` object containing Strava account data
+        Returns an `Athlete` object containing Strava account data
         for the (authenticated) athlete.
 
         Returns
         -------
-        DetailedAthlete
+        Athlete
             The detailed information of the authenticated athlete.
         """
         assert self.bound_client is not None, "Bound client is not set."
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -23,6 +23,7 @@ from typing import (
     Literal,
     Optional,
     Tuple,
+    TypeVar,
     Union,
     get_args,
 )
@@ -68,9 +69,11 @@ if TYPE_CHECKING:
 
 LOGGER = logging.getLogger(__name__)
 
+T = TypeVar("T")
+U = TypeVar("U", bound="BoundClientEntity")
 
-# TODO: how do you properly type this with more precision than Any?
-def lazy_property(fn: Callable[[Any], Any]) -> property:
+
+def lazy_property(fn: Callable[[U], T]) -> Optional[T]:
     """
     Should be used to decorate the functions that return a lazily loaded
     entity (collection), e.g., the members of a club.
@@ -99,13 +102,13 @@ def lazy_property(fn: Callable[[Any], Any]) -> property:
     """
 
     @wraps(fn)
-    def wrapper(obj: Any) -> Any:
+    def wrapper(obj: U) -> Optional[T]:
         try:
             if obj.bound_client is None:
                 raise exc.UnboundEntity(
                     f"Unable to fetch objects for unbound {obj.__class__} entity."
                 )
-            if obj.id is None:
+            if obj.id is None:  # type: ignore[attr-defined]
                 LOGGER.warning(
                     f"Cannot retrieve {obj.__class__}.{fn.__name__}, self.id is None"
                 )
@@ -116,7 +119,7 @@ def lazy_property(fn: Callable[[Any], Any]) -> property:
                 f"Unable to fetch objects for unbound {obj.__class__} entity: {e}"
             )
 
-    return property(wrapper)
+    return property(wrapper)  # type: ignore[return-value]
 
 
 # Custom validators for some edge cases:
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -32,8 +32,9 @@ from typing import (
 
 from pydantic import BaseModel, Field, root_validator, validator
 from pydantic.datetime_parse import parse_datetime
+from typing_extensions import Self
 
-from stravalib import exc
+from stravalib import exc, strava_model
 from stravalib import unithelper as uh
 from stravalib.field_conversions import enum_value, enum_values, time_interval, timezone
 from stravalib.strava_model import (
@@ -58,7 +59,6 @@ from stravalib.strava_model import (
 )
 from stravalib.strava_model import Route as RouteStrava
 from stravalib.strava_model import (
-    Split,
     SportType,
     SummaryClub,
     SummaryGear,
@@ -201,9 +201,7 @@ class DeprecatedSerializableMixin(BaseModel):
     # TODO: i used the output type from this class
     # could dict also be an int or float??
     @classmethod
-    def deserialize(
-        cls, attribute_value_mapping: Dict[str, str]
-    ) -> DeprecatedSerializableMixin:
+    def deserialize(cls, attribute_value_mapping: dict[str, str]) -> Self:
         """
         Creates and returns a new object based on serialized (dict) struct.
 
@@ -234,8 +232,7 @@ class DeprecatedSerializableMixin(BaseModel):
         )
         return cls.parse_obj(attribute_value_mapping)
 
-    # TODO: Same question here about dict elements (strings or str, Union[str,int])
-    def from_dict(self, attribute_value_mapping: Dict[str, str]) -> None:
+    def from_dict(self, attribute_value_mapping: dict[str, str]) -> None:
         """
         Deserializes v into self, resetting and ond/or overwriting existing
         fields
@@ -252,7 +249,7 @@ class DeprecatedSerializableMixin(BaseModel):
             For more details, refer to the Pydantic documentation:
             https://docs.pydantic.dev/usage/models/#helper-functions
         """
-        # TODO: Argument 1 to "warn_method_deprecation" has incompatible type "str"; expected "type
+
         exc.warn_method_deprecation(
             self.__class__,
             "from_dict()",
@@ -261,10 +258,7 @@ class DeprecatedSerializableMixin(BaseModel):
         )
         # Ugly hack is necessary because parse_obj does not behave in-place but
         # returns a new object
-        # TODO - this hack makes tests fail -  but makes mypy happy
-        # new_obj = parse_obj_as(self.__class__, attribute_value_mapping)
-        # self.__dict__.update(new_obj.__dict__)
-        self.__init__(**self.parse_obj(attribute_value_mapping).dict())
+        self.__init__(**self.parse_obj(attribute_value_mapping).dict())  # type: ignore
 
     def to_dict(self) -> Dict[str, str]:
         """
@@ -547,8 +541,6 @@ class Athlete(
 
     # Field overrides from superclass for type extensions:
     clubs: Optional[List[SummaryClub]] = None
-    # These overrides support legacy model behavior which will be deprecated
-    # for now we can ignore the type until we implement the deprecation behavior.
     bikes: Optional[List[SummaryGear]] = None
     shoes: Optional[List[SummaryGear]] = None
 
@@ -670,7 +662,6 @@ class Athlete(
 
 
 # TODO: better description
-# TODO: What is the goal of this attr override?
 class ActivityComment(Comment):
     """Field overrides the athlete attribute to be of type :class:`Athlete`
     rather than :class:`SummaryAthlete`
@@ -759,7 +750,6 @@ class ActivityPhoto(BackwardCompatibilityMixin, DeprecatedSerializableMixin):
         )
 
 
-# TODO: this wasn't working as intended - test again
 class ActivityKudos(Athlete):
     """
     Represents kudos an athlete received on an activity.
@@ -810,8 +800,10 @@ class Map(PolylineMap):
 
 # TODO -opening an issue about perhaps a different approach to adding these attrs
 # to split()
-class Split(  # type: ignore
-    Split,
+
+
+class Split(
+    strava_model.Split,
     BackwardCompatibilityMixin,
     DeprecatedSerializableMixin,
 ):
@@ -1051,18 +1043,17 @@ class Activity(
     end_latlng: Optional[LatLon] = None
     map: Optional[Map] = None
     gear: Optional[Gear] = None
-    # TODO: Incompatible types in assignment (expression has type "Optional[List[BestEffort]]", base class "DetailedActivity" defined the type as "Optional[List[DetailedSegmentEffort]]")
     best_efforts: Optional[List[DetailedSegmentEffort]] = None
     segment_efforts: Optional[List[DetailedSegmentEffort]] = None
-    splits_metric: Optional[List[Split]] = None
-    splits_standard: Optional[List[Split]] = None
+    # Ignoring types here given there are overrides
+    splits_metric: Optional[List[Split]] = None  # type: ignore
+    splits_standard: Optional[List[Split]] = None  # type: ignore
     photos: Optional[ActivityPhotoMeta] = None
     # TODO Detailed activity has this defined as - Optional[List[Lap]]
     laps: Optional[List[Union[ActivityLap, Lap]]] = None
 
     # Added for backward compatibility
     # TODO maybe deprecate?
-    # TODO: Missing type parameters for generic type "Tuple"
     TYPES: ClassVar[Tuple[Any, ...]] = get_args(
         ActivityType.__fields__["__root__"].type_
     )
@@ -1118,12 +1109,6 @@ class Activity(
         assert self.bound_client is not None
         return self.bound_client.get_activity_comments(self.id)
 
-    # TODO: question - this returns a list of model.BaseActivityZone objects
-    # that class is in this module associated with the bound client.
-    # The conditional added her addresses this
-    # "error: Item "None" of "Optional[Any]" has no attribute "get_activity_zones"  [union-attr]"
-    # This error appears several times. i just don't understand how
-    # bound client could be None OR do i need to use Optionallist[BaseActivityZone] ??
     @lazy_property
     def zones(self) -> List[BaseActivityZone]:
         """Retrieve a list of zones for an activity.
@@ -1205,8 +1190,6 @@ class Route(
     # Superclass field overrides for using extended types
     athlete: Optional[Athlete] = None
     map: Optional[Map] = None
-    # TODO: base class Route defines this as a list of SummarySegments
-    # Line 1373 in strava_model.py - are we intentionally overriding?
     segments: Optional[List[Segment]]  # type: ignore
 
     _field_conversions = {"distance": uh.meters, "elevation_gain": uh.meters}
@@ -1239,16 +1222,10 @@ class SubscriptionCallback(
     Represents a Webhook Event Subscription Callback.
     """
 
-    hub_mode: Optional[str] = None
-    hub_verify_token: Optional[str] = None
-    hub_challenge: Optional[str] = None
+    hub_mode: Optional[str] = Field(None, alias="hub.mode")
+    hub_verify_token: Optional[str] = Field(None, alias="hub.verify_token")
+    hub_challenge: Optional[str] = Field(None, alias="hub.challenge")
 
-    # TODO: error: Required dynamic aliases disallowed  [pydantic-alias]
-    class Config:
-        alias_generator = lambda field_name: field_name.replace("hub_", "hub.")
-
-    # TODO: Signature of "validate" incompatible with supertype "BaseModel"
-    # [override] - we should consider renaming this method??
     def validate_token(
         self, verify_token: str = Subscription.VERIFY_TOKEN_DEFAULT
     ) -> None:
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -137,8 +137,22 @@ def check_valid_location(
 # Create alias for this type so docs are more user-readable
 AllDateTypes = NewType("AllDateTypes", Union[datetime, str, bytes, int, float])
 
-
 def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
+    """Utility helper that parses a datetime value provided in
+    JSON, string, int or other formats and returns a datetime.datetime
+    object
+
+    Parameters
+    ----------
+    value : str, int
+        A value representing a date/time that may be presented in string,
+        int, deserialized or other format.
+
+    Returns
+    -------
+    datetime.datetime
+        A datetime object representing the datetime input value.
+    """
     if value:
         dt = parse_datetime(value)
         return dt.replace(tzinfo=None)
@@ -227,7 +241,10 @@ class BackwardCompatibilityMixin:
         return value
 
 
+# TODO: add better description of this
 class BoundClientEntity(BaseModel):
+    """Description of this class goes here."""
+
     # Using Any as type here to prevent catch-22 between circular import and
     # pydantic forward-referencing issues "resolved" by PEP-8 violations.
     # See e.g. https://github.com/pydantic/pydantic/issues/1873
@@ -417,13 +434,18 @@ class Athlete(
         return self.is_authenticated
 
 
+# TODO: better description
 class ActivityComment(Comment):
-    # Field overrides from superclass for type extensions:
+    """Field overrides from superclass for type extensions"""
+
     athlete: Optional[Athlete] = None
 
 
 class ActivityPhotoPrimary(Primary):
-    # Undocumented attributes:
+    """The primary photo for an activity.
+    Attributes are Undocumented
+    """
+
     use_primary_photo: Optional[bool] = None
 
 
@@ -533,6 +555,8 @@ class ActivityLap(
 
 
 class Map(PolylineMap):
+    """Pass through object. Inherits from PolyLineMap"""
+
     pass
 
 
@@ -713,7 +737,7 @@ class BaseEffort(
     Base class for a best effort or segment effort.
     """
 
-    # field overrides from superclass for type extensions:
+    # Field overrides from superclass for type extensions:
     segment: Optional[Segment] = None
     activity: Optional[Activity] = None
     athlete: Optional[Athlete] = None
@@ -733,6 +757,8 @@ class BestEffort(BaseEffort):
     """
     Class representing a best effort (e.g. best time for 5k)
     """
+
+    pass
 
 
 class SegmentEffort(BaseEffort):
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -20,6 +20,7 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
+    Float,
     List,
     Literal,
     NewType,
@@ -34,7 +35,7 @@ from pydantic.datetime_parse import parse_datetime
 
 from stravalib import exc
 from stravalib import unithelper as uh
-from stravalib.exc import warn_method_deprecation
+from stravalib.client import BatchedResultsIterator
 from stravalib.field_conversions import enum_value, enum_values, time_interval, timezone
 from stravalib.strava_model import (
     ActivityStats,
@@ -211,7 +212,7 @@ class DeprecatedSerializableMixin(BaseModel):
 
 
         """
-        warn_method_deprecation(
+        exc.warn_method_deprecation(
             cls,
             "deserialize()",
             "parse_obj()",
@@ -236,7 +237,7 @@ class DeprecatedSerializableMixin(BaseModel):
             For more details, refer to the Pydantic documentation:
             https://docs.pydantic.dev/usage/models/#helper-functions
         """
-        warn_method_deprecation(
+        exc.warn_method_deprecation(
             self.__class__.__name__,
             "from_dict()",
             "parse_obj()",
@@ -261,7 +262,7 @@ class DeprecatedSerializableMixin(BaseModel):
             For more details, refer to the Pydantic documentation:
             https://docs.pydantic.dev/1.10/usage/exporting_models/
         """
-        warn_method_deprecation(
+        exc.warn_method_deprecation(
             self.__class__.__name__,
             "to_dict()",
             "dict()",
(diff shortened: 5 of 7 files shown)


lwasser:
nope - fixed locally but not here 🤷‍♀️ back to drawing board.

lwasser:
there are a few documentation issues that i'll clean up here it hink

- [ ] WARNING: `footer_items` is deprecated. Use `footer_start` or `footer_end` instead. - i'm guess that's just pydata sphinx will look into
- [ ] i'm getting a bunch of unexpected indentation errors - will look into it - its in client.py which i didn't think i touched here but will fix here!

stravalib/model.py
```diff
@@ -376,29 +568,60 @@ class Athlete(
...
+        raw_type : int
+            The raw integer representing the athlete type.
+
+        Returns
+        -------
+        str
+            The string representation of the athlete type.
+        """
+
+        # TODO: Pylance - Expression of type "str | None" cannot be assigned to
```
lwasser, line 585:
```suggestion
```

stravalib/model.py
```diff
@@ -164,43 +219,89 @@ def deserialize(cls, attribute_value_mapping: Dict):
...
+            The `to_dict()` method is deprecated in favor of `dict()` method.
+            For more details, refer to the Pydantic documentation:
+            https://docs.pydantic.dev/1.10/usage/exporting_models/
         """
         warn_method_deprecation(
             self.__class__.__name__,
             "to_dict()",
             "dict()",
-            "https://docs.pydantic.dev/usage/exporting_models/",
+            "https://docs.pydantic.dev/1.10/usage/exporting_models/",
```
lwasser, line 268: this is just a page that no longer exists in the latest version of the docs so i pinned it to 1.10 where it does exist

stravalib/model.py
```diff
@@ -417,32 +639,44 @@ def is_authenticated_athlete(self):
         return self.is_authenticated
 
 
+# TODO: better description
```
lwasser, line 642: I wasn't sure exactly why this was being overwritten so wouldn't mind input on what the description here should be. there are a few instances of this where attrs are overwritten via a class like this.

stravalib/model.py
```diff
@@ -494,8 +728,13 @@ def __repr__(self):
         )
 
 
+# TODO: this wasn't working as intended - test again
```
lwasser, line 731: Once we have addressed #390 (which may be user error) i can test this again. it was not returning correct data because i was querying a activity via a date and getting a totally different subset of items back.

lwasser:
hey @enadeau @jsamoocha for model.py is it better to do this

```
@lazy_property
    def kudos(self)-> stravalib.client.BatchedResultsIterator[model.ActivityKudos]:
        return self.bound_client.get_activity_kudos(self.id)
```
or import BatchedResultsIterator at the top of the module and just do this which is a bit easier to read:
It's inherited so otherwise mypy doesn't know where it is coming from

```
@lazy_property
    def kudos(self)-> BatchedResultsIterator[model.ActivityKudos]:
        return self.bound_client.get_activity_kudos(self.id)
```

many thanks.

enadeau:
I'd say it's better to import it at the top. If you run into ciruclar import (which I think you will) take a look [here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html?highlight=TYPE_CHECKING#import-cycles) it's a classic problem people run into when they start working with type annotation.

lwasser:
i am definitely running into the circular import already (model.py + strava_model.py). i was skipping those for now but i'll look at that example to understand how to address it!!

lwasser:
Ok friends - this is ready for review - i'm going to leave some explicit questions because i'm a bit confused as to how to fix a few of the leftover typing issues. And also i want to ensure the changes i made are OK. the strava_model // model overrides are most confusing as i see we are intentionally overwriting some things from strava_model but i'm not sure if we need to use those types of the types from model.py that are there.

changelog.md
```diff
@@ -6,6 +6,7 @@
 - Update client's stream method to warn when using unofficial parameters (@enadeau, #385)
 - Type annotation to client file (@enadeau, #384)
 - Fix docstring in SleepingRateLimitRule (@enadeau)
+- Fix: docstrings in model.py, indentation errors and other documentation errors with sphinx, findfonts warning supression by removing opengraph (temporarily)
```
lwasser, line 9: can update this to "and update typing"

docs/conf.py
```diff
@@ -113,7 +115,7 @@
     "show_toc_level": 1,
     # "navbar_align": "left",  # [left, content, right] For testing that the navbar items align properly
     "github_url": "https://github.com/stravalib/stravalib",
```
lwasser, line 117: this is just reflective of the new pydata_sphinx_theme config settings

pyproject.toml
```diff
@@ -94,6 +94,7 @@ files = [
     "stravalib/protocol.py",
     "stravalib/exc.py",
     "stravalib/client.py",
+    "stravalib/model.py",
```
lwasser, line 97: we are under 20 errors now!

stravalib/model.py
```diff
@@ -59,26 +59,49 @@
...
 )
 
+if TYPE_CHECKING:
+    from stravalib.client import BatchedResultsIterator
+
 LOGGER = logging.getLogger(__name__)
 
 
-def lazy_property(fn):
+# TODO: stravalib/model.py:71: error: Please use "Callable[[<parameters>], <return type>]"  [misc]
```
lwasser, line 74: i need help typing this Callable - i can't seem to get this one right - i understand that it's a function but i don't understand how to know what the parameters are to type - @enadeau any suggestions??

stravalib/model.py
```diff
@@ -59,26 +59,49 @@
...
+    -----
     Assumes that fn (like a regular getter property) has as single
     argument a reference to self, and uses one of the (bound) client
     methods to retrieve an entity (collection) by self.id.
+
     """
 
     @wraps(fn)
-    def wrapper(obj):
+    def wrapper(obj: Any) -> Any:
```
lwasser, line 104: `Any` works but i'm not sure if this is explicit enough here.

stravalib/model.py
```diff
@@ -149,58 +190,132 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
+        ----------
+        1.0.0
+            The `deserialize()` method is deprecated in favor of `parse_obj()` method.
+            For more details, refer to the Pydantic documentation:
+            https://docs.pydantic.dev/usage/models/#helper-functions
+
+
         """
-        warn_method_deprecation(
+        exc.warn_method_deprecation(
```
lwasser, line 225: exc was imported above and then some modules were imported in a separate line. i just cleaned it up so we're calling exc. for each imported method, etc

stravalib/model.py
```diff
@@ -149,58 +190,132 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
             self.__class__.__name__,
             "from_dict()",
             "parse_obj()",
             "https://docs.pydantic.dev/usage/models/#helper-functions",
         )
-        # Ugly hack is necessary because parse_obj does not behave in-place but returns a new object
+        # Ugly hack is necessary because parse_obj does not behave in-place but
+        # returns a new object
+        # TODO - calling self.__init__ directly is not recommended because it may lead to unexpected behavior,
```
lwasser, line 260: i am not sure how to fix this typing issue. any suggestions are welcome

stravalib/model.py
```diff
@@ -259,6 +409,21 @@ class Club(
     BackwardCompatibilityMixin,
     BoundClientEntity,
 ):
+    """
+    Represents a single club with detailed information about the club including
+    club name, id, location, activity types, etc.
+
+    See Also
```
lwasser, line 416: i started to do this for docstrings but am wondering if we need it as we'd use it a lot?

stravalib/model.py
```diff
@@ -331,8 +531,23 @@ class Athlete(
...
+    """Represents high level athlete information including
+    their name, email, clubs they belong to, bikes, shoes, etc.
+
+    Notes
+    ------
+    Also provides access to detailed athlete stats upon request.
+    """
+
+    # Field overrides from superclass for type extensions:
+    # TODO: these overrides make mypy upset - i think this might be related to the circular assignment??
```
lwasser, line 543: This is an error mypy raises for every attr override in these classes. should I just listen to mypy and use the attr types in strava_model?

stravalib/model.py
```diff
@@ -373,32 +588,66 @@ class Athlete(
...
+        """
+        Returns an `stravalib.Client` object containing Strava account data
+        for the (authenticated) athlete.
+
+        Returns
+        -------
+        DetailedAthlete
+            The detailed information of the authenticated athlete.
+        """
+        assert self.bound_client is not None, "Bound client is not set."
```
lwasser, line 623: mypy wants us to ensure bound_client is not None through out either assert of a conditional work. i went with asserts as it feels cleaner. i am not actually sure how the client could be None to begin with TBH. but the check is easy to add.

stravalib/model.py
```diff
@@ -533,9 +800,17 @@ class ActivityLap(
...
+    """Pass through object. Inherits from PolyLineMap"""
+
     pass
 
 
+# if TYPE_CHECKING:
+#     from strava_model import Split
+
+
+# TODO - Split is already defined in strava_model - Name "Split" already
```
lwasser, line 812: This particularly inheritance is really confusing to me. Split i the class name but it inherits Split as well. any help here is appreciated.

stravalib/model.py
```diff
@@ -818,19 +1111,37 @@ class Activity(
     )
 
     @lazy_property
-    def comments(self):
+    def comments(self) -> BatchedResultsIterator[ActivityComment]:
+        assert self.bound_client is not None
         return self.bound_client.get_activity_comments(self.id)
 
+    # TODO: question - this returns a list of model.BaseActivityZone objects
```
lwasser, line 1118: pls ignore this comment - i asked it above and can clean up this comment once it's answered above :) .  there are a few errors that pop up over and over in this module. once i better understand them i can fix throughout!

stravalib/model.py
```diff
@@ -926,10 +1241,35 @@ class SubscriptionCallback(
     hub_verify_token: Optional[str] = None
     hub_challenge: Optional[str] = None
 
+    # TODO: error: Required dynamic aliases disallowed  [pydantic-alias]
```
lwasser, line 1244: i need help with this one as well.

lwasser: i asked chatGPT and it wants an implementation like this using a static method rather than a lambda function.
```python
    @staticmethod
    def alias_generator(field_name: str) -> str:
        return field_name.replace("hub_", "hub.")
```

stravalib/model.py
```diff
@@ -926,10 +1241,35 @@ class SubscriptionCallback(
...
     class Config:
         alias_generator = lambda field_name: field_name.replace("hub_", "hub.")
```
lwasser, line 1247: this is another override type issue. i think once i understand overrides i can easily fix them all.

stravalib/model.py
```diff
@@ -926,10 +1241,35 @@ class SubscriptionCallback(
...
 
-    def validate(self, verify_token=Subscription.VERIFY_TOKEN_DEFAULT):
+    # TODO: not sure if this docstring and typing is right- mypy is telling me
+    # there is an inconsistency -  Signature of "validate" incompatible with supertype "BaseModel"  [override]
+    def validate(
```
lwasser, line 1250: stravalib/model.py:1250: error: Signature of "validate" incompatible with supertype "BaseModel"  [override]
stravalib/model.py:1250: note:      Superclass:
stravalib/model.py:1[25](https://github.com/stravalib/stravalib/actions/runs/5719816139/job/15498500654?pr=387#step:11:26)0: note:          @classmethod
stravalib/model.py:1250: note:          def validate(cls, value: Any) -> SubscriptionCallback
stravalib/model.py:1250: note:      Subclass:
stravalib/model.py:1250: note:          def validate(self, verify_token: str = ...) -> None

enadeau:
Well done was surely a lot of work :100: 

 I'll be happy to review and try to help with the fixes but I'll probably take couple of days.

Do you want me to push fixes to the branch with some fix or should I try to explain in comments?

stravalib/client.py
```diff
@@ -959,7 +960,7 @@ def delete_activity(self, activity_id: int) -> None:
 
     def get_activity_zones(
         self, activity_id: int
-    ) -> list[model.BaseActivityZone]:
+    ) -> List[model.BaseActivityZone]:
```
enadeau, line 963: Any reason to change that to upper case. We've been quite consistent with the lower case annotation (which is the newest way)

lwasser: @enadeau gosh i must understand this incorrectly. I've been told to use uppercase `List`, `Dict` etc from the typing module. But you're saying that lower is newest? i'm actually a bit confused about this because there is the `|` annotation vs `Union` and then `from future import` for older python versions. i'm really not sure what is best practice here. but if you are saying lowercase is best i can fix this as i implement it throughout!

lwasser:
> Well done was surely a lot of work 💯

thank you!

> I'll be happy to review and try to help with the fixes but I'll probably take couple of days.
no rush!!
> Do you want me to push fixes to the branch with some fix or should I try to explain in comments?

If you have patience to explain in comments that would be wonderful. but if it's too complex / time consuming for some things to explain just push / fix things and i'll ask if i don't follow! 

there are a few items that i'm really not sure how to type (i left some comments about this).

stravalib/client.py
lwasser: fixed! just discovered how good the mypy docs are. i understand now 3.9+ is now lowercase. 3.8 and below use typing module! i just pushed a fix.

lwasser: Ok thanks @enadeau i'll add the from `__future__ import annotations` and will replace union and optional. that is going to make the docstrings a bit messier! eeks! i also see a failure on 3.8 and am curious if it's because i'm missing that import...

lwasser: Yup. it's a typing issue. OK i'm learning a lot here - thank you!!

lwasser:
ok a few things - some of the tests are failing due to typing issues. i tried adding from future... to the conftest.py file but i'm not sure that is the correct solution

2.  i get this type of error a lot: 
```
stravalib/model.py:1159: error: Incompatible types in assignment (expression has type "Optional[List[DistributionBucket]]", base class "ActivityZone" defined the type as "Optional[TimedZoneDistribution]")  [assignment]
```

and it's always because of a field override like this:

```
    distribution_buckets: list[DistributionBucket] | None = None
```

which happens a lot. should i listen to mypy and use the super class type which is overridden... or is there some other way to handle this?

lwasser:
@enadeau well i went through and removed union and optional but then i found [this comment](https://github.com/tiangolo/typer/issues/371#issuecomment-1073649190) which might be why our tests are breaking

it looks like the `|` was added in python 3.10

stravalib/model.py
lwasser: i suspect this should be Callable[Any], Any] ... thank you mypy docs :) is that right?

jsamoocha: Sorry for not responding before - too busy at the moment. I think your solution will work, but maybe it can/should be more precise (which is not my expertise). @enadeau, Any thoughts on this? Will the correct typing be inferred from the returns of the decorated methods?

enadeau:
Oooppps 🙈 that's my mistake, sorry for that. Using the more modern version won't work in 3.8, 3.9 when used in pydantic model because pydantic uses the types at runtime not only to check the correctness of the code and do autocomplete.

We need to use the older annotation for models.

Again sorry for this confusion. At least I'm learning something :smile:

stravalib/model.py
enadeau: Yes a decorator is in fact a function that takes a function and return a function. Mypy docs has a good section on it https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

We can probably be quite precise with the types that go in. I'd need to look at the code a bit. I'll try to come back to this quickly

lwasser: ok thank you both! and @jsamoocha no worries we are all busy and i get that. i'm just enjoying learning a bunch here from you both. i pushed some cleanup just now. don't worry about the tests failing now - i think it's related to some typing issues / errors that i have here. we can just chip away at this until mypy is happy!!

stravalib/model.py
enadeau: This decorator is quite complex as it turns the function into a property. I've tried to come up with a compromise between  precision and not bloating the code too much.

The best solution would be to be able to say that lazy property returns `property[T | None` but property is not generic so we can't to that.

Instead I'm pretending the return type is simply a `T | None` and use an ignore comment. I think it should do the job.

With this fix I go down to 15 errors! Let me know if you have any questions.
```diff
diff --cc stravalib/model.py
index b23d0c2,6df6e62..0000000
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@@ -22,6 -22,6 +22,7 @@@ from typing import 
      ClassVar,
      Literal,
      Optional,
++    cast,
      Tuple,
      Union,
      get_args,
@@@ -69,8 -75,11 +76,10 @@@ if TYPE_CHECKING
  LOGGER = logging.getLogger(__name__)
  
  
- # TODO: how do you properly type this with more precision than Any?
- def lazy_property(fn: Callable[Any], Any) -> property:
+ T = TypeVar("T")
+ U = TypeVar("U", bound="BoundClientEntity")
+ 
 -# TODO: stravalib/model.py:71: error: Please use "Callable[[<parameters>], <return type>]"  [misc]
 -def lazy_property(fn: Callable[[U], T]) -> Callable[[U], T]:
++def lazy_property(fn: Callable[[U], T]) -> T | None:
      """
      Should be used to decorate the functions that return a lazily loaded
      entity (collection), e.g., the members of a club.
@@@ -99,13 -108,13 +108,13 @@@
      """
  
      @wraps(fn)
-     def wrapper(obj: Any) -> Any:
 -    def wrapper(obj: U) -> T:
++    def wrapper(obj: U) -> T | None:
          try:
              if obj.bound_client is None:
                  raise exc.UnboundEntity(
                      f"Unable to fetch objects for unbound {obj.__class__} entity."
                  )
--            if obj.id is None:
++            if obj.id is None: # type: ignore[attr-defined]
                  LOGGER.warning(
                      f"Cannot retrieve {obj.__class__}.{fn.__name__}, self.id is None"
                  )
@@@ -116,7 -125,7 +125,7 @@@
                  f"Unable to fetch objects for unbound {obj.__class__} entity: {e}"
              )
  
--    return property(wrapper)
++    return property(wrapper) # type: ignore[return-value]
  
  
  # Custom validators for some edge cases:

```

stravalib/model.py
```diff
@@ -59,26 +57,49 @@
...
 
+if TYPE_CHECKING:
+    from stravalib.client import BatchedResultsIterator
+
 LOGGER = logging.getLogger(__name__)
 
 
-def lazy_property(fn):
+# TODO: how do you properly type this with more precision than Any?
+def lazy_property(fn: Callable[Any], Any) -> property:
```
enadeau, line 73: This line is actually the cause of all the failing test I think. It should read. After applying the modification from my other comment, all test pass.
```suggestion
def lazy_property(fn: Callable[[Any], Any]) -> property:
```

stravalib/model.py
```diff
@@ -149,58 +188,132 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
+        ----------
+        1.0.0
+            The `from_dict()` method is deprecated in favor of `parse_obj()` method.
+            For more details, refer to the Pydantic documentation:
+            https://docs.pydantic.dev/usage/models/#helper-functions
         """
-        warn_method_deprecation(
+        # TODO: Argument 1 to "warn_method_deprecation" has incompatible type "str"; expected "type
+        exc.warn_method_deprecation(
             self.__class__.__name__,
```
enadeau, line 251: The function except the class itself as the first argument, not its name.
```suggestion
            self.__class__
```

stravalib/model.py
lwasser: oh ok! thanks! oh yes i see that's just wrong syntax.

stravalib/model.py
lwasser: ok this one is really complex. so first you define type var's U and T. And U has a bound object associated with it - the BoundClient object. 
```
T = TypeVar("T") 
U = TypeVar("U", bound="BoundClientEntity")
```

then for the lazy_property definition you define the input as a Callable that is passed U and returns T. 

```
def lazy_property(fn: Callable[[U], T]) -> T | None:
```

i have two questions

1. Question: why wouldn't  `fn` return `T | None` as well?
2. if we have the T or None return why do you need this:

`type: ignore[return-value]`

stravalib/model.py
lwasser: i dug into this more *thank you chatgpt) and we have a few options here.

1. just use the class defined in strava_model.py but that's not quite right. 
2. use an Union / | to specify both classes
3. ignore these overrides. 

i don't know which answer is the appropriate one for this module.

lwasser: we are down to 13 errors now!! yaaaayaaah! this issue if we decide on how to address it will get us down even further as i think there are like 6 of these errors atleast!

stravalib/model.py
enadeau: 1. We don't know anything about the function `fn` that would be passed into the decorator so we can't assume that it could return `None`. In the wrapper there is a possibility to return `None` so we made the return type a union of `T` and `None`
2. If you try to remove the comment it wil complain about returning a `property` not a `T | None`. We wan't the return type to be `T | None` and not property since it will be more useful later. See more detail here https://stackoverflow.com/questions/75281747/how-to-annotate-a-decorator-that-returns-a-property
in particular, the second answer

lwasser:
so right now many tests are failing. I really think this is related to the issue of overriding attributes in model.py from strava_model.py and associated typing. 

the first test that fails yields this message:

```
FAILED stravalib/tests/unit/test_model.py::ModelTest::test_entity_collections - NameError: name 'SummaryGear' is not defined

# Followed by some pydantic errors

FAILED stravalib/tests/integration/test_client.py::test_get_athlete - pydantic.errors.ConfigError: field "bikes" not yet prepared so type is still a Forward...
FAILED stravalib/tests/integration/test_client.py::test_get_activity_zones - pydantic.errors.ConfigError: field "distribution_buckets" not yet prepared so type is ...
```

I did make this change in the code - i imported SummaryGear within the type_checking conditional:
```
if TYPE_CHECKING:
    from stravalib.client import BatchedResultsIterator
    from stravalib.strava_model import (
        SummaryGear,
        SummarySegment,
        TimedZoneDistribution,
    )
```

and i added this `Union` element to allow mypy to know that in strava_model. bikes is a llist of SummaryGear elements.

```
# but it is defined this way in strava_model.py
class DetailedAthlete(SummaryAthlete):
    bikes: Optional[List[SummaryGear]] = None
```

 But because of how we are populating gear items in model.py, it's actually a list of Bike objects. so i tried telling mypy it could be either. Is there a better way to handle this? or should i just override typing for these lines (there are a handful of these). 

```
# This is how bikes: is defined in model.py
bikes: Optional[list[Union[Bike, SummaryGear]]] = None
```
i just wonder if when tests run, the if TYPE_CHECKING conditional isn't running?? and as such SummaryGear can't be found and i get a name error. i even tried an import in the test itself as a bit of a desparate measure 😆 .

```
    def test_entity_collections(self):
        a = model.Athlete()
        d = {
            "clubs": [
                {"resource_state": 2, "id": 7, "name": "Team Roaring Mouse"},
                {"resource_state": 2, "id": 1, "name": "Team Strava Cycling"},
                {
                    "resource_state": 2,
                    "id": 34444,
                    "name": "Team Strava Cyclocross",
                },
            ]
        }
        # me trying desperate things 🥇 
        from stravalib.strava_model import SummaryGear
        a.update_forward_refs()
        a.from_dict(d)
```

I am not sure what else to try here 🤷‍♀️

jsamoocha:
These `Bike` and `Shoe` classes are just for backward compatibility with the pre-pydantic version of stravalib. Regardless of the "correct" typing solution, I think it's best to remove them eventually so our API more closely matches Strava's. We can maybe deprecate them somehow and ignore related typing issues.

Unfortunately, I don't have the time right now to help in more depth with this work.

lwasser:
ok that contact is super helpful @jsamoocha . i totally get your swamped and don't want to burden you now with it!

SO would a reasonable path forward be:

1. we can ignore the type checking on this attributes for the time being
2. we can add a deprecation warning to those types?

i could easily add the mypy ignore for the time being and align typing with strava_model.py 
i could potentially (in another PR) try to add a warning deprecation to the attributes  / class instantiation?

we could implement a release (i can do that) once the warnings are in place so people know to move away from that behavior. 

thoughts?
@enadeau does that seem reasonable to you particularly on the typing end of things?

enadeau:
> i just wonder if when tests run, the if TYPE_CHECKING conditional isn't running?? and as such SummaryGear can't be found and i get a name error. i even tried an import in the test itself as a bit of a desparate measure laughing .

You're right, the TYPE_CHECKING constant is always False at runtime. So the statement in the if won't be excuted in the test or when the library is used.

If those things are going to be deprecated, I think it makes sense to not worry too much about the typing. We can use some place some well chosen ignore or maybe it's a good use case for the use of [cast](https://mypy.readthedocs.io/en/stable/type_narrowing.html?highlight=cast#casts).

lwasser:
> cast

OK! that sounds perfect. i'll play with both cast and ignoring those lines that we do plan to deprecate the old behavior.  this will actually fix a handful of existing mypy errors.

lwasser:
Ok locally im down to 5 mypy errors and 2 broken tests... so i'm slowly making progress but we'll see what CI says here. i skipped pre-commit on this one as it's a flake8 line length issue that i can fix once things are running again.

stravalib/model.py
```diff
@@ -926,10 +1242,35 @@ class SubscriptionCallback(
...
     hub_challenge: Optional[str] = None
 
+    # TODO: error: Required dynamic aliases disallowed  [pydantic-alias]
     class Config:
         alias_generator = lambda field_name: field_name.replace("hub_", "hub.")
 
-    def validate(self, verify_token=Subscription.VERIFY_TOKEN_DEFAULT):
+    # TODO: not sure if this docstring and typing is right- mypy is telling me
+    # there is an inconsistency -  Signature of "validate" incompatible with supertype "BaseModel"  [override]
+    def validate(
```
lwasser, line 1251: if either of you happen to be able to help me understand what's happening here? it seems like it's somehow overriding pydantic's `BaseModel` looking at the errors and it does inherit from `DeprecatedSerializableMixin`.

`DeprecatedSerializableMixin(BaseModel)`

it seems like this is another override supporting legacy behavior. 

the full error is below
**************

1. it seems to want it to be a `@classmethod` - would that be a correct decorator for this object? and it doesn't like the mismatch in param names. 
```
Method "validate" overrides class "BaseModel" in an incompatible manner
  Base method is declared as a classmethod but override is not
  Parameter 2 name mismatch: base parameter is named "value", override parameter is named "verify_token"
  Return type mismatch: base method returns type "Model@validate", override returns type "None"
    Type "None" cannot be assigned to type "BaseModel*"Pylance[reportIncompatibleMethodOverride](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIncompatibleMethodOverride)
main.py(683, 9): Overridden method
(method) def validate(
    self: Self@SubscriptionCallback,
    verify_token: str = Subscription.VERIFY_TOKEN_DEFAULT
) -> None
Validate the subscription's hub_verify_token against the provided token.

Parameters
verify_token : str
    The token to verify subscription If not provided, it uses the default VERIFY_TOKEN_DEFAULT.

Returns
None
    None if an assertion error is not raised

Raises
AssertionError
    If the hub_verify_token does not match the provided verify_token.
```
not sure if this is another spot to simply ignore the type check but seems like it might be if it's soon-to-be deprecated behavior??

enadeau: I find this one a bit problematic. We are here overwriting the `validate` method from `pydantic.BaseModel` this could cause some unexpected behavior. Ideally, this method should be renamed to something else. 
It's probably enough part of the internal of the library to allow us to change it?

I think the best solution here would be to put ignore here and open another issue to treat that problem

jsamoocha: > It's probably enough part of the internal of the library to allow us to change it?

This also came from the pre-pydantic version of stravalib, and I retained the name to be sure nothing was getting broken. Renaming this method looks like a relatively small risk of breaking stuff.

stravalib/model.py
```diff
@@ -533,9 +803,17 @@ class ActivityLap(
...
+
     pass
 
 
+# if TYPE_CHECKING:
+#     from strava_model import Split
+
+
+# TODO - Split is already defined in strava_model - Name "Split" already
+# defined (possibly by an import) - i'm not sure how to deal with this.
```
lwasser, line 816: Ok here is another thing that i'd like to IGNORE for now but then open an issue about a better approach. 

if you look at this class `Split()`
it inherits from Split in the strava_model.py module.

it seems to be appending a few undocumented attrs and add a helper conversation attr.

i wondered if there might be an alternative way of doing this with composition? or creating a small class here called - SplitUndocumented and allowing the Split() class to inherit from that?

Thoughts? again i can open an issue but i'm going to ignore for now unless there is something here i don't understand.

stravalib/model.py
```diff
@@ -856,11 +1170,11 @@ class BaseActivityZone(
     A collection of :class:`stravalib.model.DistributionBucket` objects.
     """
 
-    # field overrides from superclass for type extensions:
-    distribution_buckets: Optional[List[DistributionBucket]] = None
+    # Field overrides from superclass for type extensions:
```
lwasser, line 1173: Ok making a note here - i'm defaulting back to the original typing for these overrides - this is because they are what's breaking our tests right now. It expects the types associated with the override that support legacy behavior.

SO to fix this we will need to 

1. drop support for legacy behavior
2. remove the overrides and use the types associated with the new strava_model
3. this approach is going to fail for the instances where we are supporting undocumented behavior. i'm thinking maybe we want to come up with a slightly different approach to append new attrs to strava_model classes using some sort of composition or inheritance of a new class that adds those attrs .

lwasser:
tests are running again locally on python 3.9-3.11 (not 3.8 i dont have 3.8 locally)

i think i have 3 mypy issues left to resolve. one likely will be an ignore as it's another override. 
the other two i may need feedback on. but yay for more green test checks!

stravalib/model.py
jsamoocha: Without having tested it with mypy, could this naming clash not easily be resolved by defining the class as

```
class Split(strava_model.Split, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
    ...
```
and importing `strava_model`?

enadeau: Yeah that should work. I've run into this issue a couple of time with classes that have the same name in models ans strava_model

stravalib/model.py
```diff
@@ -459,8 +722,9 @@ class ActivityPhoto(BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     created_at: Optional[datetime] = None
     created_at_local: Optional[datetime] = None
     location: Optional[LatLon] = None
-    urls: Optional[Dict] = None
-    sizes: Optional[Dict] = None
+    urls: Optional[dict[str, str]] = None
+    # TODO - is this a float return?
+    sizes: Optional[dict[str, float]] = None
```
enadeau, line 727: These two annotations are my best guess for the 3.8 test failing. Tried to replace them by `Dict` from the typing module. The use of `dict` for annotation has only been introduced in 3.9

lwasser: oh gosh - i forgot about that. you are right - it's running now locally. thank you!!

lwasser:
all tests are now passing locally! 🎉 4 mypy issues left! - i'll see if i can chip away at these tonight and will leave a question if i'm still stumped! thank you both!! ✨ 

```
travalib/model.py:267: error: Accessing "__init__" on an instance is unsound, since instance.__init__ could be from an incompatible subclass  [misc]
stravalib/model.py:1247: error: Required dynamic aliases disallowed  [pydantic-alias]
stravalib/client.py:1841: error: "DeprecatedSerializableMixin" has no attribute "validate_token"  [attr-defined]
stravalib/client.py:1842: error: "DeprecatedSerializableMixin" has no attribute "hub_challenge"  [attr-defined]
```

enadeau:
Well done, it's getting really close. I think this file had more challenge than the other one I took.

I take a look at the remaining errors and here's my quick take

> travalib/model.py:267: error: Accessing "__init__" on an instance is unsound, since instance.__init__ could be from an incompatible subclass  [misc]

This is in a deprecated method from pre-pydantic time. I'd just ignore it

> stravalib/model.py:1247: error: Required dynamic aliases disallowed  [pydantic-alias]

I'd solve this one by not having the alias generator and defining the alias on the three attributes of this class manually

> stravalib/client.py:1841: error: "DeprecatedSerializableMixin" has no attribute "validate_token"  [attr-defined]
> stravalib/client.py:1842: error: "DeprecatedSerializableMixin" has no attribute "hub_challenge"  [attr-defined]

Those two are in the same method and are caused by the fact that the return type of the deserialize method is not precise enough. You wan't to say it returns an object of the same type as the cls given in the first argument. This can be achieved using [`Self`](https://mypy.readthedocs.io/en/stable/generics.html?highlight=Self#automatic-self-types-using-typing-self) or a type variable

Hope this helps :)

lwasser:
@enadeau WOW thank you. that would have taken me a really long time to figure those out!! look what just happened - FINALLY!! i'm going to do another check / cleanup of comments and such and then i think this is ready for a final real review to ensure what i implemented is generally correct, etc!!  phew!!

Screen Shot 2023-08-09 at 7 02 03 PM

lwasser:
one oddity is that i don't understand why we use `dict[]` in some modules with no errors but it was causing me errors with python 3.8 running tests in module.py. that is confusing. i also noticed that sometimes as i fixed things other errors popped up OR more things that i expected resolved.

lwasser:
OMG! all tests AND mypy and precommit are happy now!! 

**ready for a review y'all!  🎉**

Note - i did end up ignoring some of the overrides in the `Split()` class but adding `strava_model.Split` import did help with the inheritance confusion mypy had.

lwasser:
also feel free to push changes to this PR or edit inline on GitHub if you wish ✨ otherwise i'm happy to keep iterating until this looks right.

lwasser:
great- i just invited you to my fork and can give you access there,

but also - @jsamoocha and i wondered if you had any interest in joining our maintainer team? you've already done so much and helped so much!! if so i can give you access so you can push commits to pr's here if the author clicks that check that allows maintainers to push to pr's. 

let us know!!

lwasser:
oh and a request - if you don't mind - could you kindly point out things you're fixing / let me know where i went wrong? i was uncertain about some of my decisions but mypy was ok with them. so i'd love to learn a bit more here if you have time to point things out!

enadeau:
> 

> but also - @jsamoocha and i wondered if you had any interest in joining our maintainer team? you've already done so much and helped so much!! if so i can give you access so you can push commits to pr's here if the author clicks that check that allows maintainers to push to pr's.

I'll be happy to join the team! It's been really nice working with you so far :smile: 



> oh and a request - if you don't mind - could you kindly point out things you're fixing / let me know where i went wrong? i was uncertain about some of my decisions but mypy was ok with them. so i'd love to learn a bit more here if you have time to point things out!

I'll try to go for more comment than to let you know what needs fixing

stravalib/client.py
```diff
@@ -1819,7 +1823,7 @@ def handle_subscription_callback(
         self,
         raw: dict[str, Any],
         verify_token: str = model.Subscription.VERIFY_TOKEN_DEFAULT,
-    ) -> dict[str, str]:
+    ) -> dict[str, Optional[str]]:
```
enadeau, line 1826: The optional is not needed here better to only have str as in the docstring since it will always return a string. To silence the warning we can use an assert like this 
```py
assert callback.hub_challenge is not None
```

stravalib/model.py
```diff
@@ -149,58 +194,128 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
     Provides backward compatibility with legacy BaseEntity
+
+    Inherits from the `pydantic.BaseModel` class.
     """
 
+    # TODO: i used the output type from this class
+    # could dict also be an int or float??
     @classmethod
-    def deserialize(cls, attribute_value_mapping: Dict):
+    def deserialize(cls, attribute_value_mapping: dict[str, str]) -> Self:
```
enadeau, line 204: I think the best type for the dict would be `dict[str, Any]`

stravalib/model.py
enadeau: Looks good to me

stravalib/model.py
```diff
@@ -149,58 +194,128 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
+        exc.warn_method_deprecation(
             cls,
             "deserialize()",
             "parse_obj()",
             "https://docs.pydantic.dev/usage/models/#helper-functions",
         )
         return cls.parse_obj(attribute_value_mapping)
 
-    def from_dict(self, attribute_value_mapping: Dict):
+    def from_dict(self, attribute_value_mapping: dict[str, str]) -> None:
```
enadeau, line 235: Again I think here the type should be `dict[str, Any]`

stravalib/model.py
```diff
@@ -149,58 +194,128 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
             "https://docs.pydantic.dev/usage/models/#helper-functions",
         )
-        # Ugly hack is necessary because parse_obj does not behave in-place but returns a new object
-        self.__init__(**self.parse_obj(attribute_value_mapping).dict())
+        # Ugly hack is necessary because parse_obj does not behave in-place but
+        # returns a new object
+        self.__init__(**self.parse_obj(attribute_value_mapping).dict())  # type: ignore
 
-    def to_dict(self):
+    def to_dict(self) -> Dict[str, str]:
```
enadeau, line 263: Also here `dict[str, Any]`

stravalib/model.py
```diff
@@ -149,58 +194,128 @@ def naive_datetime(value: Optional[AllDateTypes]) -> Optional[datetime]:
...
+        Deprecated
+        ----------
+            The `to_dict()` method is deprecated in favor of `dict()` method.
+            For more details, refer to the Pydantic documentation:
+            https://docs.pydantic.dev/1.10/usage/exporting_models/
         """
-        warn_method_deprecation(
-            self.__class__.__name__,
+        exc.warn_method_deprecation(
+            self.__class__,  # expects klass: type (not a str)
```
enadeau, line 279: The comment can probably be removed or at least `klass -> class`

stravalib/model.py
```diff
@@ -240,16 +357,49 @@ class LatLon(LatLng, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     """
 
     @root_validator
-    def check_valid_latlng(cls, values):
-        # Strava sometimes returns an empty list in case of activities without GPS
+    def check_valid_latlng(
+        cls, values: List[Optional[float]]
+    ) -> Optional[List[Optional[float]]]:
```
enadeau, line 362: I feel like here the types shoud be `list[float]` for input and `list[float] | None` for output unless I'm misunderstanding and the list themselves can contain None value?

stravalib/model.py
```diff
@@ -373,32 +581,66 @@ class Athlete(
...
     admin: Optional[bool] = None
     owner: Optional[bool] = None
-    subscription_permissions: Optional[list] = None
+    # TODO: can permission take other inputs other than bool?
+    subscription_permissions: Optional[List[bool]] = None
 
     @validator("athlete_type")
-    def to_str_representation(cls, raw_type):
-        # Replaces legacy "ChoicesAttribute" class
+    def to_str_representation(cls, raw_type: int) -> Optional[str]:
```
enadeau, line 588: Here the return type is always string, as you wrote in the comment below. The optional is not needed

stravalib/model.py
enadeau: Yeah if there's no good reason for the value to be None I think it's better going with an assert

stravalib/model.py
```diff
@@ -412,37 +654,50 @@ def is_authenticated_athlete(self):
             else:
                 # We need to check this athlete's id matches the authenticated athlete's id
                 authenticated_athlete = self.authenticated_athlete
+                assert authenticated_athlete is not None
```
enadeau, line 657: Here are we sure the authenticated_athlete will always be not None? Maybe it's clear that it will always be not None or perhaps it should return False if it is None?

stravalib/model.py
```diff
@@ -459,8 +714,9 @@ class ActivityPhoto(BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     created_at: Optional[datetime] = None
     created_at_local: Optional[datetime] = None
     location: Optional[LatLon] = None
-    urls: Optional[Dict] = None
-    sizes: Optional[Dict] = None
+    urls: Optional[Dict[str, str]] = None
+    # TODO - is this a float return?
```
enadeau, line 718: There's still a TODO here I'm not sure what you meant

stravalib/model.py
```diff
@@ -533,10 +793,20 @@ class ActivityLap(
...
 
 class Map(PolylineMap):
+    """Pass through object. Inherits from PolyLineMap"""
+
     pass
 
 
-class Split(Split, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
+# TODO -opening an issue about perhaps a different approach to adding these attrs
+# to split()
```
enadeau, line 802: Also a TODO here I think this one has been addressed (meaning the issue is open now)

stravalib/model.py
```diff
@@ -649,13 +930,14 @@ class Segment(
...
 
-    # field overrides from superclass for type extensions:
+    # Field overrides from superclass for type extensions:
     start_latlng: Optional[LatLon] = None
     end_latlng: Optional[LatLon] = None
     map: Optional[Map] = None
     athlete_segment_stats: Optional[AthleteSegmentStats] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
-    activity_type: Optional[ActivityType] = None
+    # TODO Incompatible types in assignment (expression has type "Optional[ActivityType]", base class "DetailedSegment" defined the type as "Optional[Literal['Ride', 'Run']]")
```
enadeau, line 939: There's not error atm so I'll assume the TODO can go

stravalib/model.py
```diff
@@ -759,20 +1043,23 @@ class Activity(
...
     gear: Optional[Gear] = None
-    best_efforts: Optional[List[BestEffort]] = None
-    segment_efforts: Optional[List[SegmentEffort]] = None
-    splits_metric: Optional[List[Split]] = None
-    splits_standard: Optional[List[Split]] = None
+    best_efforts: Optional[List[DetailedSegmentEffort]] = None
+    segment_efforts: Optional[List[DetailedSegmentEffort]] = None
+    # Ignoring types here given there are overrides
+    splits_metric: Optional[List[Split]] = None  # type: ignore
+    splits_standard: Optional[List[Split]] = None  # type: ignore
```
enadeau, line 1050: I think it's good the be a bit more precise in the ignore comment and specify the actual error that we want to ignore (here `# type: ignore[assignment]`). Then if a warning for another rule is raised we will still get warned.

stravalib/model.py
```diff
@@ -759,20 +1043,23 @@ class Activity(
...
     photos: Optional[ActivityPhotoMeta] = None
-    laps: Optional[List[ActivityLap]] = None
+    # TODO Detailed activity has this defined as - Optional[List[Lap]]
+    laps: Optional[List[Union[ActivityLap, Lap]]] = None
```
enadeau, line 1053: Here since AcitivtyLap is a subclass of Lap maybe it's better to have simply Lap?

stravalib/model.py
```diff
@@ -759,20 +1043,23 @@ class Activity(
...
 
     # Added for backward compatibility
     # TODO maybe deprecate?
-    TYPES: ClassVar[Tuple] = get_args(
+    TYPES: ClassVar[Tuple[Any, ...]] = get_args(
```
enadeau, line 1057: It sounds like a good idea to deprecate those they also seems quite intertwined with some pydantic v1 magic that might cause problem migration to v2

stravalib/model.py
```diff
@@ -856,11 +1155,12 @@ class BaseActivityZone(
...
-    # field overrides from superclass for type extensions:
-    distribution_buckets: Optional[List[DistributionBucket]] = None
+    # Field overrides from superclass for type extensions:
+    # Using type that is currently mimicking legacy behavior...
+    distribution_buckets: Optional[List[DistributionBucket]] = None  # type: ignore
 
-    # overriding the superclass type: it should also support pace as value
-    type: Optional[Literal["heartrate", "power", "pace"]] = None
+    # TODO: ignoring type given legacy support will be deprecated
+    type: Optional[Literal["heartrate", "power", "pace"]] = None  # type: ignore
```
enadeau, line 1163: More precise ignore

stravalib/model.py
```diff
@@ -890,7 +1190,7 @@ class Route(
     # Superclass field overrides for using extended types
     athlete: Optional[Athlete] = None
     map: Optional[Map] = None
-    segments: Optional[List[Segment]]
+    segments: Optional[List[Segment]]  # type: ignore
```
enadeau, line 1193: More precise ignore

enadeau (commented):
Okay here's my review, I've focused mostly on the typing part and left a lot of comments. Most of them should be quite straightforward to fix and we can discuss the others.

The biggest issue, I still can't quite grasp is the whole thing related to type overrides. My understanding is some the type annotation as they are now reflecting the deprecated types and not new types. If that the case it would definitely be worth fixing it so the types reflect the more expected way of working and not the deprecated one. If you can point me at a good example to look at maybe I can try to investigate a bit more

stravalib/client.py
lwasser: that worked - thank you!

stravalib/model.py
lwasser: Fixed.

Are you suggesting Any because we don't know what will be in the attribute value mapping dict? I wasn't sure how to handle these instances where i wasn't sure what the range of possible values was but was nervous about just using `Any`

stravalib/model.py
lwasser: perfect!

stravalib/model.py
lwasser: fixed!

stravalib/model.py
lwasser: i had no idea where that comment came from but think it was something mypy spit out. deleted!

stravalib/model.py
lwasser: Ok so i *may* be reading this wrong but here is my understanding... there was a comment about the lat/lon values returned by Strava being empty sometimes (when say someone creates an activity but there is no gps attached). as such you could get an empty list here as an input to this method. 

i did not test this however (i'm still wrapping my head around our test suite) but i do know having activities with no x,y data is possible in strava.

stravalib/model.py
lwasser: Ok so here i was responding to this error:
```
stravalib/model.py:603: error: Incompatible return value type (got "Optional[str]", expected "str")  [return-value]
```

this error confuses me as i'm not sure how mypy is looking at this method and determine potential outputs. 

when running this IF for some reason raw_type is not 0 or 1 it will return None. i tried this

```
        athlete_type = {0: "cyclist", 1: "runner"}
        assert raw_type in athlete_type.keys()
        return athlete_type.get(raw_type)
```

 as a check but that doesn't seem to make mypy happy. any guidance is appreciated of if it's easier feel free to suggest a change directly / push here. i struggled with this one.

stravalib/model.py
lwasser: ok great - i just am not sure why `to_str_representation ` above isn't fixed with an assert. i'll resolve this one given it sounds like my approach is generally good here!!

stravalib/model.py
lwasser: ok this also is making my brain 🤯 a bit 😆 authenticated_athlete is a bool

i implemented this fix:

```python
if authenticated_athlete is None:
     return False
```
 
which i think makes more sense. i'm struggling to know whether there are edge cases where we could encounter  unexpected None values and where they would be. but i do think here it makes sense to simply return a False given it's an authentication check so if the code has gotten that far... we'd assume `false` is a safe response..

stravalib/model.py
lwasser: ah yes - for sizes i assumed that this relates to photo sizes and would thus be of type float. BUT could it be an int or some other numeric value? in short, I wasn't sure how to be certain that float  was the best type here. 

because this is an undocumented entity i feel like i'm guessing a bit which doesn't feel great.

stravalib/model.py
lwasser: my mistake - i missed this one. it's gone now!

stravalib/model.py
lwasser: cool - thanks. i wasn't paying attention closely to the fact that ignores had a specific type of error assigned.

stravalib/model.py
lwasser: of course. that worked - thank you!

stravalib/model.py
lwasser: cool - i just opened an issue here:

https://github.com/stravalib/stravalib/issues/400

i know that the migration to v2 will potentially be significant so let's implement this fix in a separate (smaller!) pr

stravalib/model.py
lwasser: cool - i'll do a find and replace for all ignores with out a specific error specified!

lwasser:
Ok

✅ typing ignores are all now more specific
✅ i removed spurious TODOs
✅ fixed all of the type situations that were not quite right
✅ tests are all passing but mypy has one complaint (i'm not sure how to handle it)

see:

https://github.com/stravalib/stravalib/pull/387/files#r1290634486 - my attempted fix broke everything and mypy was still angry. but if that raw_key is anything other than 0 or 1 it could return None. 

many thanks for all the time here in reviewing. i know you've spent a lot of time (AND i've learned a TON!!) so feel free to push a final fix IF you wish!! or i'll happily try something else with guidance. this is super close.

stravalib/model.py
enadeau: Exactly, we have no idea what will come in the dict. Could even be more nested dict. So Any is the more appropriate type

stravalib/model.py
enadeau: Okay, on closer inspection I think there might actually be a small bug in the code/typing. The validator is taking any value and transforming it into a `Optional[str]` but the `athlete_type` attribute on the class is of type `Optional[int]` which does not quite make sense as the value that ends up in the attribute is a string.

I think we need to alter the validator to a pre-validator that converts the int into a string before pydantic runs it's own validation. This way, the athlete type can be the correct type.

My proposed change is below and passes both type checking and unit test. @jsamoocha I'd like to get your opinion on this since this is undocumented behaviour I'm not 100% sure I understand.

```diff
diff --git a/stravalib/model.py b/stravalib/model.py
index cc873f4..335313d 100644
--- a/stravalib/model.py
+++ b/stravalib/model.py
@@ -544,7 +549,7 @@ class Athlete(
 
     # Undocumented attributes:
     # Undocumented attributes:
     is_authenticated: Optional[bool] = None
-    athlete_type: Optional[int] = None
+    athlete_type: Optional[Literal["cyclist", "runner"]] = None
     friend: Optional[str] = None
     follower: Optional[str] = None
     approve_followers: Optional[bool] = None
@@ -581,8 +586,10 @@ class Athlete(
     owner: Optional[bool] = None
     subscription_permissions: Optional[List[bool]] = None
 
-    @validator("athlete_type")
-    def to_str_representation(cls, raw_type: int) -> str:
+    @validator("athlete_type", pre=True)
+    def to_str_representation(
+        cls, raw_type: Any
+    ) -> Optional[Literal["cyclist", "runner"]]:
         """Replaces legacy 'ChoicesAttribute' class.
 
         Parameters
@@ -595,9 +602,16 @@ class Athlete(
         str
             The string representation of the athlete type.
         """
-
-        athlete_type = {0: "cyclist", 1: "runner"}
-        return athlete_type.get(raw_type)
+        if raw_type == 0:
+            return "cyclist"
+        elif raw_type == 1:
+            return "runner"
+        elif raw_type is None:
+            return None
+        else:
+            raise ValueError(
+                "value is not a valid representation ofr athlete type"
+            )
 
     @lazy_property
     def authenticated_athlete(self) -> Athlete:

**Comment by lwasser** [comment]
>I'll be happy to join the team! It's been really nice working with you so far 
> 😄

yay @enadeau !! it's been great working with you as well!! i invited you to the organization and can give you privileges for this repo as well. So psyched that we can all work together on this!!

File: stravalib/model.py (hunk 28 above)
Comment (lwasser) [reply stravalib/model.py]: ok got it. thanks! i'll resolve this now!

File: stravalib/model.py (hunk 33 above)
Comment (lwasser) [reply stravalib/model.py]: thanks!! If we moved forward with this solution,  the conditional could be a bit simpler if we did this:

```python
# Define dict - this would allow us to add additional types easily if they arise in the future
athlete_type = {0: "cyclist", 1: "runner"}
if raw_type in [0, 1]:
  return athlete_type.get(raw_type)
elif raw_type is None:
   return None
else:
   raise ValueError("raw_value is not a valid representation of athlete type")
```

But also i have a question about how this all works. When we run mypy is pydantic also running type checks?

lwasser: it looks like we only need to return none - UNLESS there is some instance where a value error would get raised that i'm missing?

stravalib/model.py
jsamoocha: The input is either a list of exactly two elements or an empty list, so `list[float]` should be its type. AFAIK we cannot specify that lists have exactly _n_ elements (as we can for tuples, right?).

The same goes for the output, but it can be `None` if the input is an empty list. So its type would become `list[float] | None`

stravalib/model.py
jsamoocha: I agree the correct (user-facing) type of that field should be `Optional[Literal["cyclist", "runner"]]`.

As this is an undocumented attribute and Strava may add values other than 0 or 1 in the future, I'd choose to be rather forgiving and have the validator implemented by just adding `pre=True` to the existing validator. It'll return our optional literal, or None if encountering another value. I don't see a need to raise an exception in this case.

enadeau: Agree you'll still need the if else construct however to get mypy happy

stravalib/model.py
lwasser: ok so if i understand correctly @jsamoocha the input should be either 

`List[Optional[float]]` which would be the same as `list[float | None]` since it can be a empty list??

then the output would be 

Optional[list[float]]  which is `list[float] | None`

i'm going to make that change now but can change if i am not understanding correctly.

stravalib/model.py
lwasser: eeks so a few questions - what i have now:

```python
 @validator("athlete_type", pre=True)
    def to_str_representation(
        cls, raw_type: int
    ) -> Optional[Literal["cyclist", "runner"]]:
        """Replaces legacy 'ChoicesAttribute' class.

        Parameters
        ----------
        raw_type : int
            The raw integer representing the athlete type.

        Returns
        -------
        str
            The string representation of the athlete type.
        """

        athlete_type = {0: "cyclist", 1: "runner"}
        if raw_type in [0, 1]:
            return athlete_type.get(raw_type)
        else:
            return None
```

1. should the type for raw_type be `raw_type: Literal[0, 1]` given those are the current supported values?
2. mypy is not happy with any of the options (mine or the one @enadeau presented with the else raising a ValueError. 

you can [see the errors here:](https://github.com/stravalib/stravalib/actions/runs/5837735990/job/15833748620?pr=387#step:11:13)

or here is what it's telling me

> stravalib/model.py:606: error: Incompatible return value type (got "Optional[str]", expected "Optional[Literal['cyclist', 'runner']]")  [return-value]

@enadeau i tried adding that ValueError and i get this

> stravalib/model.py:606: error: Incompatible return value type (got "Optional[str]", expected "Optional[Literal['cyclist', 'runner']]")  [return-value]
> stravalib/model.py:608: error: Statement is unreachable  [unreachable]
> stravalib/model.py:610: error: Name "raiseValueError" is not defined  [name-defined]

i generally understand what is happening here but i don't understand what mypy needs to pass

stravalib/model.py
```diff
@@ -240,16 +355,49 @@ class LatLon(LatLng, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     """
 
     @root_validator
-    def check_valid_latlng(cls, values):
-        # Strava sometimes returns an empty list in case of activities without GPS
+    def check_valid_latlng(
```
lwasser, line 361: based on our conversation, i made this change to the types:

```
    def check_valid_latlng(
        cls, values: List[Optional[float]]
    ) -> Optional[List[float]]:
```
but am still getting this error

```
stravalib/model.py:380: error: Incompatible return value type (got "Optional[List[Optional[float]]]", expected "Optional[List[float]]")  [return-value]
```

so i guess there must be some way for n input value  of None here??

stravalib/model.py
jsamoocha: I think `list[float]` is fine as input type since it can be empty. The type `list[float | None]` indicates that the list elements can be `None`, which is not the case here.

stravalib/model.py
jsamoocha: Based on @enadeau 's comment above, I think this should be the validator's logic:

```python
if raw_type == 0:
    return "cyclist"
elif raw_type == 1:
    return "runner"
else:
    return None
```

I see now that the dictionary `.get()` shortcut will not work because the dict itself is not typed, so mypy cannot check if one of the literal values will come out of it.

Let's also keep the `raw_type`'s type as `Any` as this is an undocumented attribute and we don't know what Strava will return in the future. Maybe we could log a warning in the `else` case to indicate we got an unexpected value.

stravalib/model.py
lwasser: ok thank you for the clarification. fixed! i think in my mind a None was the same as empty. but I understand now that they are two different things. (`None` vs empty list).  mypy is ok with this solution. resolving this

stravalib/model.py
lwasser: ok @enadeau i'm sorry, i didn't understand why we needed the more explicit construct but @jsamoocha now helps me understand that with `.get()` it's hard to type that dictionary. so by being explicit  mypy will "See" that there are two (Literals) options here. for some reason i thought we had to raise that error for mypy. but i get it now. i've also added a LOGGER.warning element as suggested.

lwasser:
Ok!

✅ ive fixed the last two issues. my apologies - i didn't fully understand why @enadeau solution was so explicit and now i do
✅ also didn't understand that `None` was different from empty (now did i know how to type that!)
✅ i spent some time playing with pydantic to better understand how it is ingesting  and validating our JSON api response data. now i get how validation works and what `BaseModel` does. 

i think the TODO's for a future effort might be to carefully deprecate some of the overrides supporting legacy behavior. Right now these items are mostly ignoring in our type checking. 

feel free to give this another review. everything is passing now and my understanding of pydantic, typing and how some of the inheritance works here is better now thanks to you both! 🎉 thank you for the patience!

enadeau (approved):
All of this is very nice, it's been a very big effort to get this file correct. I think it is ready to merge, we can always improve the typing in future PR 🚀

jsamoocha:
Awesome! Thanks to both of you for this Herculean effort 💪

lwasser:
@jsamoocha i feel pretty odd merging my own pr's so i'll wait for you to approve and merge :)  what a learning experience this was!! 

now that our team is growing a bit do we want some sort of agreement from x maintainer (maybe 2?) before merging or just one? what makes sense here y'all?? i'm fine with whatever you all want to implement. i just am never quite sure when to merge. if it's a tiny fix or something that just needs to get done we can merge. but this was a bigger pr with lots of issues addressed.

enadeau:
Having two of use seems reasonable to me. One would also do it in my opinion if we agree to consult on bigger change.

lwasser:
@enadeau that sounds good to me!

PR #398, opened 2023-08-09
Fix: rename validate to validate_token

### LINKED ISSUE #394: Rename validate - BaseModel override in stravalib model.py
In the stravalib model module, there is a method called validate that seems to override some of the expected functionality of `Basemodel.validate`. You can see the [discussion here](https://github.com/stravalib/stravalib/pull/387#discussion_r1286798070). It might make the most sense to rename this method to avoid the conflict with BaseModel. For the time being i've told mypy to ignore checking this method until the fix is in place. 

```python
class SubscriptionCallback(
    BackwardCompatibilityMixin, DeprecatedSerializableMixin
):
    """
    Represents a Webhook Event Subscription Callback.
    """

    hub_mode: Optional[str] = None
    hub_verify_token: Optional[str] = None
    hub_challenge: Optional[str] = None

    # TODO: error: Required dynamic aliases disallowed  [pydantic-alias]
    class Config:
        alias_generator = lambda field_name: field_name.replace("hub_", "hub.")

    # TODO: Signature of "validate" incompatible with supertype "BaseModel"  
    # [override] - we should consider renaming this method??
    @no_type_check
    def validate(
        self, verify_token: str = Subscription.VERIFY_TOKEN_DEFAULT
    ) -> None:
```

### LINKED ISSUE #396: Rename validate() to something different so it doesn't conflict with pydantic.BaseModel.validate
Currently we have a method called [`validate()`](https://github.com/stravalib/stravalib/blob/master/stravalib/model.py#L932) that is a part of the `SubscriptionCallback` class that inherits from `pydantic.BaseModel`. The pydantic model also has a method called `validate()` with different parameters and of a different method "type". Upon typing model.py, conflicts arise because the two methods have somewhat different behavior - [a discussion about this is here](https://github.com/stravalib/stravalib/pull/387#discussion_r1286537648).

We decided that it would be best to rename that method so it doesn't conflict with pydantic's method. 
Perhaps we could consider calling it something like `validate_token()` to be clear that it's specific to stravalib.

this should be a small, non breaking change (_i think?_) given it's only called through that Callback class once in Client.py.

lwasser:
closes #394

## Description

This addresses the issue of validate having the same name as pydantic.BaseModel and as such raising errors when typing stravalib... but also leaving us open to unexpected behavior. 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

tests ARE running locally. i know that it makes sense to add some tests here but i'm still getting to know our test suite and am not quite confident enough YET to add tests. but this should fix the typing warning we are getting in #387 as a starting place.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,7 +6,7 @@
 - Update client's stream method to warn when using unofficial parameters (@enadeau, #385)
 - Type annotation to client file (@enadeau, #384)
 - Fix docstring in SleepingRateLimitRule (@enadeau)
-- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #396)
+- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #394)
 
 ## v1.3.3
 
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #396)
+
 ## v1.3.3
 
 ### Fixed
--- a/stravalib/client.py
+++ b/stravalib/client.py
@@ -1720,7 +1720,7 @@ class Client(object):
 
         """
         callback = model.SubscriptionCallback.deserialize(raw)
-        callback.validate(verify_token)
+        callback.validate_token(verify_token)
         response_raw = {"hub.challenge": callback.hub_challenge}
         return response_raw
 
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -6,6 +6,7 @@
 - Update client's stream method to warn when using unofficial parameters (@enadeau, #385)
 - Type annotation to client file (@enadeau, #384)
 - Fix docstring in SleepingRateLimitRule (@enadeau)
+- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #396)
```
lwasser, line 9:
```suggestion
- fix: rename `SubscriptionCallback.validate` -> `SubscriptionCallback.validate_token` to avoid conflict with `pydantic.BaseModel` (@lwasser, #394)
```

jsamoocha:
That looks good, thanks!

PR #415, opened 2023-08-21
Add: typing field_conversions & unithelper modules

enadeau:
More works on #373

## Description

It done!! I've added type annotation to the last file from the library!

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update
- [x] Other (please describe)

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [x] This change doesn't require tests

## Did you include your contribution to the change log?

- [x] Yes, `changelog.md` is up-to-date.

Issue #254, opened 2022-11-12
Python version support


lwasser:
Hey y'all - i think we've informally decided to drop python 2.x which is a breaking change but i think really necessary. Also when building i think we should drop 3.6 which i @jsamoocha did. i just want to ensure that we are all on the same page as i'm modernizing the packaging infrastructure i also am dropping support for Python =< 3.6

[Here is the pr opened in my repo so i could play with CI](https://github.com/lwasser/stravalib/pull/7/files) - i'm not fully sure about poetry yet but the other changes i think are necessary to update our package's structure. are we all on the same page:

Drop support for. 
* python 2.x
* drop 3.6 or lower
And then related to that the current version is

version = "0.10.4"

Given we have some breaking changes would we bump to 0.11 here? we aren't quite at 1.0 given all of hte refactoring but this isn't exactly a minor update we are all doing. 

Many thanks for input @jsamoocha @hozn @yihong0618 !! Please respond that you are in support of (or not) dropping 2.x and 3.6 and lower. and then let me know what version we should bump to for this initial release.

jsamoocha:
I'm not sure about what problem is solved by poetry in this case. Will it force developers to also use poetry if they want to work with stravalib? With a plain `requirements.txt` file, everybody can at least use their own preferred virtual env + dependency management system.

Besides that, python >=3.7 and version 0.11 look great to me.

Just out of curiosity (I'm not an expert on versioning/releases), is there any (automated) relationship between: 

- the version mentioned in pyproject.toml,
- a tag/release pushed to this repo,
- the actual shipping of a release to pypi?

lwasser:
@jsamoocha so happy you responded here as i'm really super torn on what tools we should use. 

So it sounds like we all support python >3.7 as being our python base. perfect.

# TLDR 

* Poetry is ideal for dependency management. Just because we use it to manage dependency and build our wheel doesn't mean that all maintainers / contributors have to use it. We could keep a requirements file too - 
* Poetry might make it easy for l ocal builds of docs and such `poetry build docs` vs creating a make or nox file (i prefer nox).
* python semantic release is for versioning - and requires semantic commit messages. We could enforce those when merging pr's but it could be a bit more effort to think about commit names and be consistent. To automate we'd setup a workflow_dipatch item in ci and would click a button to bump and push to pypi. What is nice here is the tool manages what version so we aren't typing in a version when creating a tag. 
[...]

Issue #301, opened 2022-12-15
Create guidelines for what version of python we support


lwasser:
I've been chatting with colleagues and a standard for python spuport is the[ NEP-29 protocol](https://numpy.org/neps/nep-0029-deprecation_policy.html)

I suggest that we follow that policy in terms of dropping older versions of python to adhere to the broader community's approach. This actually is perfectly inilne with the drop that we just made for python 3.7. 

What do you all think about implementing this as a development / release policy?
[Python end of life map here:](https://endoflife.date/python).

essentially we'd then support python versions released in the past 3.5 years (42 months) to align with a broader community of packages.

enadeau:
I think it's a good idea to adhere to this broader standard. The fact that pint also follows this is also pushing me in that direction. Seems like a good compromise between being able to use modern feature and support many versions of python

lwasser:
i agree. i am happy to add this to the dev guide documentation. i think i opened this when i was initially thinking about dev and writing our dev guide. 

we can be a bit less aggressive too if we wish about dropping versions but 3.8 is fairly dated at this point. so i'd feel ok with us dropping it now especially given the issues you're running into !

enadeau:
Nice let's move forward with this I can take care of the PR that removes support for 3.8

lwasser:
i created a quick query to look at downloads. [the tool i used ](https://github.com/ofek/pypinfo/issues)didn't specify the timeframe for a stock query but i'm guessing a month based on our typical weekly downloads. most of our users are on 3.9 so i feel good about this drop. perhaps it will encourage folks to upgrade as well as 3.11 is the current recommended version!

i'll be able to look more closely at your pr, @enadeau soon! i just wanted to share this as i was curious. 

| python_version | download_count |
| -------------- | -------------- |
| 3.9            |         15,152 |
| 3.8            |          2,924 |
| 3.10           |          1,676 |
| 3.11           |          1,604 |
| 3.7            |          1,131 |
| 2.7            |             40 |
| 3.6            |             10 |
| 3.12           |              3 |
| Total          |         22,540 |

lwasser:
i think that we can close this issue now. If I am missing something please let me know! i think we should just follow the general lifecycle of supported versions of python!

PR #416, opened 2023-08-23
Drop python 3.8 support

enadeau:
Remove supports for python 3.8 following discussion on #301

## Description

- Remove python 3.8 support and ci run against that versions
- run pyupgrade for python 3.9+

## Type of change

- [x] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ maintainers = [
      {name = "Hans Lellelid"},
      {name = "Jonatan Samoocha"},
      {name = "Yihong"},
+     {name = "Émile Nadeau"},
     ]
 requires-python = ">=3.9"
 classifiers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ maintainers = [
      {name = "Jonatan Samoocha"},
      {name = "Yihong"},
     ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 classifiers = [
   "Development Status :: 4 - Beta",
   "Programming Language :: Python",
@@ -39,7 +39,6 @@ classifiers = [
   "Topic :: Scientific/Engineering",
   "Topic :: Software Development :: Libraries",
   "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
@@ -75,7 +74,7 @@ filterwarnings = [
 ]
 
 [tool.mypy]
-python_version = "3.8"
+python_version = "3.9"
 follow_imports = "silent"
 warn_redundant_casts = true
 warn_unused_ignores = true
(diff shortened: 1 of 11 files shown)


lwasser:
@enadeau this is cool - how does pyupgrade work? i just looked at the readme. do you tell it `pyupgrade python-version` or something to that effect and it just fixes everything?

lwasser:
ok i did just a bit more digging. 

I think we should release a minor version based on the existing new things BEFORE this merge. Then we can create a new version with this drop of 3.8. this will allow anyone running python 3.8 a bit of time to pin their workflow to stravalib 1.3.4 which still supports 3.8. Then maybe this is a 1.4 bump - i guess it's breaking in theory for anyone running Python 3.8. but i'm not sure if we want to jump to 2.0 - i was thinking 2.0 might be the pydantic 2.0 upgrade?

i'm not sure if @jsamoocha has time to weight in on this - but just want to think this through prior to merging something that will break some peoples applications if they are running 3.8.

lwasser:
One other thought @enadeau  - i'm talking with folks and thinking about our process here. If we make the changes in this PR it will REALLY break people's code running python 3.8. I wonder if we always want to wait on big code updates like this one UNTIL 3.8 is EOL but we could still drop support for 3.8 in our builds and such. 

OR is the issue with pint such that it requires code that is >=3.9? 
thank you. i'm just trying to carefully think this through in case there is a way that we can accomodate some 3.8 users for the time being in a more undocumented way (we say we support 3.9 + but we accomodate 3.8 IF that is possible. thoughts?

I'm happy to make a release regardless of everything except this pr in the upcoming days!

jsamoocha:
Having a minor release with the typing changes, then going to 1.4 on the drop of Python 3.8, and then going to 2.0 whenever we drop support for the pre-pydantic API seems a totally reasonable approach, @lwasser.

lwasser:
great! i'm fine with a minor bump for what's been merged to date. My only concern is allowing people to pin to the most recent version with bugs fixed if they need to. 

so to proceed:

1. i'll create a 1.4 release with what's been merged on main. 
2. Then we can drop support for 3.8 here, rebase to fix #415  and create another minor release - fully typed with those two additional pr's. 1.5 
3. After that i'll work a bit more on the nox migration (moving stravalib to a src/ layout and adding a nox build session) and updating docs with NEP-29 / python version support details (and then we can close #254 as well!  🎉 

i feel good about it given the consensus we have here at this point. let me know if i'm missing a step. i'll create a new issue to track all of the steps here as well :)

pyproject.toml
```diff
@@ -26,7 +26,7 @@ maintainers = [
      {name = "Jonatan Samoocha"},
      {name = "Yihong"},
     ]
```
lwasser, line 28:
```suggestion
     {name = "Yihong"},
     {name = "Émile Nadeau"},
    ]
```

@enadeau adding you here as a maintainer.

lwasser (approved):
This PR looks good to me! i made a small change adding you to the maintainer list in the pyproject.toml file, @enadeau but otherwise it's good and i'll merge.

Would you like to create the next release to make sure the steps are clear / you have correct access? or would like prefer that I do this one. it's up to you!!

lwasser:
Everything is running here. i'm going to merge this y'all. then @enadeau you can rebase and your other PR should build properly.

PR #423, opened 2023-08-28
Publish type annotations

### LINKED ISSUE #373: Working towards better type annotation

enadeau:
This is the last step to close #373 !!!

## Description

Just adding the marking file to mark the package as PEP 561 compliant

## Type of change

- [x] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 - Add: Add nox to run tests & build docs (@lwasser, #395)
 - Type annotation for all file in the library (@enadeau, #384, #415)
 - Remove python 3.8 support following NEP-29 (@enadeau, #416)
+- Stravalib now includes types annotation, the package is PEP 561 compatible (@enadeau, #423)
 
 ## v1.4
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,8 @@ local_scheme = "no-local-version"
 write_to =  "stravalib/_version_generated.py"
 write_to_template = '__version__ = "v{version}"'
 
+[tool.setuptools.package-data]
+"stravalib" = ["py.typed"]
 
 [project]
 name = "stravalib"


enadeau:
I'm really excited to get that done!

Once merge I'll need to check with test pypi to be sure the file is included correctly and everything works.

@lwasser  when you make the change to the src layout you'll have to be careful not break that. There's seems to be an additional config needed in with that layout https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data

lwasser:
> @lwasser when you make the change to the src layout you'll have to be careful not break that. There's seems to be an additional config needed in with that layout https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data

yes!! lucky for us i'm actually working on a [python package guide](https://www.pyopensci.org/python-package-guide/) and have spent some time talking with PyPA folks about these tools and others. I setup the original setuptools / setuptools scm config in the fall when i was initially doing that research. so i do know how to make the changes! i just didn't want to make too many changes at once in the fall because it was a fairly big overhaul of all of the infrastructure for the package!

but i'll be sure to unpack the wheel and such before pushing here as a PR to ensure things install, unpack as they should. it's actually working well locally but i need to do a few more tests before opening a pr. 🎉  i'll have a closer look at this pr tomorrow and then we can merge it as well. yahoo!!!

lwasser:
oh also -- we can actually test that the file lives in the wheel and sdist now. the wheel is just a zip file so you can just modify the extension and unzip it to see if the files are added correctly. i can also look tomorrow.but we don't need to wait for a push to test pypi to test that the typing file is being added to our package archives correctly.

lwasser (approved):
ok so this is what the unpacked wheel looks like - and the mypy file is there. i don't fully understand how that file works with mypy / what the benefits are to the user but do see the spec in the PEP. do you happen to know how the file behaves? it's empty but somehow it tells the user's gui to provide type hints?? 

Screen Shot 2023-08-29 at 2 34 37 PM

Other than that i can install from the wheel and import stravalib so all seems to be working. 
if you can help me understand what that file does i'd appreciate it! if not I can ask pypa folks!

lwasser:
ok so i'll go ahead and merge this. i left a question for you in the recently-opened bot pr. but this PR is solid and we're just about (or ready??) for a 1.5 release with the python 3.8 drop i think???

enadeau:
The file is mandatory to have type checker analyze our library when used as a third party. If we don't include it user will get this mypy warning when importing.
```
foo.py:1: error: Skipping analyzing "stravalib.client": module is installed, but missing library stubs or py.typed marker  [import]
```

You can read more about it https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker

That being said I don't know more about the rationale behind this desingn

PR #397, opened 2023-08-09
Add: nox config for docs, tests

lwasser:
closes #395

## Description

This change adds Nox to our build workflow. Right now we are using Make for everything. Nox is much cleaner because it's python and as such easy to read. it also supports virtual environments and running things across various versions for local debugging. 

This pr does NOT include building our package but that step will be easy to add. I just wanted folks to test it out to see if everyone is comfortable with moving to nox and removing the old-school Makefile that is much harder to read / maintain (IMO) :) with less fun features like environments and such for robust test dev locally.  ✨ 

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update
- [X] Other (please describe) 

This is an infrastructure update!

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [x] This change doesn't require tests

This helps us run our tests!

## Did you include your contribution to the change log?

- [ ] Yes, `changelog.md` is up-to-date.

*If you are having a hard time getting the tests to run correctly feel free to
ping one of the maintainers here!*

There is no change log yet, if folks like this approach i'll 

- [ ] add a build step
- [ ] remove make from our workflow
- [ ] update our documentation 

i was inspired to do this given the current issues with python 3.8. it's just easier to be able to run test locally on the version that i want without switching environments all of the time. 

test this out now!!

try
``` 
pip install nox
nox -s docs-live
```

docs-live session will start an interactive session where as you edit documentation files, the docs themselves will update in your browser. NOTE: i've had issues with api docs not updating but everything else should update!

```
nox -s test
```
if you use venv for environments this is for you. run tests on python 3.8 thru 3.11 locally!!
NOTE: you will need to install the various versions of python locally for this to work on your machine!

or if you are a mamba fan like me:

```
nox -s test_mamba
```

this will use conda/mamba environments!

PR #425, opened 2023-08-29
Fix: move stravalib to src/ layout and add nox build steps

lwasser:
closes #397
This is the second part of upgrading to nox - adding build and cleanup steps and moving to a src/ layout for easier testing. 

## Type of change

- [x] Other (please describe) - infrastructure update

## Does your PR include tests

- [ ] Yes


lwasser:
This only looks like a massive PR because i moved all of our code into a new directory. but i didn't actually modify any of the files - only the file structure. i did modify the noxfile, our init.py and other config files to work in the new structure.

lwasser:
Ok - things are working generally now but i have to do a bit more testing. 
Once i'm happy with things i'll update docs on the build SDist / wheel / package distribution changes that i made associated with the src/ layout. but overall the output wheel is looking generally ok and things are installing, running as they should using nox as the foundation for all of our dev steps.

lwasser:
ok i'll run pre-commit after a review. but if someone wants to take this for a test drive that would be great. i may have missed things as there were a lot of small changes that were required to get this all working... so let me know what you catch!!

enadeau (commented):
Hey very nice work. I have to. I've cloned and could at least run the test and build the doc. I don't know if there's anything specific you think I should try.

Otherwise, I have two comments:
1. I think it's necessary to run mypy on multiple version, as it will always return the same result.
2. I wonder if we should also move the tests outside the of the stravalib folder. The standard setup is to have the tests folder alongside the src folder

lwasser:
> * I think it's necessary to run mypy on multiple version, as it will always return the same result.

@enadeau can you please be a bit more specific. do you mean it's not necessary to run it across versions? the reason why i thought we should was because previously we were finding some annotation types failed on some python versions. I think there are still a few things that 3.9 won't support right? or am i wrong? i'm happy to move it back to a single version build if we really don't need it. i'm just explaining my rational which may or may not be correct! 

Also are you referring to the action / nox build or both? with a bit more info i can adjust the pr accordingly.


> * I wonder if we should also move the tests outside the of the stravalib folder. The standard setup is to have the tests folder alongside the src folder

We could. there is a lot of debate around this in the packaging community right now. if we do move it i can make sure that setuptools atleast bundles tests in our SDist. but technically they really aren't needed in our wheel given this is a pure python package. some feel strongly tests belong in the code dir because it then ships in all builds automagically. and others argue for it as it's own separate folder for other numerous reasons which are mainly driven by organization and structural preferences. i don't have a strong opinion. If you'd  prefer it, i can move it and then set things up so tests ship in the SDIst only, I can!

enadeau:
### About mypy:

The type checking is not dependent on python version. Mypy is doing static analysis, meaning it is not running any of our code. The results will be influenced by the target version we set but not the version of python that runs mypy.

It's best to think of it as the other linter. It wouldn't make sense to run flake8, black or isort on multiple python versions

If we use python feature from version >3.9 the unit test have to catch this

### About the tests folder:

I didn't know there is big debate about that. I've just always seen in outside the code in the project I worked on. Never quite seen the point of having them package in the distribution as well.

I don't really have a strong opinion on it either so let's go the easiest way and to nothing for now?

enadeau:
> I think it's necessary to run mypy on multiple version, as it will always return the same result.

Also realizing that I meant to says "I think it is NOT necessary"...

lwasser:
ok great... so 

1. i'll pull the multi-python version run from nox and the ci build - no problem
2. regarding tests - if that can wait (i'm happy to open an issue) that would be good. i just ran into a weird tricky situation where we're importing a fixture from conftest but i don't know how to do that if the tests are not a part of the package.

so that is failing now. 

As for the debate - i think i've been in too many packaging discussions with people debating about what packages should look like 😆  but again i don't have an opinion here at all. if we decide to make that move which we can. I'll open another pr just because i'm not sure about the module import part but i'm sure i can figure it out (or maybe you know?). pushing a few fixes in a bit here...

lwasser:
ok @enadeau !! thank you for the mypy explanation. I've reverted the mypy run to be single python version only. And i've left tests in the src/stravalib dir for now and will look into conftest imports at another time.  let me know if you catch anything else!! ✨

enadeau (approved):
I don't have anything else to say. LGTM

lwasser:
great!! i'll merge this now. then will do a test install from test-pypi!! just to double check everything.

Issue #432, opened 2023-09-01
BUG: AlpineSki segments not validating (again)


JohnScolaro:
### Stravalib version checks

- [X] I have tested this in a new clean environment with only stravalib and core python files.

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/stravalib/) of stravalib.

- [X] I have confirmed this bug exists on the [main branch](https://github.com/stravalib/stravalib) of stravalib.

### What operating system are you seeing the problem on?

Windows

### What version of python or you running?

```python
3.10.12
```

### Reproducible Example

```python
activity = client.get_activity(XXXXXXX, include_all_efforts=True) # Activity must be an AlpineSki with segments
```

### Issue Description

In stravalib v1.4, the issue previously reported here: https://github.com/stravalib/stravalib/issues/368 was re-introducted. Segments can be for activity type "AplineSki" and probably other activities too. Pydantic has an issue with that. It works in v1.3.3.

```
Traceback (most recent call last):
  File "/home/john/code/test_stravalib/test.py", line 18, in <module>
    activity = client.get_activity(8684395669, include_all_efforts=True)
  File "/home/john/code/stravalib/stravalib/client.py", line 619, in get_activity
    return model.Activity.parse_obj({**raw, **{"bound_client": self}})
  File "pydantic/main.py", line 526, in pydantic.main.BaseModel.parse_obj
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 6 validation errors for Activity
segment_efforts -> 0 -> segment -> activity_type
  unexpected value; permitted: 'Ride', 'Run' (type=value_error.const; given=AlpineSki; permitted=('Ride', 'Run'))
segment_efforts -> 1 -> segment -> activity_type
  unexpected value; permitted: 'Ride', 'Run' (type=value_error.const; given=AlpineSki; permitted=('Ride', 'Run'))
segment_efforts -> 2 -> segment -> activity_type
  unexpected value; permitted: 'Ride', 'Run' (type=value_error.const; given=AlpineSki; permitted=('Ride', 'Run'))
segment_efforts -> 3 -> segment -> activity_type
  unexpected value; permitted: 'Ride', 'Run' (type=value_error.const; given=AlpineSki; permitted=('Ride', 'Run'))
segment_efforts -> 4 -> segment -> activity_type
[...]

PR #369, opened 2023-06-04
Add field override for class Segment

solorisx:
closes #368 

*The text above will ensure the issue will be closed when this PR is merged*

## Description

Added a field override in class Segment to support all activity types as the actual Strava API does provide segment efforts for any GPS based activity even though the documentation only states Run and Ride as allowed types.

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update
- [ ] Other (please describe)

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [x] This change doesn't require tests

## Did you include your contribution to the change log?

- [x] Yes, `changelog.md` is up-to-date.

I am not sure whether this requires additional tests, it passes all integration test and I did test it with activities including segments other than Run or Ride (I was not able to get the functional tests to pass even before the change).

PR #434, opened 2023-09-01
Relax Segment ActivityTypes

JohnScolaro:
closes https://github.com/stravalib/stravalib/issues/432

## Description

See https://github.com/stravalib/stravalib/issues/432 for a detailed explanation of the issue. This is a fix I made that works, but I don't know if it's the correct way of doing it, so I'm putting it here for review. Essentially when downloading activities (that aren't rides and runs) that have segments, Pydantic fails to validate the activities because it's expecting the `ActivityType` in the `Segment` in the `SegmentEffort` in the `Activity` to be either ride or run. Since Strava has been adding segments for new activity types, this breaks.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -935,7 +935,7 @@ class Segment(
     map: Optional[Map] = None
     athlete_segment_stats: Optional[AthleteSegmentStats] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
-    activity_type: Optional[Literal["Ride", "Run"]] = None
+    activity_type: Optional[ActivityType] = None
 
     # Undocumented attributes:
     start_latitude: Optional[float] = None
@@ -1041,9 +1041,9 @@ class Activity(
     end_latlng: Optional[LatLon] = None
     map: Optional[Map] = None
     gear: Optional[Gear] = None
-    best_efforts: Optional[list[DetailedSegmentEffort]] = None
-    segment_efforts: Optional[list[DetailedSegmentEffort]] = None
     # Ignoring types here given there are overrides
+    best_efforts: Optional[list[BestEffort]] = None
+    segment_efforts: Optional[list[SegmentEffort]] = None
     splits_metric: Optional[list[Split]] = None  # type: ignore[assignment]
     splits_standard: Optional[list[Split]] = None  # type: ignore[assignment]
     photos: Optional[ActivityPhotoMeta] = None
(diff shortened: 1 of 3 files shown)


lwasser:
thank you for this pr @JohnScolaro !! i just fixed a merge conflict and gave the pr access to run Ci! i understand generally what the good of this is about - we had just added those typed literal values for run and ride but it seems like that then causes some issues with interfacing with other activity types (which presumably aren't documented in the API?)

src/stravalib/model.py
```diff
@@ -938,7 +938,7 @@ class Segment(
     map: Optional[Map] = None
     athlete_segment_stats: Optional[AthleteSegmentStats] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
-    activity_type: Optional[Literal["Ride", "Run"]] = None
+    activity_type: Optional[ActivityType] = None
```
lwasser, line 941: @jsamoocha i'm pretty sure it's ok for us to relax this - as it's pretty much the way things were previously. but i also remember another part of this module where i had to add a literal type for one of the methods. i think @enadeau helped me with that. so i just want to make sure this fix will actually work

lwasser: oh perhaps that was athlete type?

jsamoocha: This same fix was applied in #369. It led to typing errors (which we see now in CI). I have no idea ATM if we have to choose between two "evils" or if there's a (typing) workaround.

enadeau:
I don't see any more elegant solution at the moment. Of course it would be nice to see Strava fix the open api specification, but there's not much we can do about that. I think this is the best fix in the meantime.

I've added a test to make sure we don't regress this issue again and fixed the last mypy issue.

If another maintainer approves the change, I'll merge it and include that in the 1.5 release.

src/stravalib/model.py
lwasser: ahhh ok thanks. so we should merge following @enadeau commits below.

lwasser (approved):
ok so we are back to the original work around with a test to make sure this works as expected. thank you @JohnScolaro and @enadeau for this. i'm approving it.

JohnScolaro:
Thanks guys :)

enadeau:
Merged :rocket:  Thanks @JohnScolaro for your contribution

Issue #437, opened 2023-09-10
BUG: Mypy types possibly incorrect


JohnScolaro:
### Stravalib version checks

- [X] I have tested this in a new clean environment with only stravalib and core python files.

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/stravalib/) of stravalib.

- [X] I have confirmed this bug exists on the [main branch](https://github.com/stravalib/stravalib) of stravalib.

### What operating system are you seeing the problem on?

Windows

### What version of python or you running?

```python
3.10.12
```

### Reproducible Example

```python
from stravalib.client import Client

client = Client()
authorize_url = client.authorization_url(
    client_id=YOUR_CLIENT_ID, redirect_uri="https://active-statistics.com"
)

print(authorize_url)

token = input()

token_response = client.exchange_code_for_token(
    client_id=YOUR_CLIENT_ID,
    client_secret=YOUR_CLIENT_SECRET,
    code=token,
)
access_token = token_response["access_token"]
refresh_token = token_response["refresh_token"]
expires_at = token_response["expires_at"]

# Now store that short-lived access token somewhere (a database?)
client.access_token = access_token

activity = client.get_activity(ACTIVITY_ID, include_all_efforts=True)

elapsed_time = activity.elapsed_time
if elapsed_time is not None:
    elapsed_time_in_seconds = elapsed_time.total_seconds()

print(elapsed_time_in_seconds)
```

### Issue Description

Hey guys, I'm not sure if this is an issue or not but:

Because of how the `_field_conversions` work in Stravalib, I can have perfectly valid code like in the example I pasted above, and mypy fails. Above, I query an activity, then get the elapsed time in seconds from it. The `Activity` object is typed to hold the `elapsed_time` attribute as an integer. However, because (I'm not sure the exact process here, but) `_field_conversions` is turning it into a `timedelta` when accessed, when I run mypy I get the following errors:

```
test.py:32: error: "int" has no attribute "total_seconds"  [attr-defined]
```

Am I missing something here, or should the `Activity` types somehow be changed so I don't get hundreds of mypy errors from working code?

### Expected Behavior

I expect mypy to pass when working with Activities in a valid manor.

### Your environment

arrow==1.2.3
certifi==2023.7.22
charset-normalizer==3.2.0
idna==3.4
mypy==1.5.1
mypy-extensions==1.0.0
Pint==0.22
pydantic==1.10.9
[...]

PR #440, opened 2023-09-12
fix type for timedelta element on models

enadeau:
Partial fix for #437

## Description

When an attributes represents a duration we parse it directly with pydantic
instead of using field conversion to convert the value when attribute is
accessed.

This allow to have the right type annotation for those field.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/field_conversions.py
+++ b/src/stravalib/field_conversions.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Sequence
-from datetime import timedelta
 from functools import wraps
 from typing import Callable, TypeVar
 
@@ -53,14 +52,6 @@ def enum_values(enums: Sequence[ActivityType | SportType]) -> list[str | None]:
     # it doesn't work for lists of enums.
     # See https://github.com/pydantic/pydantic/issues/5005
     return [enum_value(e) for e in enums]
-
-
-@optional_input
-def time_interval(seconds: int) -> timedelta:
-    """
-    Replaces legacy TimeIntervalAttribute
-    """
-    return timedelta(seconds=seconds)
 
 
 @optional_input
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -13,7 +13,7 @@ related entities from the API.
 from __future__ import annotations
 
 import logging
-from datetime import date, datetime
+from datetime import date, datetime, timedelta
 from functools import wraps
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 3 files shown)


src/stravalib/model.py
```diff
@@ -946,7 +944,7 @@ class Segment(
     start_longitude: Optional[float] = None
     end_longitude: Optional[float] = None
     starred: Optional[bool] = None
-    pr_time: Optional[int] = None
```
lwasser, line 949: @enadeau  i'm just curious - do you think those int assignments were a part of the unexpected breaking behavior? i may have added that missing it was actually an object returned rather than a numeric value

src/stravalib/field_conversions.py
```diff
@@ -55,14 +54,6 @@ def enum_values(enums: Sequence[ActivityType | SportType]) -> list[str | None]:
     return [enum_value(e) for e in enums]
 
 
-@optional_input
-def time_interval(seconds: int) -> timedelta:
```
lwasser, line 59: i have two questions

1. can you help me understand what this function's purpose was before - i understand it is a legacy behavior function BUT, all it did was return an timedelta object. 
2. If we remove it altogether - are we breaking something for some 
i see it has an optional decorator but i'm still wrapping my head around what removing it means for users (and thinking about what it will look like to migrate to pydantic 2.0 and perhaps drop some of this legacy behavior support

lwasser: perhaps it is true that no one would call this directly but it is not defined as a private method.

lwasser:
thanks for this @enadeau - i just had a few questions about this and am thinking about how we migrate while looking at it. i think @jsamoocha should look at it before we merge. just to make sure i'm not missing anything but it seems reasonable to me other than asking whether anyone might call that method directly in an app or something...  (i think the answer is no but i wanted to ask). And we do have a version they can pin against as well even if this is a minor breaking change for someone.

src/stravalib/field_conversions.py
enadeau: 1. The only thing it is doing is turning seconds into a timedelta it was probably needed to pass to the field conversion. Because of the decorator, it is actually doing a bit more. If you pass it None it returns None
2. In theory it could break someone's if they imported and used this function directly. It's really unlikely to be the case I think.

src/stravalib/model.py
enadeau: No they've always been like that. I just check the blame and this line was not modified when you typed to model file. It's just that now that we publish the type mypy is noticing.

lwasser (approved):
ok i think this looks ok to me. @jsamoocha are you ok with us merging this? i think it's relatively safe.

jsamoocha (approved):
This looks good. Getting rid of these custom `field_conversions` simplifies the code 👍

PR #446, opened 2023-10-24
Read Rate Limits

### LINKED ISSUE #393: Upcoming changes to API rate limiting
Strava's new [developer program](https://communityhub.strava.com/t5/developer-knowledge-base/our-developer-program/ta-p/8849) introduces a new "read rate limit", which is for specific endpoints. This means that our existing rate limiters should be modified to:

1. Handle the new headers from a response
2. Execute based on which endpoint is called (which requires a redesign of our rate limiter approach)

It is not clear when Strava will enforce these new rules.

jsamoocha:
closes #393

## Description

This change allows apps enrolled in Strava's new (per 2023) Developer Program to use rate limiters based on the newly introduced "ReadRateLimits."

In addition, this change removes two legacy rate limiters in the library that either ignored Strava's rate limits or did not throttle requests. The existing default `XRateLimitRule` is now replaced by the `SleepingRateLimitRule` (with priority set to "high," which leads to similar behavior as in the previous version).

A potentially breaking change is that the `RateLimiter`'s `__call__()` method now has an additional `method` argument representing the HTTP method used for that request. Existing custom rate limiters from users must be updated to this change.

## Type of change

## Does your PR include tests

- [x] Yes


docs/reference/utilities.rst
```diff
@@ -21,11 +21,7 @@ Summary Classes
...
    :toctree: api/
 
-   XRateLimitRule
    SleepingRateLimitRule
-   RateLimitRule
    RateLimiter
    DefaultRateLimiter
    RequestRate
-   LimitStructure
-   LimitsStructure
```
jsamoocha, line 31: I'm not sure if these are the only changes I need to make to the docs. Building the docs locally still fails with errors caused by these classes not existing anymore.

src/stravalib/protocol.py
```diff
@@ -314,7 +319,7 @@ def _request(
         raw = requester(url, params=params)  # type: ignore[operator]
         # Rate limits are taken from HTTP response headers
         # https://developers.strava.com/docs/rate-limits/
-        self.rate_limiter(raw.headers)
+        self.rate_limiter(raw.headers, method)
```
jsamoocha, line 322: Passing the method to the rate limiter will allow us to choose wether to use the overall rate limits or the specific read rate limits (for GET requests)

src/stravalib/tests/unit/test_limiter.py
jsamoocha, line 1: The main changes here are a rewrite of the remaining relevant tests into pytest.

docs/reference/utilities.rst
lwasser: i can have a look at the docs if you'd like. sorry i've been so quiet. the past few weeks have been really busy w work travel and then catchup when i got back. i'll try to check this out this weekend / early next week.

lwasser:
@jsamoocha just so i understand. it looks like the rate limiter should be called as it was previously. but this just adapts to strava's new implementation? previously i had a hard time with getting it to work as you mentioned in an issue about this so i wanted to ask that first before i play with the pr.

also it looks like all of our checks are  green and the [docs build](https://stravalib--446.org.readthedocs.build/en/446/) are you still seeing documentation errors locally?

lwasser:
ahhh i see [these warnings](https://github.com/stravalib/stravalib/actions/runs/6628660521/job/18006243504?pr=446#step:8:185) - perhaps that is what you're seeing locally?

docs/reference/utilities.rst
lwasser: ok the docs was a quick check as i suspected a simple solution. when you build the docs the API docs are build on top of stub files that are automagically generated by sphinx when we build. These files live in the docs/reference/api dir. When we remove methods, classes etc we can do 1 of two things. 

1. we can clean the entire API directory (probably best) and let sphinx autogenerate new API stubs for rendering
2. we can just remove the files related to removed methods, etc

so below that XRateLimitRule.rst file is still in our API stubs folder which throws an warning because it's a file that represents a docstring that doesn't exist because it was removed. 

https://github.com/stravalib/stravalib/blob/main/docs/reference/api/stravalib.util.limiter.XRateLimitRule.rst

Locally i just cleaned out the entire directory and let it rebuild and the warnings went away. 

i can push a fix if you are ok with that. It will add a bunch of modified/removed rst files in this pr however. are you ok with my doing that? or do you prefer to do it?

also in a future fix if we wanted i could add a step to our [nox clean_docs](https://github.com/stravalib/stravalib/blob/main/noxfile.py#L86) function to clean that directory as well? we'd want to remember to rebuild is the only thing and push the stubs up after

lwasser: on another note i just asked a colleague whether these stubs need to actually be in version control at all. maybe we can just remove them.

jsamoocha: Thanks for your other PR #447, that solved the issue!

jsamoocha:
> it looks like the rate limiter should be called as it was previously. but this just adapts to strava's new implementation?

I added a small section to the documentation on using and configuring rate limiting. Existing users or client apps that didn't do any customization re. rate limiters will experience no changes other than client instances blocking instead of raising exceptions when rate limits are reached.

Whoever uses custom rate limiter classes has to change its `__call__()` method to accept the additional `method` argument. This is something we need to document when making the release.

I hope this clarifies.

enadeau:
Great update of this whole ratelimiter buisness that was really needed! I think the breaking change should be mentioned in the changelog. It feels like the best place to put it

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
    # 22530.8
    print(float(unithelper.miles(activity.distance)))
    # 13.9999900581
+
+
+Rate Limits
+===========
+
+Strava imposes rate limits on the usage of its API. This means that the number of
```
lwasser, line 132: THANK YOU for updating the docs!!
i'm trying to wrap my head around this
Strava says this - https://developers.strava.com/docs/rate-limits/ : 

Strava API usage is limited on a per-application basis using both a 15-minute and daily request limit. The default rate limit allows 100 requests every 15 minutes, with up to 1,000 requests per day. As an application grows, its rate limit may need to be [adjusted](https://developers.strava.com/docs/rate-limits/#adjustment).

is your language that it's not fixed related to a large use app may be provided with more bandwidth / a higher rate limit? If so can we use that language or link to the docs for more specifics? i'll make a suggestion if you think it's reasonable

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
...
+requests sent to Strava has an upper limit per 15 minutes and per day. These limits
+are not fixed but depend on the client app.
```
lwasser, line 134:
```suggestion
requests sent to Strava has an upper limit per 15 minutes and per day. Per strava's documentation: 

> The default rate limit allows 100 requests every 15 minutes, with up to 1,000 requests per day.

These limits
are not fixed but depend on the "size" of client app. Strava _may_ chose to adjust rate limits for apps as they grow.
```

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
...
+
+When initializing a Client instance, the default rate limiter allows requests until
+the short - or daily limits are reached, after which it will wait until the end of
+the 15-minute or day period.
```
lwasser, line 138:
```suggestion
When initializing a `stravalib.Client` instance, the default rate limiter allows requests until
the short - or daily limits are reached. Once limits are reached. the client object will wait until the end of
the 15-minute or day period.
```

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
...
+
+In case you want to configure the limiter to throttle requests (i.e., making sure
+the time between requests for the remaining period is evenly spread), you can
+initialize the client object as::
+
+    client = stravalib.Client(
+        my_access_token, rate_limiter=DefaultRateLimiter(priority='medium')
+    )
```
lwasser, line 146:
```suggestion
   from stravalib.util.limiter import DefaultRateLimiter
    client = stravalib.Client(
        my_access_token, rate_limiter=DefaultRateLimiter(priority='medium')
    )
```

lwasser: when i ran this i copied your code and then realized i needed to import the `DefaultRateLimiter` object.

docs/get-started/overview.rst
lwasser: @jsamoocha i'm playing with this locally and am really confused by the rate limits. i'll add a comment at the end of this review as i may just not understand how this works. I also have a paid account AND my developer app is fairly old so i'm not sure how new rate limits compare to older ones.

lwasser (commented):
Ok - i created a workflow locally and captured streaming data which is where i've hit rate limits in the past. 

It seems like the rate limiter does work now - as i've been running / collecting data now for over 15 minutes and am at over 1100 requests for streaming data (if i'm counting a request correctly that is). 

I'm super confused so please help me understand so i can better contribute to the docs here! 

1. strava says there's a limit of 100 requests every 15 minutes. BUT would a request to grab streaming data from 1 activity == 1 request? and 
2.  How do the 1000 requests per day work?

Right now i'm well over the 1000 requests per day (i'm capturing data from 1300 activities using `client.get_activity_streams(act, types=types)`

what is awesome is that it is just running without a point of failure. but what i don't understand is haven't i already surpassed the daily limits? and should it pause after 100 each time or is the limiter spreading out the requests to allow 100 across 15 minutes so it can continue collecting information from the api. 

i think once i understand this we might want to update the docs just a bit. but overall things are running really well for me locally. i haven't run tests yet but can look at them in CI and see they are running. 

thank you Jonatan!

lwasser (approved):
Ok! this ran through the night. and it seems to be working as expected i think

```
1677 9183894241
Short term API rate limit exceeded
1678 9189165166
Short term API rate limit exceeded
9189165166 429 Client Error: Too Many Requests [Rate Limit Exceeded: [{'resource': 'Application', 'field': 'overall rate limit', 'code': 'exceeded'}]]
1679 9197271525
Short term API rate limit exceeded
9197271525 429 Client Error: Too Many Requests [Rate Limit Exceeded: [{'resource': 'Application', 'field': 'overall rate limit', 'code': 'exceeded'}]]
1680 9201523900
1681 9202413990
```

and then lower down

```
1696 9296389477
9296389477 ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
1697 9300858641
```

each number above represents a request number (count) associated with an activity number

I think we can merge this but it might be nice to clarify the rate limits that are in the docstring which i missed last night. I read what was on the website incorrectly and now see 30000 requests per day with 600/minute. 

> Strava API usage is limited on a per-application basis using a short term,
  15 minute, limit and a long term, daily, limit. The default rate limit allows
  600 requests every 15 minutes, with up to 30,000 requests per day.

 > This limit allows applications to make 40 requests per minute for about
  half the day.

So i didn't hit the daytime limit (that would be hard) but did hit the 15 min limit and it kept going. it also seemed to keep going when my connection went bad.

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
...
+
+The ``medium`` priority ensures that requests are throttled to comply with the
+15-minute limit, while the ``low`` priority complies with the daily limit.
```
lwasser, line 149:
```suggestion
15-minute limit, while the ``low`` priority complies with the daily limit.

When you hit a rate limit threshold, a message will be returned as follows

> Short term API rate limit exceeded

the error throw will be 

```
429 Client Error: Too Many Requests [Rate Limit Exceeded: [{'resource': 'Application', 'field': 'overall rate limit', 'code': 'exceeded'}]]
```

The limiter will then pause and allow more requests to be implemented once the threshold is below the required value.

Comment (lwasser) [reply docs/get-started/overview.rst:149]: i'm not sure if this is quite right but just tyring to help the user understand what happens when those limits are applied. i generally understand how this works now!

Comment (lwasser) [inline docs/get-started/overview.rst:149]:
```suggestion
There are three options for throttling with stravalib - ``low``, ``medium`` and ``high``.  

The ``low`` priority complies with the daily limit. The ``medium`` priority ensures that requests are throttled to comply with the
15-minute limit.
```

lwasser: take or leave my doc comments :) this can be merged once those are cleared up! i'm done reviewing now :)

docs/get-started/overview.rst
jsamoocha: The 600/15min and 30k/day limits are not the case anymore (although some Strava doc still mentions it). Although "legacy" apps (probably the one you use for testing) still have these default limits. You can check your app's limits through the settings page (https://www.strava.com/settings/api, which isn't available in Strava's mobile app). I'll update the documentation to mention this.

docs/get-started/overview.rst
```diff
@@ -124,3 +124,26 @@ or just cast to a numeric type (e.g. float).::
...
+
+
+Rate Limits
+===========
+
+Strava imposes rate limits on the usage of its API. This means that the number of
+requests sent to Strava has an upper limit per 15 minutes and per day. These limits
+are not fixed but depend on the client app.
+
+When initializing a Client instance, the default rate limiter allows requests until
```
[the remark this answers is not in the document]
jsamoocha: Handling the daily limit is not different from handling the 15min limit.

jsamoocha:
Thanks for reviewing @lwasser ! Somehow, adding your changes to the docs made mypy unhappy. Locally, it runs fine:

```
❯ python --version
Python 3.10.10
❯ nox -s mypy
nox > Running session mypy
nox > Re-using existing virtual environment at .nox/mypy.
nox > python -m pip install '.[all]'
nox > python -m pip install -r requirements.txt
nox > mypy
Success: no issues found in 12 source files
nox > Session mypy was successful.
```

Any idea what's causing this? If you're ok with it, I can still merge while ignoring the issue.

enadeau:
I just took a look at it and I suspect you are not running the newest version of mypy locally while CI is. If you update your local version you should get the warning.

 The warning are however unrelated to PR so we can save here and address them in an other PR

jsamoocha:
Thanks for checking, @enadeau: it was indeed a new version of mypy causing the warnings. I'll merge now.

lwasser:
awesome! and thanks for the clarification @jsamoocha i am fairly confused by the actual rate limits now given what i read online and such. i'll check out my app later. it is rather old / legacy so that makes sense that it has older requirements attached to it.

PR #447, opened 2023-11-04
Fix: remove autogen api dir from version control

lwasser:
This doesnt' address a specific issue. Rather it should fix the problem that @jsamoocha is having with #446  where warnings are being raised because api stubs for methods we are removing existing in the api/ directory. 

## Description

After doing a bit of research it seems to be standard practice to not include those files in the directory. they will be build locally but we don't have to commit them. As such, this pr
1. adds the dir to our .gitignore file and
2. removes those files from git cache.
3. i'm not sure if everyone now needs to run `git rm docs/reference/api/*` or not to remind git to not check these into versioning again. i removed them however in this pr so hopefully it will be easy for everyone when building docs

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes


.gitignore
```diff
@@ -27,3 +27,6 @@ stravalib/docs/api
 .vscode/
 stravalib-2023/
 .nox/*
+
+# Ignore Sphinx auto-generated API stubs #
+docs/reference/api
```
lwasser, line 32: this is really the only change in ths PR. the other 70 files are api stubs that i'm removing from versioning given sphinx will always rebuild them. this will help us maintain the api docs in a cleaner format.

lwasser:
pre-commit.ci autofix

jsamoocha:
Thanks, now I understand what was happening with the doc warnings. I'll merge and rebase my #446 branch on this change.

PR #454, opened 2024-01-16
Adds custom validator for ActivityType and SportType

### LINKED ISSUE #441: ENH: Graceful handling of Pydantic validation errors (for unexpected sports type / activity type)
### Feature Type

### Problem Description

For unexpected activity/sport types, parsing `Activity` objects crashes with a Pydantic `ValidationError`. On (rare) occasions, these unexpected (and undocumented) types pop up in responses from Strava.

### Feature Description

We can add custom types and validators for (at least) the mentioned fields `activity_type` and `sports_type`. For unexpected types, we can then parse them and log a warning instead of crashing.

### Alternative Solutions

Unfeasible: requesting Strava to always keep the documentation 100% up-to-date.

### Additional Context

_No response_

### Code of Conduct

jsamoocha:
closes #441 

## Description

Some unexpected activity types caused crashes when parsing responses from Strava. This change makes sure
there will be no pydantic `ValidationError`s in these cases by replacing the unexpected types by a default `Workout`.

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Does your PR include tests

- [X] Yes


src/stravalib/model.py
```diff
@@ -952,6 +976,7 @@ class Segment(
         "elevation_high": uh.meters,
         "elevation_low": uh.meters,
         "total_elevation_gain": uh.meters,
+        "activity_type": enum_value,
```
jsamoocha, line 979: This tiny change is there to give the segment's `activity_type` the same behavior as the activity's `type` (i.e., returning the string representation of the nested type). This is for backward compatibility with the legacy model.

src/stravalib/tests/unit/test_model.py
```diff
@@ -225,6 +225,23 @@ def test_backward_compatible_attribute_lookup(
         assert not hasattr(lookup_expression, "bound_client")
 
 
+@pytest.mark.parametrize(
```
lwasser, line 228: gosh i wish i knew about pytest.mark.parametrize a long time ago 🙃  this is so cool that you can just pass it many options without having to write separate methods (or without creating some sort of loop).

lwasser (commented):
@jsamoocha thanks for the quick fix for this!! i didn't test it locally but looking at the code and associated tests it all looks fine to me. 

what does everyone think about bumping a version given our unreleased fix list has gotten long and this particular issue seems to be impacting folks production builds?

jsamoocha:
Thanks for reviewing! A new release seems like a good idea. I may have some time for it this afternoon.

PR #456, opened 2024-01-21
Use Pydantic V2 with V1 compatibility mode

soh55:
closes #379

_The text above will ensure the issue will be closed when this PR is merged_

## Description

I added [Pydantic V1 backward compatibility features](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features) so users can move to a newer Pydantic version. Need to find some time to perform the actual migration.

Reference issue: #379 

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update
- [ ] This is a infrastructure update (docs, ci, etc)
- [ ] Other (please describe)

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [X] This change doesn't require tests

## Did you include your contribution to the change log?

- [x] Yes, `changelog.md` is up-to-date.

_If you are having a hard time getting the tests to run correctly feel free to
ping one of the maintainers here!_

PR #459, opened 2024-01-22
Add more robust logic for importing pydantic V2's V1 compatibility pa…

soh55:
## Description

PR #456 introduced backwards compatibility for Pydantic V2 users. However, I goofed and didn't account for Pydantic V1 users. Here is a more robust import logic.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes
- [X] This change doesn't require tests -> Existing tests pass


src/stravalib/strava_model.py
```diff
@@ -6,7 +6,14 @@
 from datetime import datetime
 from typing import Literal, Optional
 
-from pydantic.v1 import BaseModel, Field
+from stravalib.pydantic_version import pydantic_major_version
```
lwasser, line 9: @soh55 this is the file that @jsamoocha was referring to that is autogenerated via the swagger API - [see here](https://github.com/stravalib/stravalib/pull/424) any changes made to this file will be overwritten until we update the actual model generation code workflow that jonatan designed and setup. (which works really well)

i think we all missed that in your previous PR because we were excited about a way to support both versions. 

i suspect youre keen on this change because we've pinned pydantic in our dependencies to avoid breaking changes in 2.0. we did that because it will take some time to upgrade the entire library to the new 2.0 API.

i hope that makes sense!

lwasser:
hey @soh55 i am going to close this pr. We are migrating now to v2 on a pydantic-v2 branch! thank you for this contribution.

Issue #279, opened 2022-11-30
Add support for sport_type activity attribute and deprecate type activity attribute


jsamoocha:
From the Strava API changelog:

>June 15, 2022
>
>Introduction of [activity](https://developers.strava.com/docs/reference/#api-Activities) sport_type. This is the preferred field to use moving forward, as opposed to type, which is now considered deprecated. Please note: this change is backward compatible and type will still be supported. The SportType enumeration is distinct from ActivityType in that it includes new types (e.g. MountainBikeRide). In requests where both type and sport_type are present, type will be ignored.

jsamoocha:
#362 Adds support for `sport_type` in the `update_activity()` method. Remaining TODO:

- [ ] Support `sport_type` (and deprecate `activity_type`) in `create_activity()`
- [ ] Add deprecation warnings when using the `activity_type` attribute in `model` module (see `BackwardCompatibilityMixin.__getattribute__()`)

lwasser:
i'm attempting the first part of this. it seems to be generally working in think but i don't fully understand how to adjust the tests to ensure things work with sport_type and variations of sport type vs "type" or activity_type in our code. .

PR #471, opened 2024-02-18
Fix: validate sport / activity type & add sport type to create_activity method

lwasser:
addresses in part - #279

## Description

I'm trying to fix 279 opened by @jsamoocha however i'm a bit confused and need some direction with the test suite. 

What this does:
✅  it adds a small validate method helper so that the checks for sport type and activity type can be reused easily (assuming a python DICT format).
✅ it creates reusable docstring elements to ensure the list of types is only maintained in one place 
✅ it adds sport type to the create_activity method 

I'm struggling with tests. once i get those running i can test the create activity method further. i haven't used pytest parametrize before 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

i can't ping myself 😆 so can someone kindly have a look and direct me towards what i have to do to get the tests running. i suspect its related to adding new variables in create_activity but i'm not sure how i should go about this AND if i should also create some small tests for the small helper that i added outside of the tests setup. many thanks 🙌 for any direction!


lwasser:
i tried adding elements to here thinking it would need additional test runs that included sport type but i'm  confused as to how this works with pytest

```python
@pytest.mark.parametrize(
    "extra_create_kwargs,extra_expected_params,expected_exception",
    (
        ({}, {}, None),
        ({"activity_type": "run"}, {"type": "run"}, None),
        ({"activity_type": "Run"}, {"type": "run"}, None),
        ({"activity_type": "sleep"}, {}, ValueError),
        (
            {"start_date_local": datetime.datetime(2022, 1, 1, 10, 0, 0)},
            {"start_date_local": "2022-01-01T10:00:00Z"},
            None,
        ),
        (
            {"elapsed_time": datetime.timedelta(minutes=1)},
            {"elapsed_time": "60"},
            None,
        ),
        ({"distance": 1000}, {"distance": "1000"}, None),
        ({"distance": miles(1)}, {"distance": "1609.344"}, None),
        ({"description": "foo"}, {"description": "foo"}, None),
    ),
)
def test_create_activity(
    mock_strava_api,
    client,
    extra_create_kwargs,
    extra_expected_params,
    expected_exception,
):
    default_call_kwargs = {
        "name": "test",
        "activity_type": "Run",
        "start_date_local": "2022-01-01T09:00:00",
        "elapsed_time": 3600,
    }
    default_request_params = {
        "name": "test",
        "type": "run",
        "start_date_local": "2022-01-01T09:00:00",
        "elapsed_time": "3600",
    }
    call_kwargs = {**default_call_kwargs, **extra_create_kwargs}
    expected_params = {**default_request_params, **extra_expected_params}

```

lwasser:
also please note autodoc is not rendering my docstring variables - yet but i can work on fixing that once the tests actually run.

lwasser:
ok i *think* i figured out the test issue but i'd really like a close review on this before i do any more work. i'm setting it to "ready for review" but it's likely not ready to merge until it has a deep review and a lot of testing to make sure i didn't inadvertently break other things.

src/stravalib/client.py
```diff
@@ -656,11 +730,25 @@ def create_activity(
         ----------
         name : str
             The name of the activity.
-        activity_type : str
+        activity_type : str, default=None
```
lwasser, line 733: i set both of these "types" to optional. but we probably want to run a test on params that one or the other exists otherwise the user will get a bad call return

```python
stravalib.exc.Fault: 400 Client Error: Bad Request [Bad Request: [{'resource': 'Activity', 'field': 'type', 'code': 'missing_field'}]]
```
the above is useful but we could just run a check before the call is made and return an error then to be proactive.

src/stravalib/tests/integration/test_client.py
```diff
@@ -488,6 +488,8 @@ def _call_and_assert():
     (
         ({}, {}, None),
         ({"activity_type": "run"}, {"type": "run"}, None),
+        ({"sport_type": "TrailRun"}, {"sport_type": "TrailRun"}, None),
+        ({"sport_type": "funrun"}, {"sport_type": "funrun"}, ValueError),
```
jsamoocha, line 492: This is good, but just as with the test for `create_activity()`, we should add the case where both `activity_type` and `sport_type` are provided (and expect the same result as in the other test).

src/stravalib/tests/integration/test_client.py
```diff
@@ -512,19 +514,21 @@ def test_create_activity(
     extra_expected_params,
     expected_exception,
 ):
+    # Should we ensure required argument vals are in the default calls?
```
jsamoocha, line 517: Good question. I created these defaults out of laziness for not having to repeatedly mention the `activity_type` in each test case. But now that we're adding `sport_type`, these defaults are not valid anymore.

I suggest the following:

1. Since you already refactored the validation logic into a separate method, let's make a new test for `_validate_activity_type()` that goes over all possible combinations of provided/not-provided and valid/invalid types.
2. The existing tests for `create_activity()` and `update_activity()` could then be simplified by using single default values for `sport_type`.

jsamoocha (commented):
Thanks for this work! Good move to refactor the activity/sport type validation logic into a separate method. Let's apply this same logic to `update_activity()`, so both will behave consistently.

src/stravalib/client.py
```diff
@@ -638,12 +639,85 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
             "See https://developers.strava.com/docs/january-2018-update/"
         )
 
+    def _validate_activity_type(
+        self,
+        params: dict[str, Any],
+        activity_type: Optional[str],
+        sport_type: Optional[str],
```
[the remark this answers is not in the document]
lwasser: oops! fixed!

src/stravalib/client.py
```diff
@@ -638,12 +639,85 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
...
+            Possible values (case-sensitive): AlpineSki, BackcountrySki,
+            Badminton, Canoeing, Crossfit, EBikeRide, Elliptical,
+            EMountainBikeRide, Golf, GravelRide, Handcycle,
+            HighIntensityIntervalTraining, Hike, IceSkate, InlineSkate,
+            Kayaking, Kitesurf, MountainBikeRide, NordicSki, Pickleball,
+            Pilates, Racquetball, Ride, RockClimbing, RollerSki, Rowing, Run,
+            Sail, Skateboard, Snowboard, Snowshoe, Soccer, Squash,
+            StairStepper, StandUpPaddling, Surfing, Swim, TableTennis, Tennis,
+            TrailRun, Velomobile, VirtualRide, VirtualRow, VirtualRun, Walk,
+            WeightTraining, Wheelchair, Windsurf, Workout, Yoga
```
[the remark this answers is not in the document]
lwasser: that is so much better. the docs look like this now - 

Screenshot 2024-02-29 at 7 50 44 PM

src/stravalib/client.py
```diff
@@ -638,12 +639,85 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
...
+            params["sport_type"] = sport_type
+
+        return params
+
+    # TODO: according to the strava api docs we can add an optional trainer
+    # and commute param here with boolean 0/1 values
     def create_activity(
         self,
         name: str,
-        activity_type: ActivityType,
```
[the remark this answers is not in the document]
lwasser: ok. i didn't consider that but you are absolutely right. Do you agree however that it is what we should do if strava is moving away from activity type / type and over to sport_type? 

i'll make a note in the changelog that this is a breaking change now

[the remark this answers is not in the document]
lwasser: i added this to the changelog for now:

> - Fix: Add sport type to create_activity and create type validator method. NOTE: this fix contains a breaking change in `Client.create_activity()` activity_type is now an optional keyword argument rather than a required positional argument (@lwasser, #279)

src/stravalib/client.py
```diff
@@ -656,11 +730,25 @@ def create_activity(
         ----------
         name : str
             The name of the activity.
-        activity_type : str
+        activity_type : str, default=None
             The activity type (case-insensitive).
+            Deprecated. Prefer to use sport_type. In a request where both type
+            and sport_type are present, this field will be ignored.
+            See https://developers.strava.com/docs/reference/#api-models-UpdatableActivity.
             Possible values: ride, run, swim, workout, hike, walk, nordicski,
```
lwasser, line 738: I went ahead and refered to 

For possible values see: :class:`stravalib.model.Activity.TYPES` for these as well.

src/stravalib/client.py
```diff
@@ -656,11 +730,25 @@ def create_activity(
...
+            Possible values (case-sensitive): AlpineSki, BackcountrySki,
+            Badminton, Canoeing, Crossfit, EBikeRide, Elliptical,
+            EMountainBikeRide, Golf, GravelRide, Handcycle,
+            HighIntensityIntervalTraining, Hike, IceSkate, InlineSkate,
+            Kayaking, Kitesurf, MountainBikeRide, NordicSki, Pickleball,
+            Pilates, Racquetball, Ride, RockClimbing, RollerSki, Rowing, Run,
+            Sail, Skateboard, Snowboard, Snowshoe, Soccer, Squash,
+            StairStepper, StandUpPaddling, Surfing, Swim, TableTennis, Tennis,
+            TrailRun, Velomobile, VirtualRide, VirtualRow, VirtualRun, Walk,
+            WeightTraining, Wheelchair, Windsurf, Workout, Yoga
```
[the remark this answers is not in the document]
lwasser: fixed!!

src/stravalib/tests/integration/test_client.py
```diff
@@ -279,7 +279,7 @@ def test_get_activity_streams_series_type_unofficial(mock_strava_api, client):
         ({"sport_type": "TrailRun"}, {"sport_type": "TrailRun"}, None, None),
         (
             {"activity_type": "Run", "sport_type": "TrailRun"},
-            {"sport_type": "TrailRun"},
+            {"type": "run", "sport_type": "TrailRun"},
```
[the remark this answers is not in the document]
lwasser: hey @jsamoocha i need a bit more info here. 

i did implement that validate method here in [update_activity](https://github.com/stravalib/stravalib/pull/471/files#diff-533e37fefc1516b6ee971ab4cdd64d667e4e1038f0509aa67902eea16712675dR860) and also create_activity (if i'm understanding what you are suggesting ). 

i do understand what you're saying that if there are both items in a call, we remove `type` in favor of the supported `sport_type`. 

Are you suggesting that i remove that test case input?  or can you help me  better understand the task here?

src/stravalib/tests/integration/test_client.py
lwasser: ahhh ok. this i understand. i'll work on this over the weekend. if i do this - i suspect the comment above that i didn't fully follow will also be addressed. please let me know if i'm incorrect in that assumption!

jsamoocha (commented):
This is moving in the right direction, please let me know if you have any questions about rewriting the tests.

src/stravalib/tests/integration/test_client.py
```diff
@@ -539,6 +543,54 @@ def _call_and_assert():
...
+        (None, "junoDog", None, ValueError),
+        # Case where we have both sport and activity type provided
+        # This should ONLY return Trail run. we'd want a test for two things
+        ("TrailRun", "Run", "TrailRun", None),
+        # Need to work on a method code check for this one - it should return
+        # a error i think for create activity?
+        (None, None, None, ValueError),
+    ),
+)
+def test_validate_activity_type(
```
lwasser, line 562: @jsamoocha i haven't rewritten the other two tests - YET. but i wondered if you could look at this test to let me know if i'm headed in the right direction. i wasn't sure how to handle all of the various cases and ended up with a few conditionals. any feedback on this is welcome. i will next work on removing the various case scenarios from create_activity and update_activity as discussed. thank you.

src/stravalib/tests/integration/test_client.py
lwasser: ok got it. see my new test below. i'll work on rewriting the other tests in the next few days. I think i understand how things work now but i'd like your input on my new test first to ensure what i think i understand is indeed correct!!

src/stravalib/tests/integration/test_client.py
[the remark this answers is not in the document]
lwasser: i understand this now. i did re-add the pop to remove type in my validation helper. i will fix this test in the next few days.  ✨

lwasser:
argh. tests are now failing. 

this is ok because 
1. i have to go back and fix tests for create and update activity
2. the one test for our validator is failing because i haven't added logic for the case where a user tries to create an activity and neither type or sport_type  exist (if we want to test for that case that is.

src/stravalib/client.py
```diff
@@ -783,30 +833,6 @@ def update_activity(
...
-
-        if sport_type is not None:
-            if not sport_type in model.Activity.SPORT_TYPES:
-                raise ValueError(
-                    f"Invalid activity type: {sport_type}. Possible values: {model.Activity.SPORT_TYPES!r}"
-                )
-            params["sport_type"] = sport_type
-            params.pop(
-                "type", None
-            )  # Just to be sure we don't confuse the Strava API
```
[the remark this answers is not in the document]
lwasser: yes. good call. i did add this back into the validator.

src/stravalib/tests/conftest.py
lwasser, line 1: I moved conftest to the root of the tests dir so fixtures were available in the unit tests. i then moved the _validate tests to a unit test since that seemed like a better fit if we are only testing the validation step rather than testing it in situ.

src/stravalib/tests/integration/test_client.py
```diff
@@ -272,15 +272,18 @@ def test_get_activity_streams_series_type_unofficial(mock_strava_api, client):
         ({}, {}, None, None),
```
lwasser, line 272: @jsamoocha i do need help here. i'm a bit lost with respect to how the mocks are setup to interface as a mock API. 
i also was confused because the DeprecationWarning here being tested is not a part of our validation helper. Right now if  a user ONLY provides a activity_type to create an activity, it will throw a value error that sport type is missing. but for update activity we don't have a check. 

i guess i'm feeling confused about how to best implement this for both create and update activity. AND then i'm confused around how the tests are setup for these two methods. 

the validation helper tests do run now. Many thanks for any help that you can provide. i've only done a tiny bit of mocking so it's hard to rewrite this test as i don't fully understand what's happening.

lwasser:
tests are running! i think i generally understand what you were pointing out @jsamoocha as that is what i think i did to make everything run. BUT this still needs a hard review as i'm not confident in my test fixes.  and i'm also not sure about how i handled sport_type in the validation helper. it will break things for anyone trying to push a new activity with activity_type. many thanks!

src/stravalib/client.py
```diff
@@ -638,12 +638,88 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
             "See https://developers.strava.com/docs/january-2018-update/"
         )
 
+    def _validate_activity_type(
+        self,
+        params: dict[str, Any],
+        activity_type: str | None,
+        sport_type: str | None,
+        create: bool = False,
```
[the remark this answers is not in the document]
jsamoocha: Should we have this separate logic in the first place? In main, we still support providing only the `type` param to Strava (which is deprecated, but still supported). I favor having the same validation logic for both create - and update activity (as long as the Strava API doesn't require us to have separate logic for these two cases). This reduces complexity. Or maybe I'm missing some details?

src/stravalib/client.py
```diff
@@ -638,12 +638,87 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
...
+        Raises
+        ------
+        ValueError
+            If the activity_type or sport_type is invalid.
+        """
+
+        if create and not sport_type:
+            raise ValueError(
+                "The sport_type key is required to create a Strava activity"
+            )
```
jsamoocha, line 681: Referring to my remark above, should we really raise an error here (as the Strava API, per its documentation, supports the `type` parameter)?

src/stravalib/client.py
```diff
@@ -700,6 +771,12 @@ def create_activity(
         if distance is not None:
             params["distance"] = distance
 
+        params = self._validate_activity_type(
+            params, activity_type, sport_type, create=True
+        )
+        # TODO: throw warning if activity or sport_type not included
+        # in returned params
```
jsamoocha, line 778: Let `_validate_activity_type()` raise an error if neither are provided, as you indicated above:

> i set both of these "types" to optional. but we probably want to run a test on params that one or the other exists otherwise the user will get a bad call return

src/stravalib/tests/unit/test_validate_activity_type.py
```diff
@@ -0,0 +1,67 @@
...
+    parameters associated with a Strava activity."""
+
+    params = {
+        "name": "New Activity",
+        "start_date_local": "2024-03-04T18:56:47Z",
+        "elapsed_time": 9000,
+        "description": "An activity description here.",
+        "distance": 5700,
+    }
+    return params
```
jsamoocha, line 16: 👍, this is much better than the `default_call_kwargs` etc that I used in `test_create_activity`. Feel free to replace these with fixtures as well!

src/stravalib/tests/unit/test_validate_activity_type.py
```diff
@@ -0,0 +1,67 @@
...
+        # A keyerror should be returned
+        if sport_type and activity_type:
+            assert out["sport_type"] == expected_result
+            assert "type" not in out.keys()
+            # Run tests
+        # If only sport type is available
+        elif sport_type:
+            assert out["sport_type"] == expected_result
+        elif activity_type:
+            assert out["type"] == expected_result
```
jsamoocha, line 67: The above can probably be simplified to something like

```python
out_params = client._validate_activity_type(params, activity_type, sport_type)
assert params | expected_result == out_params
```
where `expected_result` would be a dict of the expected "type" param, e.g. `{"sport_type": "Run"}`.

src/stravalib/tests/unit/test_validate_activity_type.py
```diff
@@ -0,0 +1,67 @@
...
+    ),
+)
+def test_validate_activity_type(
+    params,
+    client,
+    activity_type,
+    sport_type,
+    expected_result,
+    expected_exception,
+    create,
```
jsamoocha, line 37: Given my other comments, unless I oversaw something, I think we can remove `create` as a parameter from this test.

src/stravalib/tests/unit/test_validate_activity_type.py
```diff
@@ -0,0 +1,67 @@
...
+        "description": "An activity description here.",
+        "distance": 5700,
+    }
+    return params
+
+
+@pytest.mark.parametrize(
+    "sport_type, activity_type, expected_result, expected_exception, create",
+    (
+        ("TrailRun", None, "TrailRun", None, False),
```
jsamoocha, line 22: As you mentioned somewhere, let's add a case here where neither `activity_type` nor `sport_type` are given, and we expect a `ValueError`.

src/stravalib/tests/integration/test_client.py
```diff
@@ -486,12 +488,10 @@ def _call_and_assert():
 @pytest.mark.parametrize(
     "extra_create_kwargs,extra_expected_params,expected_exception",
     (
-        ({}, {}, None),
-        ({"activity_type": "run"}, {"type": "run"}, None),
-        ({"activity_type": "Run"}, {"type": "run"}, None),
-        ({"activity_type": "sleep"}, {}, ValueError),
```
jsamoocha, line 492: This is good. Can you also remove similar cases from `test_update_activity()`?

jsamoocha (commented):
Thanks, this looks good! Please consider whether we need this `create` parameter that you introduced. It would be nice if we could do without the branching complexity it adds.

As an (optional) todo, you could also introduce fixtures to the client tests that were touched by this PR.

lwasser (commented):
ok @jsamoocha making progress. i need more input on validation for the create activity method. essentially i can't test for both sport type and activity type if i don't know it's coming from the create_activity method. This is because update activity doesn't require those keys in the call. (from what i understand). many thanks!

lwasser:
also tests shouldn't pass and i know that right now. once i have a bit more feedback it will be easy to get everything running as expected.

src/stravalib/client.py
jsamoocha: Ah, now I see. I was indeed missing something 🙈. Both types are optional when updating, but at least one needs to be provided when creating. Your option 2 looks the best, as this check is specific to the create method.

lwasser: ok wonderful. i wasn't certain what you were thinking but we're on the same page now!

fixed locally. thanks!

src/stravalib/tests/unit/test_validate_activity_type.py
lwasser: ok that is much cleaner. But i have to say - i'd never seen that dictionary merge / "union" operator before. i looked at it and saw a "or" value because of the `|` . that will take my brain some time to get used to 😆 but it's definitely clean written this way. i just pushed a few changes and will see if things all build. they are building locally just fine.

src/stravalib/client.py
lwasser: this has been addressed via a value error.

src/stravalib/tests/integration/test_client.py
```diff
@@ -295,6 +316,12 @@ def test_get_activity_streams_series_type_unofficial(mock_strava_api, client):
             None,
         ),
         ({"hide_from_home": False}, {"hide_from_home": "0"}, None, None),
+        (
+            {"name": "My awesome activity"},
```
lwasser, line 320: The tiny decline is cod coverage should be addressed by this additional test. 
🤞 
Screenshot 2024-03-18 at 7 42 49 PM

lwasser:
ok @jsamoocha i think i covered everything. that one change you suggested simplified the test significantly for validate activity. see what you think about the current state of things.  to make the create_activity test workflow work, i had to remove sport_type from default parameters fixture. i couldn't figure out a way to use pytest parametrize, and use the same fixtures in both tests so i opted to just add sport_type to the create activity tests. 

i also added a small test in update activity as there were 2 lines not covered by the current sets of tests params. this is ready for another review. ✨

src/stravalib/client.py
```diff
@@ -638,12 +638,78 @@ def get_friend_activities(self, limit: int | None = None) -> NoReturn:
...
+        """
+
+        # Check for sport_type first. If provided, it is favored over
+        # activity_type / type
+        if sport_type is not None:
+            if not sport_type in model.Activity.SPORT_TYPES:
+                raise ValueError(
+                    f"Invalid activity type: {sport_type}. Possible values: {model.Activity.SPORT_TYPES!r}"
+                )
+            params["sport_type"] = sport_type
```
jsamoocha, line 682: I removed the `pop(type)` here because it wasn't necessary anymore (since you changed the order of checking the types to `sport_type` first).

jsamoocha (approved):
Excellent, and thanks! I made two tiny changes. If you agree with these, feel free to merge.

lwasser:
excellent. thanks @jsamoocha merging this now!! ✨ i think this is my first real code / test contribution to stravalib. and i feel better about working with the test suite now because of it which is great. i appreciate all of the reviews.

PR #475, opened 2024-04-04
Fix: update actions to current versions

lwasser:
## Description

When we run all of our actions we get this error which relates to node js 16 actions being deprecated. This is a security issue but also it's good practice to keep actions current. This pr only updates actions to their most current versions! No code is going to be impacted by merging and everything should run as usual! 

Screenshot 2024-04-04 at 5 07 22 PM

## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This is a documentation update
- [x] This is a infrastructure update (docs, ci, etc)
- [ ] Other (please describe)

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [x] This change doesn't require tests

## Did you include your contribution to the change log?

- [ ] Yes, `changelog.md` is up-to-date.

i didn't update the change log for this one as it's a tiny update.

PR #477, opened 2024-04-06
FIX: corrects src path

jsamoocha:
This workflow config was not updated after adding the `src/` dir, causing the automatic pull request to fail.

- [x] This is an infrastructure update (docs, ci, etc)

## Does your PR include tests


jsamoocha:
Although the strava API change PR bot runs again, it seems to generate code that conflicts with some editing we did (see some of the typos in comments that originate from Strava.

Also, the generated container types are not the "plain" python `list,` but rather `typing.List`, which is odd since we set 3.9 as the python version for the generator. Will need to look into that later (that's in the strava_swagger2pydantic repo).

Finally, I keep seeing these node16 deprecation warnings, even after updating the action dependencies in the strava_swagger2pydantic repo and making a new release. Any ideas about that are welcome 🙏.

yihong0618:
> Although the strava API change PR bot runs again, it seems to generate code that conflicts with some editing we did (see some of the typos in comments that originate from Strava.
> 
> Also, the generated container types are not the "plain" python `list,` but rather `typing.List`, which is odd since we set 3.9 as the python version for the generator. Will need to look into that later (that's in the strava_swagger2pydantic repo).
> 
> Finally, I keep seeing these node16 deprecation warnings, even after updating the action dependencies in the strava_swagger2pydantic repo and making a new release. Any ideas about that are welcome 🙏.

node16 is easy to fix.
just update the checkout@version and python@version in Actions I will fix them in another pull request

yihong0618 (approved):
Nice catch

enadeau:
> Also, the generated container types are not the "plain" python `list,` but rather `typing.List`, which is odd since we set 3.9 as the python version for the generator. Will need to look into that later (that's in the strava_swagger2pydantic repo).

This is something I patched at the time by using ruff to update the syntax to the most modern one. See https://github.com/stravalib/strava_swagger2pydantic/pull/7/files

I don't have time to investigate that now but maybe this stopped working after some changes?

yihong0618:
> > Although the strava API change PR bot runs again, it seems to generate code that conflicts with some editing we did (see some of the typos in comments that originate from Strava.
> > Also, the generated container types are not the "plain" python `list,` but rather `typing.List`, which is odd since we set 3.9 as the python version for the generator. Will need to look into that later (that's in the strava_swagger2pydantic repo).
> > Finally, I keep seeing these node16 deprecation warnings, even after updating the action dependencies in the strava_swagger2pydantic repo and making a new release. Any ideas about that are welcome 🙏.
> 
> node16 is easy to fix. just update the checkout@version and python@version in Actions I will fix them in another pull request

no need another pr change to `peter-evans/create-pull-request@v6` seems ok

jsamoocha:
> This is something I patched at the time by using ruff to update the syntax to the most modern one. See https://github.com/stravalib/strava_swagger2pydantic/pull/7/files

Thanks for that hint; I now see how ruff is used to post-process the generated file. Will dig into that later...

jsamoocha:
I think all the issues mentioned are addressed by now, so please give this another review (as well as the new tiny PR in the strava_swagger2pydantic repo). After merging these two, we can try to rerun the bot. I'll remove the existing PR generated by the bot.

lwasser (approved):
ahh shoot i didn't even think about this workflow. great fix @jsamoocha  this means we haven't been getting updates for a while?! 🙈 let's merge!

lwasser:
let's merge this! @yihong0618 nailed it. those node errors are what i attempted to fix in #475 but it seems like i missed a critical workflow! they are popping up in everyone's ci workflows that aren't up to date now!

PR #482, opened 2024-04-16
Remove: deserialize method in favor of pydantic.parse_obj()

### LINKED ISSUE #358: Remove legacy model deserialize() calls from client
We need some cleanup that removes deprecated `deserialize()` calls on model classes and replaces these with pydantic's `parse_obj()`.

lwasser:
closes #358

## Description

This removes the deserialize method and replaces calls to is with pydantic.parse_obj. It also cleans up the tests accordingly.

## Type of change

- [x] Other (please describe) this is a deprecation / removal of deprecated features.

## Does your PR include tests

- [x] Yes no NEW tests but i cleaned up the tests accordingly

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 - Fix: Fixes the Strava API update bot (@jsamoocha, #477)
 - Fix: Cleanup dependencies to only use pyproject toml (@lwasser, #466)
 - Add: Strava API change - Route objects have a new waypoints attribute (@bot, #480)
+- Remove: Deprecated deserialize method from deprecated mixin (@lwasser, #358)
 
 ## v1.6
 
--- a/changelog.md
+++ b/changelog.md
@@ -11,7 +11,7 @@
 - Fix: Fixes the Strava API update bot (@jsamoocha, #477)
 - Fix: Cleanup dependencies to only use pyproject toml (@lwasser, #466)
 - Add: Strava API change - Route objects have a new waypoints attribute (@bot, #480)
-- Remove: Deprecated deserialize method from deprecated mixin (@lwasser, #358)
+- Fix: use parse_obj rather than deserialize internally where possible (@lwasser, #358)
 
 ## v1.6
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ tests = [
   "responses"
 ]
 docs = [
-  "sphinx",
+  "sphinx==7.3.5",
   "pydata-sphinx-theme",
   "autodoc_pydantic",
   "sphinx_remove_toctrees",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,9 +125,9 @@ warn_untyped_fields = true
 skip = './docs/_build, ./build, ./src/stravalib/tests/resources'
 
 [tool.ruff]
-select = [
+lint.select = [
   "F", # pyflakes
   "I", # isort
 ]
-ignore = ["F841"]
+lint.ignore = ["F841"]
 line-length = 79
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -1871,7 +1871,7 @@ class Client:
             The JSON response expected by Strava to the challenge request.
 
         """
-        callback = model.SubscriptionCallback.deserialize(raw)
+        callback = model.SubscriptionCallback.parse_obj(raw)
         callback.validate_token(verify_token)
 
         assert callback.hub_challenge is not None
@@ -1962,8 +1962,7 @@ T = TypeVar("T", bound=BaseModel)
 
 
 class ResultFetcher(Protocol):
-    def __call__(self, *, page: int, per_page: int) -> Iterable[Any]:
-        ...
+    def __call__(self, *, page: int, per_page: int) -> Iterable[Any]: ...
 
 
 class BatchedResultsIterator(Generic[T]):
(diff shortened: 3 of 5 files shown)


src/stravalib/model.py
```diff
@@ -197,38 +196,6 @@ class DeprecatedSerializableMixin(BaseModel):
     Inherits from the `pydantic.BaseModel` class.
     """
 
-    @classmethod
```
lwasser, line 200: I removed this completely. it's really just returning parse_obj anyway.

src/stravalib/tests/unit/test_model.py
```diff
@@ -29,10 +29,6 @@
 
 @pytest.mark.parametrize("model_class,attr,value", ((Club, "name", "foo"),))
 class TestLegacyModelSerialization:
-    def test_legacy_deserialize(self, model_class, attr, value):
```
lwasser, line 32: This test is now no longer needed

src/stravalib/client.py
```diff
@@ -2043,7 +2042,7 @@ def _fill_buffer(self) -> None:
             except AttributeError:
                 # Entity doesn't have a parse_obj() method, so must be of a
                 # legacy type
-                new_entity = self.entity.deserialize(  # type: ignore[attr-defined]
+                new_entity = self.entity.parse_obj(  # type: ignore[attr-defined]
```
lwasser, line 2045: parse_obj doesn't take a bind_client kwarg, only a dictionary. i'm actually not sure how this worked previously as is.  @stravalib/stravalib-maintainers can anyone provide me with some direction here? i generally understand the issue but am not sure how to fix it. tests run, mypy fails. so likely we need a new / improved test as well for this case once i understand how the legacy behavior should work? 

thank you!

lwasser: This seems to run if i keep deserialize as is for the legacy behavior. But what i still don't understand is how to recreate legacy behavior to test it.

src/stravalib/tests/unit/test_model.py
lwasser: i'm now not sure if this test should remain if we are indeed supporting this legacy behavior. i initially read the issue as we are not but perhaps we are.

src/stravalib/tests/unit/test_model.py
```diff
@@ -347,7 +343,7 @@ def test_subscription_update_deser(self):
             "aspect_type": "create",
             "event_time": 1297286541,
         }
-        subupd = model.SubscriptionUpdate.deserialize(d)
+        subupd = model.SubscriptionUpdate.parse_obj(d)
```
lwasser, line 346: based on my question about - i'll fix or remove this change accordingly.

jsamoocha: This type of call to `deserialize()` should still be supported in v1 because users may be doing this. For v2, we should indeed replace it by `parse_obj()` (or its Pydantic2 equivalent).

src/stravalib/model.py
lwasser: ok that clarification is helpful. I wondered about that because it would become a breaking change if i fully remove from the API

src/stravalib/client.py
lwasser: it runs fine with that consolidation! the new_entity object is an activity object with the bound_client assigned. As far as i can see it works as expected and tests continue to run. 

Screenshot 2024-04-17 at 5 50 26 PM

lwasser: this is what i was trying to figure out. i didn't understand what would trigger that except part of the code so i wasn't sure how to test for it.

src/stravalib/model.py
```diff
@@ -197,38 +196,6 @@ class DeprecatedSerializableMixin(BaseModel):
...
-    def deserialize(cls, attribute_value_mapping: dict[str, Any]) -> Self:
```
lwasser, line 201: i am adding this back now into the code via a upcoming commit

lwasser:
The new version of sphinx (just released TODAY a few hours ago) is breaking autodoc pydantic. So i pinned it to the version released yesterday. it is building locally now just fine. ahhh open source.

changelog.md
```diff
@@ -11,6 +11,7 @@
 - Fix: Fixes the Strava API update bot (@jsamoocha, #477)
 - Fix: Cleanup dependencies to only use pyproject toml (@lwasser, #466)
 - Add: Strava API change - Route objects have a new waypoints attribute (@bot, #480)
+- Remove: Deprecated deserialize method from deprecated mixin (@lwasser, #358)
```
lwasser, line 14:
```suggestion
- Fix: use parse_obj rather than deserialize internally where possible (@lwasser, #358)
```

pyproject.toml
```diff
@@ -66,7 +66,7 @@ tests = [
   "responses"
 ]
 docs = [
-  "sphinx",
+  "sphinx==7.3.5",
```
lwasser, line 69: 7.3.6 was released today (a few hours ago) and it breaks autodoc_pydantic.

pyproject.toml
```diff
@@ -125,9 +125,9 @@ warn_untyped_fields = true
 skip = './docs/_build, ./build, ./src/stravalib/tests/resources'
 
 [tool.ruff]
-select = [
+lint.select = [
```
lwasser, line 128: Ruff raises warnings locally. this fixes those and adheres to the newer release.

src/stravalib/tests/unit/test_model.py
lwasser: ok great. i've reverted this change!

lwasser:
ok. ready for another review when there is time!

jsamoocha (approved):
Thanks, this looks good!

lwasser:
thanks for the review and the direction @jsamoocha !!

PR #484, opened 2024-04-27
Fix docstrings in model.py 

### LINKED ISSUE #326: DOC: update / fix docstrings throughout to be numpy style

lwasser:
closes #326 

## Description

This is just a bunch of small updates to docstrings - i am finally following up on 326.  

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

I don't think this requires a changelog entry. it's more housekeeping

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -292,6 +292,10 @@ class BackwardCompatibilityMixin:
     return values based on what is defined in the following class attributes:
 
     * _field_conversions
+
+    Notes
+    ------
+    The class attributes below are not yet implemented:
     * _deprecated_fields (TODO)
     * _unsupported_fields (TODO)
     """
@@ -588,6 +592,8 @@ class Athlete(
     Notes
     ------
     Also provides access to detailed athlete stats upon request.
+    Many attributes in this object are undocumented by Strava and could be
+    modified at any time.
     """
 
     # Field overrides from superclass for type extensions:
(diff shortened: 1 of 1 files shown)


src/stravalib/model.py
```diff
@@ -287,6 +292,10 @@ class BackwardCompatibilityMixin:
     return values based on what is defined in the following class attributes:
 
     * _field_conversions
+
+    Notes
+    ------
+    The class attributes below are not yet implemented:
```
lwasser, line 298: @jsamoocha i'm doing some cleanup. i noticed this TODO in the notes. is it related to https://github.com/stravalib/stravalib/issues/323 or can you help me understand what the TODO actually is here?

thank you.

src/stravalib/model.py
```diff
@@ -377,9 +414,6 @@ class LatLon(LatLng, BackwardCompatibilityMixin, DeprecatedSerializableMixin):
     Enables backward compatibility for legacy namedtuple
```
lwasser, line 414: I believe this comment might have been from me when i was typing things. so i'm removing it.

src/stravalib/model.py
jsamoocha: These TODOs are nice to have, i.e., we can (potentially) use this mixin to warn users when they try to access deprecated or unsupported (by Strava) fields. However, I'm not sure we'll still use this mixin when moving to Pydantic2, and I am also unsure if it will be easy and straightforward to implement that logic. Feel free to remove these TODOs.

jsamoocha (approved):
This looks good, thanks!

src/stravalib/model.py
lwasser: ok thank you!

PR #486, opened 2024-04-27
Fix: update docs around python support

### LINKED ISSUE #254: Python version support

lwasser:
closes #254 

This is a TINY pr that will finish off #254 where we discussed when we drop versions of Python that we support. 
 
## Description

Does anyone have opposition to us dropping Python 3.9 at this point? 
The specific guidelines relate [to this spec](https://scientific-python.org/specs/spec-0000/)
which suggests supporting versions of Python for up to 3 years post release. 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

i didn't update the changelog for this either as it's minor. but if we drop 3.9 then i will refer to this documentation change in that pr via a changelog update.

### DIFF
--- a/docs/contributing/development-guide.md
+++ b/docs/contributing/development-guide.md
@@ -99,6 +99,13 @@ to Strava and handling the received responses (including rate limiting).
 It is used by methods in `client.py` to de-serialize raw response data into
 the domain entities in `model.py`.
 
+## Python support
+We loosely follow the [Numpy guidelines defined in NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)
+for Python version support. However, in some cases we may decide to
+support older versions of Python given visible community demand for
+such support.
+
+
 ## Code style, linting & typing
 
 We use the following tools to ensure consistent code format that follows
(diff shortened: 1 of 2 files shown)


enadeau (approved):

lwasser:
thank you @enadeau !!

PR #488, opened 2024-04-27
Add/Remove: Add support for python 3.12 and drop 3.9

### LINKED ISSUE #487: ENH: Drop Python 3.9 support + Add python 3.12 support
Following #301 and #486  
Python 3.8 was released 4.5 years ago adn security support for it [ends in october 2024](https://endoflife.date/python). 
Following the Numpy standards they recommend dropping support for python versions after ~3 years. I think it is justified if others are on board to drop support for 3.9 at this time and to add support for python 3.12 which many people are migrating to now and it is the currently active supported version of Python

Please let me know what ya'll think.

lwasser:
closes #487 

## Description

This drops Python 3.9 from all builds and adds 3.12. 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/.github/workflows/push-pypi.yml
+++ b/.github/workflows/push-pypi.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.9"
+          python-version: "3.10"
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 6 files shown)


lwasser:
i will include this in our next release IF all tests and things pass here.

.github/workflows/push-pypi.yml
```diff
@@ -41,17 +41,6 @@ jobs:
       - name: Check the archives
```
lwasser, line 41: I am also dropping the push to testPYPI here. it would be better for us to implement pre-releases then use test pypi. pydantic is not maintained on testpypi so nothing would ever actually work as we want it to if we test from test pypi. also we could open users to vulnerabilities because if a dependency of ours is uploaded to testpypi by a "bad actor" we might inadvertently install that bad package in our dependency suite locally .

lwasser:
because this is a small change i'll go ahead and merge and then rebase against my open release pr.

PR #491, opened 2024-04-28
Fix: add tests for datetime function

lwasser:
There is no open issue for this

## Description

This pr adds tests for the naive datetime function that is now in the model.py file. 
NOTE: We might want to migrate this to our utils in the refactor

I am adding this test because pydantic 2.x drops parse_datetime support. As such we will want to move to a different method for handling date times (possibly dateutils). adding this test now will make that migration easier when we go to implement it.

## Type of change

- [x] This is a infrastructure update (docs, ci, etc) -- tests
- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## Unreleased
+- Add: test for naive datetime (Pydantic method deprecated) (@lwasser)
+
 
 ## v1.7
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 # Change Log
 
 ## Unreleased
-- Add: test for naive datetime (Pydantic method deprecated) (@lwasser)
+- Add: test for naive datetime (Pydantic method deprecated) (#491, @lwasser)
 
 
 ## v1.7
--- a/src/stravalib/tests/unit/test_model.py
+++ b/src/stravalib/tests/unit/test_model.py
@@ -357,12 +357,10 @@ class ModelTest(TestBase):
         )
 
 
-from datetime import datetime
-
 # Creating a timezone-aware datetime object
 dt = datetime(2022, 4, 28, 12, 0, tzinfo=timezone.utc)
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## Unreleased
+- Add: test for naive datetime (Pydantic method deprecated) (@lwasser)
```
lwasser, line 4:
```suggestion
- Add: test for naive datetime (Pydantic method deprecated) (#491, @lwasser)
```

src/stravalib/tests/unit/test_model.py
```diff
@@ -353,3 +355,50 @@ def test_subscription_update_deser(self):
...
         )
+
+
+from datetime import datetime
+
+# Creating a timezone-aware datetime object
+dt = datetime(2022, 4, 28, 12, 0, tzinfo=timezone.utc)
+
+# Converting datetime to POSIX time
+posix_time = int(dt.timestamp())
```
lwasser, line 366: This is a bit hacky but allows us to create t imestamp that is timezone agnostic for us to run locally.

lwasser:
NOTE: i just tested this locally. Dateutil actually works well in place of pydantic.parse_date. (I might be missing particular formats that are not in the test suite here however!). 

If we are happy with the tests i've created here, it accepts a few more formats compared to pydantic. we could potentially swap this out sooner to support a future pydantic migration without causing breaking changes (i think?) to our current API. if you all are open to it i will open an issue about this. 

```
    if value is not None:
        print("using dateutil now")
        # Use dateutil.parser.parse for parsing date strings
        dt = parser.parse(str(value))
        # Remove timezone information, if any
        return dt.replace(tzinfo=None)
    else:
        return None
```

lwasser:
closing this in favor of the already merged #522  🚀

PR #492, opened 2024-05-03
pydantic 2.x migration p1

lwasser:
## Description

This PR is an attempt to begin to migrate from pydantic 1.x to 2.x. I'd love some feedback. i did NOT remove all of the legacy behavior in this PR. 

I also am still struggling with the activityzones test because the example data has floating point values in it whereas the API / strava_model file declares type: int .  I made some changes to the strava_model file that will need to be made via the bot IF this PR is merged and we move forward in this direction.

we can still do a lot of work to remove legacy behavior but this was a first stab at just upgrading to the new pydantic schema. 

## Type of change

- [z] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

i haven't worked on the changelog yet. i'm curious how many bugs i created with this PR first.
i did test this on some local builds and it ran but that doesnt' mean there aren't bugs or unintentional breakages of things! Please test and let me know how this works / what breaks / what looks off, etc.


pyproject.toml
```diff
@@ -49,7 +49,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
 ]
 
-dependencies = ["pint", "pytz", "arrow", "requests", "pydantic<2.0"]
+dependencies = ["pint", "pytz", "arrow", "requests", "pydantic"]
```
lwasser, line 52: Technically we should specify a lower bound now for pydantic >= 2.0

src/stravalib/client.py
```diff
@@ -988,9 +991,9 @@ def upload_activity(
     def get_activity_zones(
         self, activity_id: int
     ) -> list[model.BaseActivityZone]:
-        """Gets zones for activity.
+        """Gets activity zones for activity.
```
lwasser, line 994: This is where i'm struggling with tests as the test for this has data with floating points within it.

src/stravalib/client.py
```diff
@@ -1004,10 +1007,24 @@ def get_activity_zones(
         :class:`list`
             A list of :class:`stravalib.model.BaseActivityZone` objects.
 
+        Notes
+        -----
+        Activity zones require a Strava premium account.
+
         """
         zones = self.protocol.get("/activities/{id}/zones", id=activity_id)
+        # For troubleshooting
```
lwasser, line 1016: i would remove this - i just created this look to allow for easier line by line debugging. here the data for the first part of the test (all ints) populates into the model fine. the data for the second set of zones fails because of ints.

src/stravalib/client.py
```diff
@@ -2035,7 +2054,15 @@ def _fill_buffer(self) -> None:
...
         for raw in raw_results:
-            new_entity = self.entity.parse_obj(
+            # Parse object is deprecated - use model_validate instead
+            # Getting a validation error bceause
+            # 4 validation errors for Activity
+            # end_latlng
+            # Input should be a valid dictionary or instance of LatLon
+            # [type=model_type, input_value=[39.96658810414374, -105.2557013835758], input_type=list]
+            # Here we are trying to populate the Activity data object with
+            # raw returned data. but it wants the input to be a dict or LatLon
```
lwasser, line 2064:
```suggestion
```

src/stravalib/field_conversions.py
```diff
@@ -37,8 +37,24 @@ def fn_wrapper(field_value: T | None) -> U | None:
 
 @optional_input
 def enum_value(v: ActivityType | SportType) -> str:
+    """A method that collects the root value of a Activity or
```
lwasser, line 40: i wondered if we could remove this method because it only returns a single value from a RootModel class which can be easily accessed with .root

src/stravalib/field_conversions.py
```diff
@@ -48,9 +64,26 @@ def enum_value(v: ActivityType | SportType) -> str:
 
 @optional_input
 def enum_values(enums: Sequence[ActivityType | SportType]) -> list[str | None]:
+    """A function that processes multiple enums and returns a list.
```
lwasser, line 67: This could be kept (or removed) but i haven't had time to dig into whether pydantic 2.x handles enum lists now.

src/stravalib/strava_model.py
```diff
@@ -40,8 +40,8 @@ class ActivityTotal(BaseModel):
     """
 
 
-class ActivityType(BaseModel):
-    __root__: Literal[
+class ActivityType(RootModel):
```
lwasser, line 43: All root models now inherit from the RootModel class  and use root: rather than __root__.

src/stravalib/strava_model.py
```diff
@@ -669,6 +673,42 @@ class SummaryClub(MetaClub):
     Whether the club is verified or not.
     """
 
+    @field_validator("activity_types", mode="before")
```
lwasser, line 676: i wasn't sure about this. Here i added a field validator that allows us to loop through the Literal values and determine whether (sport / activity) type are valid values.

src/stravalib/tests/integration/test_client.py
```diff
@@ -15,6 +15,14 @@
 from stravalib.unithelper import UnitConverter, meters, miles
 
 
+@pytest.fixture
```
lwasser, line 18: made this a fixture to clean up the test a bit. however i'm still unsure about the values in example_zone_response (why we have floats and why the test passed before). i'm certain i'm missing something!

src/stravalib/tests/integration/test_client.py
```diff
@@ -157,18 +165,38 @@ def test_get_club_activities(mock_strava_api, client):
...
-    with open(
-        os.path.join(RESOURCES_DIR, "example_zone_response.json"), "r"
-    ) as zone_response_fp:
-        zone_response = json.load(zone_response_fp)
+def test_get_activity_zones(mock_strava_api, client, zone_response):
+    """Returns an activities associated zone (related to heart rate and power)
+
+    Notes
+    -----
+    There is no example response for this endpoint in swagger.json so
```
lwasser, line 179: NOTE: i didn't check the json file to see if there is an example response now.

src/stravalib/tests/integration/test_client.py
```diff
@@ -157,18 +165,38 @@ def test_get_club_activities(mock_strava_api, client):
...
+    we created a sample json file with the return that we are currently
+    seeing. Thus this method could break at any time if Strava changes
+    the response output.
+
+    """
+    #
     mock_strava_api.get("/activities/{id}/zones", json=zone_response)
+    """
+    # TODO: failing here E  pydantic_core._pydantic_core.ValidationError: 10 validation errors for BaseActivityZone
```
lwasser, line 188: This is the current failure point in the tests that i could use some guidance on.

src/stravalib/tests/unit/test_model.py
```diff
@@ -35,12 +36,6 @@ def test_legacy_deserialize(self, model_class, attr, value):
             model_obj = model_class.deserialize({attr: value})
             assert getattr(model_obj, attr) == value
 
-    def test_legacy_from_dict(self, model_class, attr, value):
```
lwasser, line 38: we could potentially remove all of these legacy tests and methods (it's only 3 i think in that mixin). we could also do that in a separate PR to allow for smaller pr's (this one is already big!).

src/stravalib/tests/unit/test_model.py
```diff
@@ -260,7 +272,7 @@ def test_entity_collections(self):
                 },
             ]
         }
-        a.from_dict(d)
+        a = model.Athlete.model_validate(d)
```
lwasser, line 275: removes legacy behavior and uses pydantic 2.x which prefers immutability and returns a new object with the data ingested.

src/stravalib/client.py
jsamoocha: Supposedly, Pydantic 2 applies stricter validation. It looks like I created that test data from a real Strava API response. It also looks like the API spec is wrong (on the Strava side): the `min` and `max` attributes of `ZoneRange` should be of type `Optional[int | float]` (or `int | float | None`).

We can solve this Strava bug by overriding the type on our model class (just as we did for `BaseActivityZone.type`, also because of a misspecified Strava API).

src/stravalib/field_conversions.py
jsamoocha: As we are introducing breaking changes, let's go ahead and return the enum instead of its `str` representation. It improves the accuracy of typing and can potentially prevent bugs for users. Let's document all these breaking changes somewhere.

src/stravalib/field_conversions.py
jsamoocha: Same as above, let's not try to force backward compatibility by converting the list of enums to a list of str.

src/stravalib/model.py
```diff
@@ -172,29 +200,6 @@ def check_valid_location(
...
-        A datetime object representing the datetime input value.
-    """
-    if value:
-        dt = parse_datetime(value)
-        return dt.replace(tzinfo=None)
-    else:
-        return None
-
-
 class DeprecatedSerializableMixin(BaseModel):
```
jsamoocha, line 203: This class can be removed entirely

src/stravalib/strava_model.py
jsamoocha: Did you regenerate this model using the `datamodel_code_generator` tool (as we do in the bot)? If not, that's the very first thing we should do.

src/stravalib/tests/unit/test_model.py
jsamoocha: It's ok to remove them now.

jsamoocha:
Thanks! I think the most important question/remark is: did you re-generate `strava_model.py` using the same tooling we use in the bot? If not, re-generating that file and building the rest upon it should be our first step.

lwasser:
> Thanks! I think the most important question/remark is: did you re-generate `strava_model.py` using the same tooling we use in the bot? If not, re-generating that file and building the rest upon it should be our first step.

thanks @jsamoocha for the feedback! i'll go in and address all of your comments but wanted to address this first. i did NOT fix the model generator so i appreciate #495 ! In this pr i wanted to see what the changes would look like first! This helped me also get a much better handle on all of the inheritance patterns. 

I will fix things here. we can merge your other pr and i'll rebase aginst it and remove my changes to strava_model which were just for testing purposes.

 Finally i'm also really curious about that dateaware pydantic object and how it will impact things . i didn't look at that at all.

lwasser:
Updates: Most of the above has been addressed but i'm next going to fix a few tests that broke with these updates. And then finally test against #495 which is the better version of the model file i manually updated here. 

i'm keeping notes of [everything changed here in this hackmd file](https://hackmd.io/1Iu1MzhSR-Ovqf8QQX__Hg)

everyone should be able to access / update the file as they wish. I will turn this document into a documentation page that we can add to as we add / remove enhance things once i get all of the tests running again here. 🚀

lwasser (commented):
Ok i'm leaving my response to review feedback now. there are two items that i need to fix now

1. mypy is throwing some warnings because of the switch to passing RootModel objects rather than strings for sport /activity type. i just haven't had time to look closely but right now it's expecting a dict but actually getting something else (possible a string still). 
2. black docs issue that is likely super minor. everything else here however is ready for another review. 

i hope to carve out time to look at the mypy issue in the next day or two.

lwasser:
this is ready for review. the typing issues were smaller than i thought. i'm not 100% confident however in the `RelaxedActivityType` validator if folks could look more closely at that.

.github/workflows/type-check.yml
```diff
@@ -20,10 +20,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
```
lwasser, line 22: i like to run checks on the "middle" python version. for typing however i'm not sure if this is best practice or not!

src/stravalib/client.py
jsamoocha: I don't think there's any problem with class attributes named `max`. Since they can only be accessed as attributes of an object, there's no clash with the builtin `max()`.

src/stravalib/client.py
```diff
@@ -326,8 +326,11 @@ def get_athlete(self) -> model.Athlete:
             The athlete model object.
 
         """
+        # The api call is slow now
```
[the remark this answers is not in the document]
jsamoocha: Let's check that later using curl or Postman (I made a section for these in your hackmd doc).

jsamoocha, line 329:
```suggestion
```

src/stravalib/client.py
```diff
@@ -326,8 +326,11 @@ def get_athlete(self) -> model.Athlete:
...
         raw = self.protocol.get("/athlete")
-        return model.Athlete.parse_obj({**raw, **{"bound_client": self}})
+        # The `parse_obj` method is deprecated; use `model_validate` instead.
```
jsamoocha, line 331:
```suggestion
```

src/stravalib/client.py
```diff
@@ -1004,10 +1007,26 @@ def get_activity_zones(
...
 
+        Notes
+        -----
+        Activity zones require a Strava premium account.
+        The Strava API has the return value for activity zones incorrectly
+        typed as `int` when it can return either `int` or `float`.
+
         """
         zones = self.protocol.get("/activities/{id}/zones", id=activity_id)
+        # For troubleshooting
```
jsamoocha, line 1018:
```suggestion
```

src/stravalib/client.py
```diff
@@ -1004,10 +1007,26 @@ def get_activity_zones(
...
+        for z in zones:
+            print(z)
+            a = model.BaseActivityZone.model_validate(
+                {**z, **{"bound_client": self}}
+            )
+            print(a)
```
jsamoocha, line 1024:
```suggestion
```

src/stravalib/model.py
```diff
@@ -357,9 +269,10 @@ class BoundClientEntity(BaseModel):
     bound_client: Optional[Any] = Field(None, exclude=True)
 
 
+# SHould this be typed differently?
```
jsamoocha, line 272: What happens when we remove `Any`? (If that's what you meant?)

src/stravalib/model.py
```diff
@@ -374,18 +287,21 @@ def check_activity_type(cls, values: dict[str, Any]) -> dict[str, Any]:
...
             LOGGER.warning(
-                f'Unexpected activity type. Given={v}, replacing by "Workout"'
+                f'Unexpected activity type. Given={values}, replacing by "Workout"'
             )
-            values["__root__"] = "Workout"
+            values = "Workout"
         return values
 
 
+# TODO: should this type just be str (why would we accept Any here?)
```
jsamoocha, line 301: Same as above, what happens when `Any` is removed?

src/stravalib/model.py
```diff
@@ -400,21 +316,24 @@ def check_sport_type(cls, values: dict[str, Any]) -> dict[str, Any]:
         dict
             A dictionary with a validated sport type value assigned.
         """
-        v = values["__root__"]
-        if v not in get_args(SportType.__fields__["__root__"].type_):
+        # Do we want to deprecated SportType? os because it's still returned
```
jsamoocha, line 319: I don't understand this. Why should `SportType` be deprecated? According to Strava, this is the type that should be used.

src/stravalib/model.py
```diff
@@ -471,20 +394,24 @@ class Club(
...
     """
 
     # Undocumented attributes:
     profile: Optional[str] = None
     description: Optional[str] = None
     club_type: Optional[str] = None
 
-    _field_conversions = {"activity_types": enum_values}
+    # Removing this so activity_types isn't converted to strings
+    # _field_conversions = {"activity_types": enum_values}
```
jsamoocha, line 414:
```suggestion
```

src/stravalib/model.py
```diff
@@ -1045,11 +951,12 @@ class Segment(
         "elevation_high": uh.meters,
         "elevation_low": uh.meters,
         "total_elevation_gain": uh.meters,
-        "activity_type": enum_value,
+        # enum_value returns a string. but we can accept a ActivityType object
+        # "activity_type": enum_value,
```
jsamoocha, line 955:
```suggestion
```

src/stravalib/model.py
```diff
@@ -1186,18 +1089,16 @@ class Activity(
         "total_elevation_gain": uh.meters,
         "average_speed": uh.meters_per_second,
         "max_speed": uh.meters_per_second,
-        "type": enum_value,
-        "sport_type": enum_value,
+        # "type": enum_value,
+        # "sport_type": enum_value,
```
jsamoocha, line 1093:
```suggestion
```

src/stravalib/tests/unit/test_field_conversions.py
```diff
@@ -2,12 +2,11 @@
...
 
 from stravalib.field_conversions import (
-    enum_value,
-    enum_values,
     optional_input,
     timezone,
 )
-from stravalib.strava_model import ActivityType, SportType
+
+# from stravalib.strava_model import ActivityType, SportType
```
jsamoocha, line 9:
```suggestion
```

src/stravalib/tests/unit/test_field_conversions.py
```diff
@@ -19,14 +18,27 @@ def foo(x):
...
+#     a = ActivityType("Run")
+#     assert enum_value(a) == "Run"
+
+
+# def test_enum_values():
+#     """Club objects may have one or more ActivityTypes associated with them.
 
+#     This tests that when provided with a list of types, enum values
+#     parses multiple types and returns a list.
+#     """
```
jsamoocha, line 37:
```suggestion
```

src/stravalib/tests/unit/test_field_conversions.py
```diff
@@ -19,14 +18,27 @@ def foo(x):
...
 
-def test_enum_values():
-    assert enum_values(
-        [ActivityType(__root__="Run"), SportType(__root__="Ride")]
-    ) == ["Run", "Ride"]
+#     a = ActivityType("Run")
+#     b = SportType("Ride")
+#     assert enum_values([a, b]) == ["Run", "Ride"]
```
jsamoocha, line 41:
```suggestion
```

jsamoocha (approved):
Thanks! This looks good as a first step. Let's merge this and continue with the removal of `BackwardCompatibilityMixin`. I added some comments/todo's in the hackmd file.

src/stravalib/client.py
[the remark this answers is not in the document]
lwasser: ok perfect! resolving this.

src/stravalib/client.py
lwasser: perfect. resolving

src/stravalib/model.py
lwasser: everything runs fine without Any. so i've removed it. right now i'm passing a string from the dict to the validator. i'll be interested to see when you work on things if there is a better way. but for now i've fixed the typing and my confusing comments (that i think you understood!) as well

src/stravalib/model.py
lwasser: we definitely don't need Any the way things are written now. i wasn't sure if it was there for some other purpose! i've removed it.

src/stravalib/model.py
lwasser: apologies - this was my mistake. You are correct. i mixed up sport and activity type.  i'll remove this comment

src/stravalib/tests/integration/test_client.py
```diff
@@ -157,18 +165,38 @@ def test_get_club_activities(mock_strava_api, client):
...
+    E   distribution_buckets.0.max
+    E     Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=3.319925953425268, input_type=float]
+    E       For further information visit https://errors.pydantic.dev/2.7/v/int_from_float
+    E   distribution_buckets.1.max
+    E     Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=3.8569727988322966, input_type=float]
+    E       For further information visit https://errors.pydantic.dev/2.7/v/int_from_float
+    """
+    # TODO: Why is the example JSON file contain floating points if
+    # strava docs suggest only ints will be returned? i believe this
+    # test is failing because of the floats.
```
lwasser, line 198:
```suggestion

```

src/stravalib/client.py
lwasser: oh gosh - i meant to remove that it was just my testing things.

lwasser:
@jsamoocha wonderful. thank you for your patience in this review. i went through and cleaned up more comments that i had left and code that was commented out. i'll make pre-commit happy and will merge now. then on the phase two! i'm happy to review and add docstrings / documentation etc whereever it is helpful.

PR #495, opened 2024-05-05
Modifies generated model by using Pydantic v2

jsamoocha:
## Description

I saw differences between the Pydantic v2 generated model and the version in #492. This PR replaces `strava_model.py` by a version we can expect to come out of the code generator. Based on migration issues, failing tests, or backward compatibility we like to maintain, we may need to further tweak the code generator tool.

## Type of change

## Does your PR include tests

Tests should be dealt with in #492 

Not applicable here.

### DIFF
--- a/src/stravalib/strava_model.py
+++ b/src/stravalib/strava_model.py
@@ -3,10 +3,9 @@
 
 from __future__ import annotations
 
-from datetime import datetime
 from typing import Literal, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import AwareDatetime, BaseModel, Field, RootModel
 
 
 class ActivityTotal(BaseModel):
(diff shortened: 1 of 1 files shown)


src/stravalib/strava_model.py
```diff
@@ -3,10 +3,9 @@
 
 from __future__ import annotations
 
-from datetime import datetime
 from typing import Literal, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import AwareDatetime, BaseModel, Field, RootModel
```
jsamoocha, line 8: I'm mostly curious about what this `AwareDatetime` will do.

lwasser:
@jsamoocha i'm going to merge this so i can rebase against my pr (tests are running locally!)

src/stravalib/strava_model.py
lwasser: @jsamoocha i know this is a very old pr now but when i run the model locally, i can't get it to actually use these new pydantic objects. Did you run the code locally and it added AwareDatetime? it's reverting your changes every time i run it locally. 

https://github.com/stravalib/strava_swagger2pydantic/pull/14/files

Issue #496, opened 2024-05-10
ENH: Improve linting via GitHub Actions


obscurerichard:
### Feature Type

- [X] Adding new functionality to stravalib

- [ ] Changing existing functionality in stravalib

- [ ] Removing existing functionality in stravalib

### Problem Description

Several of the linters mentioned in the documentation (`black`, `isort`, and `flake8`) currently do not have any CI checks enforced via GitHub actions. `mypy` type checking is enforced via GItHub Actions CI, which is awesome.

### Feature Description

Add a `lint` GitHub action that runs linters for `black`, `isort`, `flake8`, and possibly [FawltyDeps](https://github.com/tweag/FawltyDeps). Lay the groundwork for adding other linters.

### Alternative Solutions

Add any new linters via https://github.com/stravalib/stravalib/blob/main/.pre-commit-config.yaml as well.

### Additional Context

I have created an implementation of this in https://github.com/obscurerichard/stravalib/tree/gh-actions-lint and intend to raise a PR. I'm the primary maintainer of [Freezing Saddles](https://freezingsaddles.org) - see https://github.com/freezing-saddles - perhaps the first system to use Stravalib (thanks @hozn !)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lwasser:
hey @obscurerichard - thanks for the issue. We are using the pre-commit.ci bot to lint every pr. it actually will do more than just check - it will fix in place if we need it to. 
we also are using RUFF which covers isort, black and many other linters and is fast given it's written in rust. so these CI checks are in place. if you can clarify the issues that you are trying to address in your current PR that would be helpful.

obscurerichard:
Ah, that makes sense. So we are already covered via `ruff` and `pre-commit.ci` for the basics. I'll amend this issue and the PR to be more focused and omit changing things for linters that are already working well. However, it's interesting that the `isort` linting that the GitHub action 

I do see that there is some linting done via GitHub Actions, and there _is_ one new linter I'd love to apply to this, [FawltyDeps](https://github.com/tweag/FawltyDeps) - which can find undeclared or unused dependencies.  It _did_ find a couple undeclared transitive dependencies in StravaLib.

[...]

PR #497, opened 2024-05-10
Lint via GitHub Action, and add FawltyDeps checker

obscurerichard:
## Description
Implements #496.

Add linting via GitHub Actions for `black`, `isort`, `flake8`, and `fawltydeps`.

* Add CI-enforced linting with isort, black, and flake8
* Add fawltydeps linter
* fix issues isort and fawltydeps found

Add the [FawltyDeps](https://giithub.com/tweag/FawltyDeps) dependency checker - it will ensure undeclared and unused dependencies do not sneak into the project.

Note that the changes in these files are to fix issues that the `isort` plugin caught:
* `src/stravalib/model.py`
* `examples/strava-oauth/server.py`

## Type of change

- [X] This change requires a documentation update
- [X] This is a documentation update
- [X] This is a infrastructure update (docs, ci, etc)
- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes
- [X] This change doesn't require tests

- [X] Yes, `changelog.md` is up-to-date.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,44 @@
+name: Lint
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+# Use bash by default in all jobs
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  lint:
+    name: Test Run (${{ matrix.python-version }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.12"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: isort lint
+        uses: isort/isort-action@v1
+
+      - name: black lint
+        uses: psf/black@stable
+
+      - name: FawltyDeps lint
+        uses: tweag/FawltyDeps-action@v0.0.1
+        with:
+          options: --detailed
+
+      - name: flake8 lint
+        uses: py-actions/flake8@v2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,3 +68,9 @@ repos:
             )$
         additional_dependencies:
           - tomli
+
+  - repo: https://github.com/tweag/FawltyDeps
+    rev: v0.15.0
+    hooks:
+      - id: check-undeclared
+      - id: check-unused
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Added
+- Add: improved linting in CI, including `black`, `isort`, `flake8`, and `fawltydeps` (@obscurerichard, #496)
+
+### Contributors to this release
+
+@obscurerichard
+
 ## v1.7
 
 ### Added
--- a/docs/contributing/development-guide.md
+++ b/docs/contributing/development-guide.md
@@ -117,6 +117,7 @@ best practices for Python code readability and consistency:
   Because black's default line length is 88 characters, we adjust it to 79
   characters in the config [to follow PEP 8 line width guidelines](https://peps.python.org/pep-0008/#maximum-line-length).
 - [isort](https://pycqa.github.io/isort/): ensure imports are ordered following [PEP 8 import guidelines](https://peps.python.org/pep-0008/#imports)
+- [FawltyDeps](https://github.com/tweag/FawltyDeps/): FawltyDeps is a linter that ensures that finds _undeclared_ or _unused_ 3rd party dependencies in the project.
 - [flake8](https://flake8.pycqa.org/en/latest/): flake8 is a linter. It identifies other PEP 8 issues in the code that Black will not address including comments that extend beyond 79
   characters and doc string line width. It also will identify unused imports and
   unused but declared variables.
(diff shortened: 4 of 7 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,44 @@
+name: Lint
```
lwasser, line 1: we use pre-commit CI for our linting and it also allows us to fix pr's in place so this lint workflow is not needed

.pre-commit-config.yaml
```diff
@@ -68,3 +68,9 @@ repos:
             )$
         additional_dependencies:
           - tomli
+
+  - repo: https://github.com/tweag/FawltyDeps
```
lwasser, line 72: We don't need to add this hook to every single commit that a user makes. I think we can run this once if there are missing dependencies that haven't been declared that we don't know about!

changelog.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Added
+- Add: improved linting in CI, including `black`, `isort`, `flake8`, and `fawltydeps` (@obscurerichard, #496)
+
+### Contributors to this release
+
+@obscurerichard
```
lwasser, line 10:
```suggestion
```

we will add contributors at the end in our changelog right before a release is crafted.

docs/contributing/development-guide.md
```diff
@@ -117,6 +117,7 @@ best practices for Python code readability and consistency:
   Because black's default line length is 88 characters, we adjust it to 79
   characters in the config [to follow PEP 8 line width guidelines](https://peps.python.org/pep-0008/#maximum-line-length).
 - [isort](https://pycqa.github.io/isort/): ensure imports are ordered following [PEP 8 import guidelines](https://peps.python.org/pep-0008/#imports)
+- [FawltyDeps](https://github.com/tweag/FawltyDeps/): FawltyDeps is a linter that ensures that finds _undeclared_ or _unused_ 3rd party dependencies in the project.
 - [flake8](https://flake8.pycqa.org/en/latest/): flake8 is a linter. It identifies other PEP 8 issues in the code that Black will not address including comments that extend beyond 79
```
lwasser, line 121: this could be super useful to us - please remoe fawlty deps, and replace thie section with a link to ruff. and the linters that ruff runs (and what settings we have for ruff!) this doc update would be a very useful contribution to stravalib!

pyproject.toml
```diff
@@ -60,9 +60,11 @@ changelog = "https://github.com/stravalib/stravalib/blob/main/changelog.md"
 [project.optional-dependencies]
 build = ["build"]
 tests = [
+  "nox",
```
lwasser, line 63: Can you please clarify what you are trying to accomplish here. nox is the tool we use to run tests, build docs, etc. as such it should not be in the tests dev envt because when we run tests we use nox `nox -s docs`, `nox -s tests`. 

IF we need to add a section for general tools here we could but a lot of people also might install nox using `pipx` so it's a bit of a tricky decision. i'd lean towards making sure it's clear to have nox installed first over lots of nox installations across python environments

pyproject.toml
```diff
@@ -80,6 +82,7 @@ docs = [
 ]
 lint = [
   "pre-commit",
+  "fawltydeps",
```
lwasser, line 85:
```suggestion
```

pyproject.toml
```diff
@@ -91,6 +94,37 @@ line-length = 79
 # When editing the config for black in this file, be sure to make
 # the same edits in the repo stravalib/strava_swagger2pydantic
 
+[tool.fawltydeps]
+ignore_undeclared = [
+  "stravalib",
+  "_typeshed",
+]
```
lwasser, line 101:
```suggestion
```

pyproject.toml
```diff
@@ -60,9 +60,11 @@ changelog = "https://github.com/stravalib/stravalib/blob/main/changelog.md"
 [project.optional-dependencies]
 build = ["build"]
 tests = [
+  "nox",
   "pytest",
   "pytest-cov",
-  "responses"
+  "responses",
+  "typing-extensions"
```
lwasser, line 67: This one is interesting! We just dropped python 3.9 and we don't need support for experimental typing being prototyped in 3.13 as we don't support it yet. so i'm curious why this was added.

lwasser (changes requested):
Thank you for your effort here @obscurerichard i think the most valuable piece here for us is an update to our docs on using ruff if you want to submit that as a pr instead!

pyproject.toml
obscurerichard: It was one of the things that FawltyDeps found as being an undeclared dependency. Try running fawltydeps on the main branch and this pops out.

obscurerichard:
I think I'm just going to close this and open an alternate PR that adds some doc improvements, since you don't really want `fawltydeps` in the linting tools as a main line contribution.

pyproject.toml
obscurerichard: I'm going to circle back on this once I've refined the tooling a bit more and run some more tests.

`nox` popped out as one of the undeclared dependencies when I ran `fawltydeps` - so my choices to get it to run clean were: 

1. Ignore `nox` in the `fawltydeps` pragmas - the path you are suggesting
1. Add `nox` as an explicit `tests` dependency

pyproject.toml
lwasser: I believe you! i wonder if you should report a bug in the fawltydeps repo to better understand why it would suggest this tool for our build. it doesn't fully make sense to me unless we were trying to enforce backward typing compatibility  (for instance perhaps when we did support 3.9 and we wanted to use Unions `|`).  but also it's only happening / being added when running tests which is also odd to me. i might be missing something too but this seems like a bug. 

but i do agree that it's best to just close this pr. 
many thanks for your time here!

Issue #500, opened 2024-05-11
Stravalib 2.0 -> Introduce custom Pydantic types for field conversions


jsamoocha:
- [x] Introduce custom types that inherit from the generated model's base types and add accessor logic to get the (Pint) `Quantity` representation of these base types.
- [x] Introduce custom type for time zones
- [x] Introduce custom type for time delta

Depends on #498

lwasser:
this i think answers my question from the review about what the plan is.

jsamoocha:
Yes, the plan is to access the "extended" attributes as follows:

```python
d = some_object.distance  # returns float
q = some_object.distance.quantity  # returns pint.Quantity (in meters)
z = unithelper.feet(some_object.distance)  # returns pint.Quantity (in feet)
```

I found a way to achieve this and maintain a correct type hierarchy.

lwasser:
@jsamoocha this sounds great. 

i have a very basic question. The extender approach sounds like a smooth way to implement these conversions. I"m curious how this compares to returning the distance in quantities via a validator for each attr that forces the conversion when the model is populated. I have a few guesses as to why the extender is better but wanted to get your feedback! 

* perhaps involving typing and type overrides or 
* perhaps it would make the overall code base more complex having so many validators?
* Or finally perhaps it allows other users who may want to raw response values to convert as they please + the quantity conversions that are so nice to have in place.

jsamoocha:
Closed by #534

PR #503, opened 2024-05-11
Remove backward compatibility mixin

### LINKED ISSUE #498: Stravalib2.0 -> Remove BackwardCompatibilityMixin
- [x] Remove this mixin to prevent typing issues such as #437.
- [x] Remove all `_field_conversions` from model classes.

The "enhanced" fields will get the same (parent) type as the overridden fields from the generated strava model.
This will make several tests fail, which will be resolved in other issues.

jsamoocha:
closes #498

## Description

Removes `BackwardCompatibilityMixin` and its related `_field_conversions` model attributes. As a result, all the previously "modified" types that, e.g., returned a pint `Quantity` object instead of a unitless float now return the base type. Later PRs will re-add these customized types but will ensure they match the parent type.

## Type of change

## Does your PR include tests

- [x] Yes

But the new test is skipped until we implement the new custom types.

Let's add one changelog entry for when the complete `pydantic-v2` branch merges into `main`?

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -668,7 +668,7 @@ class ActivityPhoto(BaseModel):
         else:
             photo_type = "(no type)"
             idfield = "id"
-            idval = self.id
+            idval = self.uid
 
         return "<{clz} {type} {idfield}={id}>".format(
             clz=self.__class__.__name__,
@@ -792,7 +792,7 @@ class AthletePrEffort(
     _naive_local = field_validator("start_date_local")(naive_datetime)
 
     @property
-    def elapsed_time(self) -> Optional[timedelta]:
+    def elapsed_time(self) -> Optional[int]:
         """A property that supports backwards compatibility with the
         elapsed_time method used in previous versions of stravalib"""
 
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -31,10 +31,6 @@ from dateutil import parser
 from pydantic import BaseModel, Field, field_validator, model_validator
 
 from stravalib import exc, strava_model
-from stravalib import unithelper as uh
-from stravalib.field_conversions import (
-    timezone,
-)
 from stravalib.strava_model import (
     ActivityStats,
     ActivityTotal,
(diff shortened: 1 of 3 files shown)


src/stravalib/model.py
```diff
@@ -747,7 +668,7 @@ class name, photo type, and a key identifier."""
         else:
             photo_type = "(no type)"
             idfield = "id"
-            idval = self.id
+            idval = self.uid
```
jsamoocha, line 671: Mypy (rightfully) complained about `id` not existing in this class. Somehow, it passed in previous versions for unknown reasons.

src/stravalib/model.py
```diff
@@ -880,33 +777,22 @@ class AthleteSegmentStats(
...
-    _field_conversions = {
-        "distance": uh.meters,
-    }
-    pr_elapsed_time: Optional[timedelta] = None  # type: ignore[assignment]
-
     _naive_local = field_validator("start_date_local")(naive_datetime)
 
     @property
-    def elapsed_time(self) -> Optional[timedelta]:
+    def elapsed_time(self) -> Optional[int]:
```
jsamoocha, line 795: Since we don't use the overridden attribute anymore, this return type has changed.

src/stravalib/tests/integration/test_client.py
```diff
@@ -156,7 +155,7 @@ def test_get_activity_laps(mock_strava_api, client):
     )
     laps = list(client.get_activity_laps(42))
     assert len(laps) == 2
-    assert laps[0].distance == meters(1000)
+    assert laps[0].distance == 1000
```
jsamoocha, line 158: These returns will now be "plain" floats or ints (which will later get properties or other accessors to retrieve the value as a quantity).

src/stravalib/tests/unit/test_model.py
```diff
@@ -238,68 +207,6 @@ def test_entity_collections(self) -> None:
...
-        a.total_elevation_gain = 1000  # m
-        self.assertIsInstance(a.distance, Quantity)
-        self.assertIsInstance(a.total_elevation_gain, Quantity)
-        self.assertEqual(1.0, float(uh.kilometers(a.distance)))
-        self.assertEqual(1.0, float(uh.kilometers(a.total_elevation_gain)))
-
-    def test_weight_units(self):
-        """ """
-        # PowerActivityZone
-
```
jsamoocha, line 302: These tests are not valid anymore since the attributes such as `distance` will not be of type `Quantity` anymore.

src/stravalib/model.py
```diff
@@ -460,19 +392,14 @@ class Shoe(Gear):
...
     """An objecting containing a set of total values for an activity including
     elapsed time, moving time, distance and elevation gain."""
 
-    _field_conversions = {
-        "distance": uh.meters,
-        "elevation_gain": uh.meters,
-    }
-    elapsed_time: Optional[timedelta] = None  # type: ignore[assignment]
-    moving_time: Optional[timedelta] = None  # type: ignore[assignment]
+    pass
```
lwasser, line 399: @jsamoocha as it was, we were converting distance and elevation gain to meters and time to timedelta type. but now we are no longer converting. Simpler code is great. However do we want to specify the units being returned in the docstring here for users and add a note about how values are potentially different?

For people who are maintaining  applications, will this significantly impact them?

src/stravalib/model.py
```diff
@@ -460,19 +392,14 @@ class Shoe(Gear):
...
 
 
-class AthleteStats(ActivityStats, BackwardCompatibilityMixin):
+class AthleteStats(ActivityStats):
     """
```
lwasser, line 403: Similar question here - if we have changed units returned, perhaps a note would be useful.

src/stravalib/model.py
```diff
@@ -1248,7 +1099,7 @@ def validate_token(
         assert self.hub_verify_token == verify_token
 
 
-class SubscriptionUpdate(BackwardCompatibilityMixin, BoundClientEntity):
+class SubscriptionUpdate(BaseModel):
```
lwasser, line 1102: This is so much cleaner!

src/stravalib/tests/integration/test_client.py
lwasser: sounds good. what is the plan to add such properties / what will that look like in the model?

lwasser (approved):
@jsamoocha this all looks good to me. i have two general questions / comments

1. my main question is what is your vision to implement converting values to quantities? Is the plan to create inherited classes vs using that `_field_conversion={}`?
2. is it worth adding docstrings where units were changed? Or is the idea that in the end it will still return things in quantities - meters / timedelta, etc?

this is just me trying to understand the next steps. thank you for all of the work both opening issues and creating this first pr. i think we can merge it. i have not tested this locally yet  as i'm short of time but thought i could do so after your next pr to keep things moving.

src/stravalib/model.py
jsamoocha: See my comment in #500. The impact should be minimal. The main breaking change will be if users are doing explicit instance checking such as `isinstance(obj_attribute, Quantity)`. That will not work anymore since we'll return a different type of object, but the behavior will remain the same.

PR #505, opened 2024-05-12
Adds Strava type hierarchy for Activity type

### LINKED ISSUE #499: Stravalib2.0 -> Maintain Strava's meta-summary-detailed type hierarchy
This should prevent response objects with many `None` fields and make it more explicit to users what level of detail they'll get from an API call.

- [x] Update client methods to ensure the correct detail level is returned
- [x] Check the model for instances where we override stuff from other than detailed-level objects

Depends on #498 

**Update:** As a reminder: this will also close #198

jsamoocha:
This is a tiny PR to show how making the Strava type hierarchy explicit in our client responses can improve type-checking accuracy and prevent user confusion (see #494).

This approach can be used to resolve #499.

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -271,7 +271,7 @@ class Client:
         before: datetime | str | None = None,
         after: datetime | str | None = None,
         limit: int | None = None,
-    ) -> BatchedResultsIterator[model.Activity]:
+    ) -> BatchedResultsIterator[model.SummaryActivity]:
         """Get activities for authenticated user sorted by newest first.
 
         https://developers.strava.com/docs/reference/#api-Activities-getLoggedInAthleteActivities
@@ -290,7 +290,7 @@ class Client:
         Returns
         -------
         class:`BatchedResultsIterator`
-            An iterator of :class:`stravalib.model.Activity` objects.
+            An iterator of :class:`stravalib.model.SummaryActivity` objects.
 
         """
 
@@ -302,7 +302,7 @@ class Client:
         )
 
         return BatchedResultsIterator(
-            entity=model.Activity,
+            entity=model.SummaryActivity,
             bind_client=self,
             result_fetcher=result_fetcher,
             limit=limit,
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -38,7 +38,6 @@ from stravalib.strava_model import (
     ActivityZone,
     BaseStream,
     Comment,
-    DetailedActivity,
     DetailedAthlete,
     DetailedClub,
     DetailedGear,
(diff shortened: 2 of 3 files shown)


src/stravalib/client.py
```diff
@@ -271,7 +271,7 @@ def get_activities(
         before: datetime | str | None = None,
         after: datetime | str | None = None,
         limit: int | None = None,
-    ) -> BatchedResultsIterator[model.Activity]:
+    ) -> BatchedResultsIterator[model.SummaryActivity]:
```
jsamoocha, line 274: For every endpoint, we should check the Strava API docs for the correct return type

src/stravalib/model.py
```diff
@@ -889,22 +888,65 @@ class SegmentEffort(BaseEffort):
     achievements: Optional[list[SegmentEffortAchievement]] = None
 
 
+class MetaActivity(strava_model.MetaActivity, BoundClientEntity):
+    @lazy_property
+    def comments(self) -> BatchedResultsIterator[ActivityComment]:
+        """Retrieves comments for a specific activity id."""
+        assert self.bound_client is not None
+        return self.bound_client.get_activity_comments(self.id)
```
jsamoocha, line 896: All these lazy properties can be defined at the meta level, because they only depend on the `id` field.

src/stravalib/model.py
```diff
@@ -889,22 +888,65 @@ class SegmentEffort(BaseEffort):
...
     """
 
     # field overrides from superclass for type extensions:
-    athlete: Optional[Athlete] = None
-    start_latlng: Optional[LatLon] = None
-    end_latlng: Optional[LatLon] = None
-    map: Optional[Map] = None
     gear: Optional[Gear] = None
-    type: Optional[RelaxedActivityType] = None
-    sport_type: Optional[RelaxedSportType] = None
```
jsamoocha, line 907: These fields are move to the `SummaryActivity` superclass, because that's where they're defined in Strava's type hierarchy.

src/stravalib/tests/integration/test_client.py
```diff
@@ -813,6 +814,7 @@ def test_get_activities(
     activity_list = list(client.get_activities(**kwargs))
     assert len(activity_list) == expected_n_activities
     if expected_n_activities > 0:
+        assert isinstance(activity_list[0], SummaryActivity)
```
jsamoocha, line 817: So now, the response from `get_activities()` doesn't have the confusing, always-`None` `laps` attribute anymore.

src/stravalib/client.py
lwasser: should this be a new issue @jsamoocha ?

src/stravalib/model.py
lwasser: this will make things a lot cleaner!

src/stravalib/tests/integration/test_client.py
lwasser: I just tested locally. this works great. i only got information that i could actually retrieve back! 
NOTE: i rarely use laps so at some point i'll create an activity with laps to fully test this. but i don't see an empty laps attr anymore in the response

lwasser (approved):
Fantastic. i did actually test this out locally but only on high level activities. let's merge!

src/stravalib/client.py
jsamoocha: I added the list of all endpoints to your new issue.

PR #507, opened 2024-05-13
Athlete stats

jsamoocha:
This fixes the type hierarchy for athlete-related objects.

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -528,7 +528,7 @@ class Client:
 
     def get_club_members(
         self, club_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.Athlete]:
+    ) -> BatchedResultsIterator[strava_model.ClubAthlete]:
         """Gets the member objects for specified club ID.
 
         https://developers.strava.com/docs/reference/#api-Clubs-getClubMembersById
@@ -551,7 +551,7 @@ class Client:
         )
 
         return BatchedResultsIterator(
-            entity=model.Athlete,
+            entity=strava_model.ClubAthlete,
             bind_client=self,
             result_fetcher=result_fetcher,
             limit=limit,
@@ -1062,7 +1062,7 @@ class Client:
 
     def get_activity_kudos(
         self, activity_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.ActivityKudos]:
+    ) -> BatchedResultsIterator[model.SummaryAthlete]:
         """Gets the kudos for an activity.
 
         https://developers.strava.com/docs/reference/#api-Activities-getKudoersByActivityId
@@ -1085,7 +1085,7 @@ class Client:
         )
 
         return BatchedResultsIterator(
-            entity=model.ActivityKudos,
+            entity=model.SummaryAthlete,
             bind_client=self,
             result_fetcher=result_fetcher,
             limit=limit,
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -588,6 +588,38 @@ class Client:
             limit=limit,
         )
 
+    def get_club_admins(
+        self, club_id: int, limit: int | None = None
+    ) -> BatchedResultsIterator[model.SummaryAthlete]:
+        """Gets the activities associated with specified club.
+
+        https://developers.strava.com/docs/reference/#api-Clubs-getClubAdminsById
+
+        Parameters
+        ----------
+        club_id : int
+            The numeric ID for the club.
+        limit : int
+            Maximum number of admins to return. (default unlimited)
+
+        Returns
+        -------
+        class:`BatchedResultsIterator`
+            An iterator of :class:`stravalib.model.SummaryAthlete` objects.
+
+        """
+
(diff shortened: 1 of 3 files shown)


src/stravalib/client.py
```diff
@@ -528,7 +528,7 @@ def get_club(self, club_id: int) -> model.Club:
 
     def get_club_members(
         self, club_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.Athlete]:
+    ) -> BatchedResultsIterator[strava_model.ClubAthlete]:
```
jsamoocha, line 531: This is a new type introduced by Strava a while ago. We never updated this client method to reflect that update.

src/stravalib/client.py
```diff
@@ -588,6 +588,38 @@ def get_club_activities(
...
+        result_fetcher = functools.partial(
+            self.protocol.get, "/clubs/{id}/admins", id=club_id
+        )
+
+        return BatchedResultsIterator(
+            entity=model.SummaryAthlete,
+            bind_client=self,
+            result_fetcher=result_fetcher,
+            limit=limit,
+        )
```
jsamoocha, line 621: This API method was never implemented

src/stravalib/model.py
```diff
@@ -499,20 +500,6 @@ def to_str_representation(
...
-        for the (authenticated) athlete.
-
-        Returns
-        -------
-        Athlete
-            The detailed information of the authenticated athlete.
-        """
-        assert self.bound_client is not None, "Bound client is not set."
-        return self.bound_client.get_athlete()
-
```
jsamoocha, line 515: Removing this (and related methods). It seems this construction aimed to prevent the `stats` property from being accessed on a non-authenticated athlete instance (which would raise a 401). But since we now know that the `stats` property is only available on the (detailed) `Athlete` class, we know for sure this is the authenticated user.

src/stravalib/model.py
```diff
@@ -677,18 +634,6 @@ class name, photo type, and a key identifier."""
...
-    """
-    Represents kudos an athlete received on an activity.
-
-    Notes
-    -----
-    Activity kudos are a subset of athlete properties.
-    """
-
-    pass
-
```
jsamoocha, line 690: This is empty and should be `SummaryAthlete`, so I replaced its use by the latter.

src/stravalib/client.py
```diff
@@ -588,6 +588,38 @@ def get_club_activities(
             limit=limit,
         )
 
+    def get_club_admins(
```
lwasser, line 591: can you please test this one out? i'm not a club admin myself so i'm not sure if i have access to this via the clubs i'm in. 

i tried to call it and i don't get anything back which is odd. 
```
all_admins = client.get_club_admins(club_id)
for admin in all_admins:
    print(admin)
``` 

i don't have time to troubleshoot right now but it could just be a permissions thing. if so we might want to adjust the docstring.

src/stravalib/client.py
```diff
@@ -588,6 +588,38 @@ def get_club_activities(
...
+        self, club_id: int, limit: int | None = None
+    ) -> BatchedResultsIterator[model.SummaryAthlete]:
+        """Gets the activities associated with specified club.
```
lwasser, line 594:
```suggestion
        """Returns a list of the administrators of a given club.
```

i think?  i copied this from swagger.json

src/stravalib/client.py
```diff
@@ -1085,7 +1117,7 @@ def get_activity_kudos(
         )
 
         return BatchedResultsIterator(
-            entity=model.ActivityKudos,
+            entity=model.SummaryAthlete,
```
lwasser, line 1120: this seems to work well. I am noticing some none fields for returned athletes. i haven't checked to see if their profiles are empty or not but these are definitely people who have used strava for some time. 
```
activities = client.get_activities(after="2024-05-01")
print(activities)

# Currently the activity object does have laps: None
first_activity = next(iter(activities))
first_activity.kudos_count  # 8
kudos = client.get_activity_kudos(first_activity.id)
for kudo in kudos:
    print(kudo)
```
Responses do consistently contain Nones and i know most atleast have their country in their profiles.

```
bound_client=<stravalib.client.Client object at 0x1306a3a90> id=None city=None country=None created_at=None firstname='Dan' lastname='F.' premium=None profile=None profile_medium=None resource_state=2 sex=None state=None summit=None updated_at=None
```

lwasser (approved):
@jsamoocha this looks good. i've left a few notes. i'm getting a bunch of none values for kudo summary athlete returns. i'm not sure if that is intended or not but haven't had time tonight to look into it more. 

i also couldn't get club admins to return anything. Note that i'm also getting a really slow response time so i'm not sure if that is my internet or something else that might be throwing off the calls and returns.  i think it could be worth looking into to see if you get the same thing of ir this is my internet returning "blank" values because things are slow.

src/stravalib/client.py
jsamoocha: I'm getting a normal response with an iterator of `SummaryAthlete` objects, regardless of whether I'm an admin or not in that club.

PR #516, opened 2024-06-01
Add: test for latlon values

lwasser:
@jsamoocha can you give a tip or help me trouble shoot this? i think having this test is good and am totally open to writing it in a better way. 

✅ Everything validates for summary activity if we have valid lat long numbers. 
✅ Nones are also acceptable and ingested into the model properly. 
❌ When stravalib gets an empty list, it fails. I understand generally why it's failing. but that means our validator is not working as expected and/or things are typed incorrectly. (when i debug the validator seems to work) 

A few notes:

▶️  we have [this validator here](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L145) that lives in the LatLon object. 
▶️  then we [have this](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L145) -- a method that lives in the model.py module.  

This method is called 4 times in module.py 

☑️  [in ActivityPhoto here](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L605)
☑️  [in activityPhoto here](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L690)
☑️  [In summarize segment here](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L755)
☑️  [In ActivityPhoto here](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L935)

It seems to me that we could create a single method and use it several times as a validator (if needed). 

## Type weirdness
But here is the real issue - When i look at the [generated model -- it expects ](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/strava_model.py#L227) a list[float] - None is not an option nor is an empty list. But we know Strava can return an empty list or have missing data.

In the LatLon object in [the model module](https://github.com/stravalib/stravalib/blob/pydantic-v2/src/stravalib/model.py#L254) we ingest lat/lon but there is no type override allowing for None values. But i think we had discussed wanting to avoid type overrides where possible? 

regardless, I tried to override the field like this:

```
root: list[float] | None = Field(..., max_length=2, min_length=2)
```

But the test still doesn't pass. i'm sure i'm missing something. I was hoping you could help me as i've spent a good bit of time trying to troubleshoot and i must be missing something. many many thanks!


lwasser:
so sorry - i keep opening pr's against main by mistake 🙈

src/stravalib/model.py
```diff
@@ -281,7 +283,8 @@ def check_valid_latlng(cls, values: list[float]) -> Optional[list[float]]:
 
         """
         # Strava sometimes returns empty list in case of activities without GPS
-        return values if values else None
+        a = values if values else None
```
lwasser, line 286: this was just me debugging.. i'll remove once i understand what is going on here.

src/stravalib/model.py
```diff
@@ -687,6 +690,7 @@ class SegmentExplorerResult(
     # Undocumented attributes:
     starred: Optional[bool] = None
 
+    # Could we avoid this step and just assign variables in the validator? the validator
```
lwasser, line 693: i was just wondering here why we explicitly call the validator vs just adding a validator to the property. perhaps we can't do that? 

i also was confused as to why this is a property at all vs a list of values. again i'm certain i'm missing something. i've used the .lat attr before i think.

src/stravalib/model.py
```diff
@@ -871,9 +875,8 @@ def full_photos(self) -> BatchedResultsIterator[ActivityPhoto]:
 class SummaryActivity(MetaActivity, strava_model.SummaryActivity):
     # field overrides from superclass for type extensions:
     athlete: Optional[MetaAthlete] = None
-    start_latlng: Optional[LatLon] = (
-        None  # Do we still need these overrides for latlon?
-    )
+    # These force validator to run on lat/lon
```
lwasser, line 878: i wasn't sure why there was this output with tuple like brackets so i removed it. again there could be some meaning here that i'm missing. we do need the overrides because of how the generated model gets typed. it does not allow None as typed.

src/stravalib/tests/unit/test_lat_long.py
```diff
@@ -0,0 +1,29 @@
...
+    "start_latlng, end_latlng, start_expected_return, end_expected_return",
+    [
+        (
+            [10.2, 1.02],
+            [10.2, 1.02],
+            LatLon(root=[10.2, 1.02]),
+            LatLon(root=[10.2, 1.02]),
+        ),
+        (None, None, None, None),
+        ([], [], None, None),
```
lwasser, line 16: This fails consistently. i have an activity that for some reason had an empty list - that is how i found this. so i thought a test would be warranted here and then we will want to fix this too. many thanks. i'm just not sure where the typing issue is coming from and how we can clean things up.

lwasser:
NOTE: i just updated the pr summary so hopefully it makes more sense and is more specific!

src/stravalib/client.py
```diff
@@ -2035,6 +2035,10 @@ def _fill_buffer(self) -> None:
 
         entities = []
         for raw in raw_results:
+            # Note: this fails to validate if the start and end latlng are empty
```
lwasser, line 2038: If they are empty lists. so this is a "run" that was super short and had gps data (you can see the map on straval) but no start/end latlon data propulated

src/stravalib/tests/unit/test_lat_long.py
```diff
@@ -0,0 +1,29 @@
...
+    "start_latlng, end_latlng, start_expected_return, end_expected_return",
+    [
+        (
+            [10.2, 1.02],
+            [10.2, 1.02],
+            LatLon(root=[10.2, 1.02]),
+            LatLon(root=[10.2, 1.02]),
+        ),
+        (None, None, None, None),
+        ([], [], LatLon(root=None), LatLon(root=None)),
```
lwasser, line 16: this test was failing because i thought it woudl return none but really it returns a LatLon object with root = None

lwasser:
huh - things are working now.

* I'd still like a review to ensure my fix was valid. as i did override the default model generated type. 
* Maybe i can merge the two validator methods. 
* mypy is not happy about the override

src/stravalib/model.py
```diff
@@ -253,13 +253,17 @@ def check_sport_type(cls, values: str) -> str:
 
 class LatLon(LatLng):
     """
-    Enables backward compatibility for legacy namedtuple
+    Stores lat / lon values or None.
```
lwasser, line 256: if root=None, then LatLon.lat will return a TypeError

jsamoocha:
My bad, and I totally understand your confusion. In #505, I created the `SummaryActivity` type and moved some fields there from our (detailed-) `Activity`, among which `start_latlng`. However, I forgot to move the _validator_ for `latlng` fields to that superclass. So whenever we try to deserialize a `SummaryActivity` (instead of a detailed `Activity`), the latlng fields are not validated properly, leading to validation errors in some edge cases.

The following validator in `model.Activity` just needs to be moved to the `model.SummaryActivity` superclass:

```python
_latlng_check = field_validator(
    "start_latlng", "end_latlng", mode="before"
)(check_valid_location)
```

I don't think any of the other changes in this PR are necessary to make lat/lon handling work again. The added test is ok, but we can also add a few cases to the `test_deserialization_edge_cases()` test in `test_model`.

lwasser:
ok great. i'll move the validator, remove the other field change i made and add some tests later today! i also want to test what happens to the .lat and .lon properties.

src/stravalib/tests/unit/test_model.py
```diff
@@ -30,16 +30,31 @@
     "model_class,raw,expected_value",
     (
         (
-            # Removing root
+            Activity,
```
lwasser, line 33: this is a bit redundant. but i did test for both start and end. 

we don't actually call the `.lat` and `.lon` attrs anywhere. do we need to? if we do that would need to be a separate test.

lwasser:
ok everything is running here. gosh that small thing reallyl threw me off. next time i'll check the previous merged pr's more closely.

@jsamoocha only question i have is - we aren't explicitly checking .lat and .lon attrs. should we be? thanks for the help on this one!

src/stravalib/model.py
```diff
@@ -871,14 +865,17 @@ def full_photos(self) -> BatchedResultsIterator[ActivityPhoto]:
...
-        None  # Do we still need these overrides for latlon?
-    )
+    # These force validator to run on lat/lon
+    start_latlng: Optional[LatLon] = None
     end_latlng: Optional[LatLon] = None
     map: Optional[Map] = None
     type: Optional[RelaxedActivityType] = None
     sport_type: Optional[RelaxedSportType] = None
 
+    _latlng_check = field_validator(
```
lwasser, line 875: this is what was breaking everything. i totally missed that the attrs weren't defined in the other class but here (had been moved) thanks again @jsamoocha ✨

jsamoocha (approved):
Ok, great, this is good to merge!

src/stravalib/tests/unit/test_model.py
jsamoocha: These properties do not have any complicated business logic, so they do not need to be tested separately, IMO.

lwasser:
thank you @jsamoocha !! merging!

Issue #506, opened 2024-05-12
Stravalib 2.0 -> For every endpoint, we should check the Strava API docs for the correct return type


lwasser:
For every endpoint, we should check the Strava API docs for the correct return type

_Originally posted by @jsamoocha in https://github.com/stravalib/stravalib/pull/505#discussion_r1597580080_
            
I'm just turning this into an issue so we remember to do it. i'm happy to help with this when we are ready to do it.

jsamoocha:
Copying this list from #250 for reference. We can let one another know what we're working on by checking the boxes below:

- [ ] /athletes/{id}/stats (GET)
- [ ] /athlete (GET)
- [ ] /athlete (PUT)
- [ ] /athlete/zones (GET)
- [ ] /segments/{id} (GET)
- [ ] /segments/starred (GET)
- [ ] ~~/segments/{id}/starred (PUT)~~ - NOT IMPLEMENTED AT ALL
- [ ] /segment_efforts (GET)
- [ ] /segments/explore (GET)
- [ ] /segment_efforts/{id} (GET)
- [ ] /activities (POST)
- [ ] /activities/{id} (GET)
- [ ] /activities/{id} (PUT)
- [x] /athlete/activities (GET)
- [ ] /activities/{id}/laps (GET)
- [ ] /activities/{id}/zones (GET)
- [ ] /activities/{id}/comments (GET)
- [ ] /activities/{id}/kudos (GET)
- [ ] /clubs/{id} (GET)
- [ ] /clubs/{id}/members (GET)
- [ ] /clubs/{id}/admins (GET)
- [ ] /clubs/{id}/activities (GET)
- [ ] /athlete/clubs (GET)
- [ ] /gear/{id} (GET)
- [ ] /routes/{id} (GET)
- [ ] /athletes/{id}/routes (GET)
- [ ] /routes/{id}/export_gpx (GET)
- [ ] /routes/{id}/export_tcx (GET)
- [ ] /uploads (POST)
- [ ] /uploads/{uploadId} (GET)
- [ ] /activities/{id}/streams (GET)
- [ ] /segment_efforts/{id}/streams (GET)
- [ ] /segments/{id}/streams (GET)
- [ ] /routes/{id}/streams (GET)

lwasser:
all items in the list for this issue have been checked. I am going to close this as complete. please reopen if there are things i'm missing!!

Issue #510, opened 2024-05-18
Stravalib 2.0 -> Replace invariant by covariant container types


jsamoocha:
This should resolve various typing issues we're now suppressing that are due to overriding invariant container types. See https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generics. In practice, our model classes should use `Sequence` or `Mapping` container type instead of, e.g, `list` or `dict`. There's an issue in the bot repo to modify the code generator in order to have covariant container types in the strava model as well.

lwasser:
i am going to take a stab at this next @jsamoocha if that sounds good to you. i've wanted to also go through and update types (dropping Optional even tho i really like the expressiveness of Optional) so i could do this in two pr's. i'll start with the [bot repo pr first](https://github.com/stravalib/strava_swagger2pydantic/issues/13).

lwasser:
[pr here](https://github.com/stravalib/strava_swagger2pydantic/pull/14/files)

I am thinking about also implementing the use Union = True. this would then require updating ALL instances of optional throughout our code which would create merge conflicts potentially.  so i'll do this in pieces and will coordinate with you @jsamoocha for when you are not working on things before implementing that change!

lwasser:
@jsamoocha i started to play with this in #530 do we want to convert ALL list elements to Sequence or should we do it strategically? I understand generally how this works and the value of Sequence relative to our typing issues. However I wasn't sure if we wanted to migrate list /dict universally --> Sequence or not.

lwasser:
are there other pieces to this that need to be implemented @jsamoocha or can we close this issue now?

jsamoocha:
This is all done 👍

PR #517, opened 2024-06-02
Fix: fix return types - athlete clubs endpoint

lwasser:
This adjusts the return types for the athlete/clubs endpoint. It also updates a test and modifies the return object type to be an iterator vs a list. 

#506

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -468,8 +468,6 @@ class Client:
 
         """
 
-        # TODO: This should return a BatchedResultsIterator or otherwise at
-        # most 30 clubs are returned!
         result_fetcher = functools.partial(self.protocol.get, "/athlete/clubs")
         # club_structs = self.protocol.get("/athlete/clubs")
 
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -469,8 +469,6 @@ class Client:
         """
 
         result_fetcher = functools.partial(self.protocol.get, "/athlete/clubs")
-        # club_structs = self.protocol.get("/athlete/clubs")
-
         return BatchedResultsIterator(
             entity=model.SummaryClub,
             bind_client=self,
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -451,7 +451,6 @@ class Client:
 
         return model.AthleteStats.model_validate(raw)
 
-    # TODO - create model.summaryClub
     def get_athlete_clubs(
         self, limit: int | None = None
     ) -> BatchedResultsIterator[model.SummaryClub]:
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -285,7 +285,7 @@ class Client:
             Result will start with activities whose start date is after
             specified value. (UTC)
         limit : int or None, default=None
-            How many maximum activities to return.
+            Maximum number of activities to return.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


lwasser:
i'm going to add endpoint docs for each one that i test in this pr. that way it's easy to maintain links - 

i can setup htmlproofer or lychee to run on our docs. that will find broken links. so if an endpoint link fails we will know about it. i'll turn this into a draft and will work on that here as well.

src/stravalib/client.py
```diff
@@ -468,7 +468,7 @@ def get_athlete_clubs(self) -> list[model.Club]:
         # most 30 clubs are returned!
```
lwasser, line 468: I see this comment here. it is true that if someone is a member of more then 30 clubs, it will only return 30. should we fix this now?

src/stravalib/client.py
```diff
@@ -451,26 +451,35 @@ def get_athlete_stats(
 
         return model.AthleteStats.model_validate(raw)
 
-    def get_athlete_clubs(self) -> list[model.Club]:
+    # TODO - create model.summaryClub
```
lwasser, line 454:
```suggestion
```

src/stravalib/model.py
```diff
@@ -361,6 +364,16 @@ def activities(self) -> BatchedResultsIterator[Activity]:
         return self.bound_client.get_club_activities(self.id)
 
 
+class SummaryClub(strava_model.SummaryClub):
```
lwasser, line 367: I created this and added the single undocumented attr

src/stravalib/client.py
```diff
@@ -451,26 +451,30 @@ def get_athlete_stats(
...
         """
 
-        # TODO: This should return a BatchedResultsIterator or otherwise at
-        # most 30 clubs are returned!
-        club_structs = self.protocol.get("/athlete/clubs")
-        return [
-            model.Club.model_validate({**raw, **{"bound_client": self}})
-            for raw in club_structs
-        ]
+        result_fetcher = functools.partial(self.protocol.get, "/athlete/clubs")
```
lwasser, line 471: do we need params to specify per page, etc? i wasn't sure. i only have 3 clubs and one is a dummy club that i created. so i can't test on that many clubs.

lwasser:
@jsamoocha can you please review? I made some changes to the response following comments and also the spec returned  `SummaryClub` so i created that object and inherited from strava_model. 

The tests are now failing because this no longer returns a list. I am still wrapping my head around how the API mock + batched iterator works so i could use a bit of guidance (once i know that the changes i made are ok). i did look at the get activities test. i kind of understand how you're creating x number of elements within the iterator but i don't understand the "matcher" part of it and how to set things up for clubs. 

many thanks!

src/stravalib/client.py
```diff
@@ -451,26 +451,30 @@ def get_athlete_stats(
...
+        return BatchedResultsIterator(
```
lwasser, line 472: now returning an iterator following the comment to support >30 clubs (do people have that many clubs?!)

src/stravalib/tests/integration/test_client.py
```diff
@@ -733,14 +733,19 @@ def test_get_club(mock_strava_api, client):
...
+    """Test that athlete clubs returns the correct
+    number of clubs"""
+
+    # Create a mock instance with the club name update to my club
     mock_strava_api.get(
-        "/athlete/clubs", response_update={"name": "foo"}, n_results=n_clubs
+        "/athlete/clubs", response_update={"name": "myclub"}, n_results=n_clubs
     )
-    clubs = client.get_athlete_clubs()
+    # Convert iterator to list
```
lwasser, line 744: Small fix to acomodate the new return type of iterator vs list.

lwasser:
everything here is happy except mypy now because of the new SummaryClub type return.

src/stravalib/client.py
jsamoocha: Do you mean we maybe need to pass additional arguments to `functools.partial()`?

jsamoocha (approved):
Thanks, this is good! The mypy issue will be fixed by #510.

src/stravalib/client.py
lwasser: @jsamoocha - sorry! your interpretation is correct. Do we need to pass `functools.partial()` any additional arguments? I was worried about this because other methods do get other arguments for instance get_club_members ingests club_id in the id param. and get activities gets the params of before and after, etc. 

I think the answer here is no given this just returns the authenticated athlete's clubs so there might be nothing to pass. But i wanted to be certain there wasnt something like a param to specify how many clubs to return that i was missing.

lwasser:
awesome. i'll merge this one. thanks @jsamoocha  i'll then rebase the other branches

PR #518, opened 2024-06-02
Adds SummarySegmentEffort superclass

jsamoocha:
Tiny PR for adding the `SummarySegmentEffort` superclass.

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -28,7 +28,13 @@ from typing import (
 )
 
 from dateutil import parser
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import (
+    AliasChoices,
+    BaseModel,
+    Field,
+    field_validator,
+    model_validator,
+)
 
 from stravalib import exc, strava_model
 from stravalib.strava_model import (
@@ -48,7 +54,6 @@ from stravalib.strava_model import (
     SportType,
     SummaryClub,
     SummaryGear,
-    SummaryPRSegmentEffort,
     TimedZoneRange,
 )
 
@@ -707,11 +712,19 @@ class SegmentExplorerResult(
 
 
 class AthletePrEffort(
-    SummaryPRSegmentEffort,
+    strava_model.SummaryPRSegmentEffort,
 ):
+    # Override fields from superclass to match actual responses by Strava API:
+    activity_id: Optional[int] = Field(
+        validation_alias=AliasChoices("pr_activity_id", "activity_id"),
+        default=None,
+    )
+    elapsed_time: Optional[timedelta] = Field(
+        validation_alias=AliasChoices("pr_elapsed_time", "elapsed_time"),
+        default=None,
+    )
+
     # Undocumented attributes:
-    activity_id: Optional[int] = None  # see pr_activity_id in superclass
-    elapsed_time: Optional[int] = None  # see pr_elapsed_time in superclass
     distance: Optional[float] = None
     start_date: Optional[datetime] = None
     start_date_local: Optional[datetime] = None
@@ -724,9 +737,11 @@ class SummarySegment(strava_model.SummarySegment, BoundClientEntity):
     # Field overrides from superclass for type extensions:
     start_latlng: Optional[LatLon] = None
     end_latlng: Optional[LatLon] = None
-    athlete_segment_stats: Optional[AthleteSegmentStats] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
     activity_type: Optional[RelaxedActivityType] = None  # type: ignore[assignment]
+    athlete_segment_stats: Optional[AthleteSegmentStats] = (
+        None  # Actually, this is only part of a (detailed) segment response
+    )
 
     _latlng_check = field_validator(
         "start_latlng", "end_latlng", mode="before"
@@ -780,6 +795,16 @@ class SegmentEffortAchievement(BaseModel):
 
 
 class SummarySegmentEffort(strava_model.SummarySegmentEffort):
+    # Override superclass fields to match actual Strava API responses
+    activity_id: Optional[int] = Field(
+        validation_alias=AliasChoices("pr_activity_id", "activity_id"),
+        default=None,
+    )
+    elapsed_time: Optional[timedelta] = Field(
+        validation_alias=AliasChoices("pr_elapsed_time", "elapsed_time"),
+        default=None,
+    )
+
     _naive_local = field_validator("start_date_local")(naive_datetime)
 
 
@@ -823,7 +848,6 @@ class AthleteSegmentStats(
 
     # Undocumented attributes:
     effort_count: Optional[int] = None
-    pr_elapsed_time: Optional[timedelta] = None
     pr_date: Optional[date] = None
 
 
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -719,7 +719,7 @@ class AthletePrEffort(
         validation_alias=AliasChoices("pr_activity_id", "activity_id"),
         default=None,
     )
-    elapsed_time: Optional[timedelta] = Field(
+    elapsed_time: Optional[int] = Field(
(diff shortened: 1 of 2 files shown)


src/stravalib/model.py
```diff
@@ -708,21 +706,6 @@ def segment(self) -> Segment:
...
-    """
-
-    # Undocumented attributes:
-    effort_count: Optional[int] = None
-    pr_elapsed_time: Optional[timedelta] = None
-    pr_date: Optional[date] = None
-
-    _naive_local = field_validator("start_date_local")(naive_datetime)
-
-
```
jsamoocha, line 725: This class is just moved down to prevent SummarySegmentEffort from being undefined

src/stravalib/model.py
```diff
@@ -811,10 +799,8 @@ class BaseEffort(
 
     # Field overrides from superclass for type extensions:
     segment: Optional[SummarySegment] = None
-    activity: Optional[Activity] = None
-    athlete: Optional[Athlete] = None
-
-    _naive_local = field_validator("start_date_local")(naive_datetime)
+    activity: Optional[MetaActivity] = None
+    athlete: Optional[MetaAthlete] = None
```
jsamoocha, line 803: Tiny fix: these should be meta types per the API spec

src/stravalib/model.py
```diff
@@ -1063,8 +1062,3 @@ class SubscriptionUpdate(BaseModel):
     aspect_type: Optional[str] = None
     event_time: Optional[datetime] = None
     updates: Optional[dict[str, Any]] = None
-
-
-SegmentEffort.update_forward_refs()
-ActivityLap.update_forward_refs()
-BestEffort.update_forward_refs()
```
jsamoocha, line 1070: These are not necessary anymore

lwasser:
hey @jsamoocha thanks for this! i've actually never played with the segment data before. so a few odd things when i test this out i get two items that should be objects according to the [strava spec for summarySegment](https://developers.strava.com/docs/reference/#api-models-SummarySegment).

# First try - get_starred_segments - 
I just tried to explore segments i've starred. Some data were returned with missing things - see below


```python
athlete_info = client.get_athlete()
athlete_info.lastname # all seems to be working
starred = client.get_starred_segments()
# If you look at the data in a starred segment, there are some missing items -- see below
seg = next(iter(starred))

```

## athlete_pr_effort -- 

_SummaryPRSegmentEffort | An instance of SummaryPRSegmentEffort._

This attribute  is populated with an AthletePrEffort object. however most of the data in the return is empty

Screenshot 2024-06-02 at 6 48 42 PM

i looked at this for several efforts that i've done probably dozens if not more times.  and the data are missing. not all of the data but it should give me these [elements at a minimum](https://developers.strava.com/docs/reference/#api-models-SummaryPRSegmentEffort) 

Screenshot 2024-06-02 at 6 50 30 PM

And those values are all None for me. NOTE: i haven't traced this back to the raw data yet to see what is actually returned. 

# athlete_segment_stats 

this item should return an object [according to the spec](https://developers.strava.com/docs/reference/#api-models-SummarySegment). But instead it just returns None. 

SummarySegmentEffort | An instance of [SummarySegmentEffort](https://developers.strava.com/docs/reference/#api-models-SummarySegmentEffort).

Screenshot 2024-06-02 at 6 56 23 PM

********

Do you need more info to track this down? it seems like the athlete_segment_stats are not implemented properly?
What is interesting is in the model it says some of the summary segment stats are not documented. they are in the spec

# Try 2 -- get segment by id
so now i try to get a segment that above is missing info by id 

```python
sanitas_segment_id = 709312
seg_sanitas = client.get_segment(sanitas_segment_id)
```
this time athlete_segment_stats is there. and as some info but i've got some fields populated with None (for instance distance of the segment is None but the return in the spec data shows a value). Thus - i think it's  missing some info that is [shown in the same response](https://developers.strava.com/docs/reference/#api-Segments-getSegmentById) in the stravadocs

Screenshot 2024-06-02 at 7 09 51 PM

Screenshot 2024-06-02 at 7 12 18 PM

pr date and time are actually there in athlete_segment stats but also there is some None values there as well. 
Screenshot 2024-06-02 at 7 13 55 PM


*******
So i think this is partially working but it looks like a few wires may be crossed here. can you please have a look to see if you get a similar return ?

jsamoocha:
Thanks for reviewing this.

1. I haven't started working on `SummaryPRSegmentEffort` yet. I can add it to this PR to reduce the overhead of another PR.
2. Regardless of the spec, we need to figure out what Strava actually returns using curl, Postman, or a debugger. We can then determine whether our model has problems parsing this (raw) response or whether Strava returns something different than specified.

jsamoocha:
Here's the raw response from Strava for `/segments/starred`:

```
"athlete_pr_effort": {
          "id": 2785851230931490230,
          "activity_id": 4634661314,
          "elapsed_time": 361,
          "distance": 1627.0,
          "start_date": "2021-01-17T11:19:09Z",
          "start_date_local": "2021-01-17T12:19:09Z",
          "is_kom": false
      }
```

So Strava is using different field names than in the API spec. I fixed that in https://github.com/stravalib/stravalib/pull/518/commits/ebef7b15f75ff53efde2a1ba26bb97abdb704a9c. I'll quickly check the other issue as well.

jsamoocha:
Ok, that was quite a mess: many field names from actual responses did not match the Strava API spec:

1. `athlete_segment_stats` is specified in `SummarySegment`, but only returned in `DetailedSegment`s,
2. `athlete_pr_effort` is specified in `SummarySegment`, but _not_ returned in `DetailedSegment`s (while these are a subclass of the summary),
3. fields with a `pr_` prefix are returned without that prefix, and fields without the prefix are returned with it. I added several aliases to deal with the current situation and anticipate on a potential future fix by Strava.

This should take care of the issues you raised.

src/stravalib/tests/unit/test_model.py
```diff
@@ -73,11 +124,6 @@ def test_subscription_callback_field_names():
     assert sub_callback.hub_verify_token == "STRAVA"
 
 
-class ExtLatLng(LatLng):
-    def foo(self):
-        return f"[{self[0], self[1]}]"
-
```
jsamoocha, line 79: This wasn't used anywhere

src/stravalib/model.py
```diff
@@ -801,8 +794,23 @@ class SegmentEffortAchievement(BaseModel):
...
+class SummarySegmentEffort(strava_model.SummarySegmentEffort):
+    # Override superclass fields to match actual Strava API responses
+    activity_id: Optional[int] = Field(
+        validation_alias=AliasChoices("pr_activity_id", "activity_id"),
+        default=None,
+    )
+    elapsed_time: Optional[int] = Field(
+        validation_alias=AliasChoices("pr_elapsed_time", "elapsed_time"),
+        default=None,
+    )
```
jsamoocha, line 806: These are the new aliases that should handle both flavors (i.e., with or without `pr_` prefix) of field names.

lwasser (approved):
> Ok, that was quite a mess: many field names from actual responses did not match the Strava API spec:
> 
> 1. `athlete_segment_stats` is specified in `SummarySegment`, but only returned in `DetailedSegment`s,
> 2. `athlete_pr_effort` is specified in `SummarySegment`, but _not_ returned in `DetailedSegment`s (while these are a subclass of the summary),
> 3. fields with a `pr_` prefix are returned without that prefix, and fields without the prefix are returned with it. I added several aliases to deal with the current situation and anticipate on a potential future fix by Strava.
> 
> This should take care of the issues you raised. It was good you took a closer look at what was actually returned. If you see no problems with these fixes, we can merge them.

WOW - @jsamoocha i have postman setup and i now understand what you're saying. 

this all looks good. i've tested it locally with stravalib and mapped it against the actual returns in postman... and now 🤯 let's merge. 

is it worth submitting an issue to strava somewhere about the api? there are a LOT of inconsistencies in the spec vs the actual returned data.

PR #519, opened 2024-06-06
Add: correct type returns following clubs/{id}  & clubs/{id}/activities spec

lwasser:
This pr adjusts the expected return types following strava's spec but also looking at the actual return elements in postman. 

Also addresses part of #506

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -512,7 +512,7 @@ class Client:
         """
         self.protocol.post("clubs/{id}/leave", id=club_id)
 
-    def get_club(self, club_id: int) -> model.Club:
+    def get_club(self, club_id: int) -> model.DetailedClub:
         """Return a specific club object.
 
         https://developers.strava.com/docs/reference/#api-Clubs-getClubById
@@ -524,11 +524,13 @@ class Client:
 
         Returns
         -------
-        class: `model.Club` object containing the club data.
+        class: `model.DetailedClub` object containing the club data.
 
         """
         raw = self.protocol.get("/clubs/{id}", id=club_id)
-        return model.Club.model_validate({**raw, **{"bound_client": self}})
+        return model.DetailedClub.model_validate(
+            {**raw, **{"bound_client": self}}
+        )
 
     def get_club_members(
         self, club_id: int, limit: int | None = None
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -565,7 +565,7 @@ class Client:
 
     def get_club_activities(
         self, club_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.Activity]:
+    ) -> BatchedResultsIterator[model.ClubActivity]:
         """Gets the activities associated with specified club.
 
         https://developers.strava.com/docs/reference/#api-Clubs-getClubActivitiesById
@@ -588,7 +588,7 @@ class Client:
         )
 
         return BatchedResultsIterator(
-            entity=model.Activity,
+            entity=model.ClubActivity,
             bind_client=self,
             result_fetcher=result_fetcher,
             limit=limit,
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -51,6 +51,7 @@ from stravalib.strava_model import (
     PolylineMap,
     Primary,
     SportType,
+    SummaryAthlete,
     SummaryGear,
     TimedZoneRange,
 )
@@ -421,11 +422,7 @@ class MetaAthlete(strava_model.MetaAthlete, BoundClientEntity):
     pass
 
 
-class SummaryAthlete(MetaAthlete, strava_model.SummaryAthlete):
-    pass
-
-
-class Athlete(SummaryAthlete, strava_model.DetailedAthlete):
+class Athlete(strava_model.DetailedAthlete):
     """Represents high level athlete information including
     their name, email, clubs they belong to, bikes, shoes, etc.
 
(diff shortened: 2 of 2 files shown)


src/stravalib/client.py
```diff
@@ -468,7 +468,7 @@ def get_athlete_clubs(self) -> list[model.Club]:
         # most 30 clubs are returned!
         club_structs = self.protocol.get("/athlete/clubs")
         return [
-            model.Club.model_validate({**raw, **{"bound_client": self}})
+            model.SummaryClub.model_validate({**raw, **{"bound_client": self}})
```
lwasser, line 471: Club doesn't exist - rather we have MetaClub, SummaryClub, Detailed Club.

src/stravalib/model.py
```diff
@@ -333,6 +330,14 @@ class Club(
...
     description: Optional[str] = None
     club_type: Optional[str] = None
+    activity_types_icon: Optional[str] = None
+    dimensions: Optional[list[str]] = None
+    localized_sport_type: Optional[str] = None
+    # These items BELOW should come from detailed club but
+    # they are returned in the clubs/{id} endpoint. why?
+    # membership: Optional[bool] = None
+    # admin: Optional[bool] = None
+    # owner: Optional[bool] = None
```
lwasser, line 340:
```suggestion
```

lwasser:
Point of failure is still mypy. but otherwise this runs and aligns with returns via postman.

src/stravalib/client.py
```diff
@@ -559,7 +561,7 @@ def get_club_members(
 
     def get_club_activities(
         self, club_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.Activity]:
+    ) -> BatchedResultsIterator[model.ClubActivity]:
```
lwasser, line 564: This is a new(ish) to the spec class that we hadn't brought into our model module yet.

src/stravalib/model.py
```diff
@@ -880,6 +884,7 @@ class SummaryActivity(MetaActivity, strava_model.SummaryActivity):
     sport_type: Optional[RelaxedSportType] = None
 
 
+# TODO: should this be named Detailed Activity?
```
lwasser, line 887: i suspect this will be addressed in a future pr. Activities hasn't been done yet.

src/stravalib/model.py
```diff
@@ -936,6 +941,33 @@ class Activity(
     _naive_local = field_validator("start_date_local")(naive_datetime)
 
 
+class ClubAthlete(strava_model.MetaAthlete):
```
lwasser, line 944: Ok @jsamoocha i went a little off plan here because i wasn't sure how to best do this.

the spec says that club activities should contain a metaAthlete object. But metaAthlete only has the id field but the spec says this should return first name and last initial. 

so to follow and allow the user to get the same thing that the actual API returns, i created this club athlete object. if this already exists or is documented somewhere i must be missing it. this allows for us to return the data exactly as strava returns it.

lwasser:
ok i'm done with all of the clubs/* endpoints. some had already been done so this just captures /clubs/{id} and  /clubs/{id}/activities

src/stravalib/model.py
jsamoocha: Activities were done already. I just hesitated to rename this class to DetailedActivity. But since you started doing this for clubs, I think it's a good idea to be consistent with the Strava API docs. I'll deal with that in a final cleanup PR.

src/stravalib/model.py
jsamoocha: Have you considered inheriting (or simply reusing) `strava_model.ClubAthlete`? It has the same fields. This looks like another inconsistency in the Strava API, where `ClubActivity.athlete` should be of type `ClubAthlete` (instead of `MetaAthlete`).

This, however, will trip up mypy because `ClubAthlete` is not a subtype of `MetaAthlete` (so it will require a `# type: ignore[assignment]`).

jsamoocha (commented):
This is good. Can you try to reuse the existing `strava_model.ClubAthlete` when possible?

The two mypy issues will be resolved by merging #517 and fixing #510.

src/stravalib/model.py
lwasser: perfect!

lwasser (commented):
Ok - time for another review. i think i've addressed the issue. This just means we have some None values returned for the clubs/{id}/activities athlete associated with an activity values.

src/stravalib/model.py
```diff
@@ -961,6 +954,25 @@ class Activity(
...
 
+class ClubActivity(strava_model.ClubActivity):
+    """Represents an activity returned from a club.
+
+    Notes
+    -----
+    The actual strava API specification suggests that this should
+    return a MetaAthlete Object for the activities' athlete information.
+    However, while that object should return the correct values, it is missing what is
+     actually returned returns is actually returned is resource_state, first name and
```
jsamoocha, line 965:
```suggestion
     actually returned, i.e., resource_state, first name and
```

jsamoocha (approved):
Great, thanks, another box checked!

lwasser:
sweet. thanks!
merging!! i'll next rebase my other pr and will work on correct return types given they should be available now on the pydantic-v2 branch

src/stravalib/model.py
```diff
@@ -962,7 +962,7 @@ class ClubActivity(strava_model.ClubActivity):
     The actual strava API specification suggests that this should
     return a MetaAthlete Object for the activities' athlete information.
     However, while that object should return the correct values, it is missing what is
-     actually returned returns is actually returned is resource_state, first name and
+     actually returned, i.e., resource_state, first name and
```
lwasser, line 965: yikes. good fix - thanks!

PR #520, opened 2024-06-07
Checks for activities/id endpoint 🙈 

lwasser:
Addresses in part #506

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -565,7 +565,7 @@ class Client:
 
     def get_club_activities(
         self, club_id: int, limit: int | None = None
-    ) -> BatchedResultsIterator[model.ClubActivity]:
+    ) -> BatchedResultsIterator[model.DetailedActivity]:
         """Gets the activities associated with specified club.
 
         https://developers.strava.com/docs/reference/#api-Clubs-getClubActivitiesById
@@ -580,7 +580,7 @@ class Client:
         Returns
         -------
         class:`BatchedResultsIterator`
-            An iterator of :class:`stravalib.model.Activity` objects.
+            An iterator of :class:`stravalib.model.DetailedActivity` objects.
 
         """
         result_fetcher = functools.partial(
@@ -677,10 +677,10 @@ class Client:
             Deprecated. Prefer to use sport_type. In a request where both type
             and sport_type are present, this field will be ignored.
             See https://developers.strava.com/docs/reference/#api-models-UpdatableActivity.
-            For possible values see: :class:`stravalib.model.Activity.TYPES`
+            For possible values see: :class:`stravalib.model.DetailedActivity.TYPES`
         sport_type : str, default=None
             For possible values (case-sensitive) see:
-            :class:`stravalib.model.Activity.SPORT_TYPES`
+            :class:`stravalib.model.DetailedActivity.SPORT_TYPES`
 
         Returns
         -------
@@ -696,9 +696,9 @@ class Client:
         # Check for sport_type first. If provided, it is favored over
         # activity_type / type
         if sport_type is not None:
-            if not sport_type in model.Activity.SPORT_TYPES:
+            if not sport_type in model.DetailedActivity.SPORT_TYPES:
                 raise ValueError(
-                    f"Invalid activity type: {sport_type}. Possible values: {model.Activity.SPORT_TYPES!r}"
+                    f"Invalid activity type: {sport_type}. Possible values: {model.DetailedActivity.SPORT_TYPES!r}"
                 )
             params["sport_type"] = sport_type
             return params
@@ -708,10 +708,10 @@ class Client:
         # sport_type.
         elif activity_type is not None:
             if not activity_type.lower() in [
-                t.lower() for t in model.Activity.TYPES
+                t.lower() for t in model.DetailedActivity.TYPES
             ]:
                 raise ValueError(
-                    f"Invalid activity type: {activity_type}. Possible values: {model.Activity.TYPES!r}"
+                    f"Invalid activity type: {activity_type}. Possible values: {model.DetailedActivity.TYPES!r}"
                 )
             params["type"] = activity_type.lower()
             warn_param_deprecation(
@@ -733,7 +733,7 @@ class Client:
         activity_type: ActivityType | None = None,
         description: str | None = None,
         distance: pint.Quantity | float | None = None,
-    ) -> model.Activity:
+    ) -> model.DetailedActivity:
         """Create a new manual activity.
 
         If you would like to create an activity from an uploaded GPS file, see the
@@ -748,10 +748,10 @@ class Client:
             Deprecated. Prefer to use sport_type. In a request where both type
             and sport_type are present, this field will be ignored.
             See https://developers.strava.com/docs/reference/#api-models-UpdatableActivity.
-            For possible values see: :class:`stravalib.model.Activity.TYPES`
+            For possible values see: :class:`stravalib.model.DetailedActivity.TYPES`
         sport_type : str, default=None
             For possible values (case-sensitive) see: For possible values
-            see: :class:`stravalib.model.Activity.SPORT_TYPES`
+            see: :class:`stravalib.model.DetailedActivity.SPORT_TYPES`
         start_date_local : class:`datetime.datetime` or string in ISO8601 format
             Local date/time of activity start. (TZ info will be ignored)
         elapsed_time : class:`datetime.timedelta` or int (seconds)
@@ -797,7 +797,7 @@ class Client:
 
         raw_activity = self.protocol.post("/activities", **params)
 
-        return model.Activity.model_validate(
+        return model.DetailedActivity.model_validate(
             {**raw_activity, **{"bound_client": self}}
         )
 
@@ -814,7 +814,7 @@ class Client:
         gear_id: int | None = None,
         device_name: str | None = None,
         hide_from_home: bool | None = None,
-    ) -> model.Activity:
+    ) -> model.DetailedActivity:
         """Updates the properties of a specific activity.
 
         https://developers.strava.com/docs/reference/#api-Activities-updateActivityById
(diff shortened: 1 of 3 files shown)


src/stravalib/model.py
```diff
@@ -410,7 +410,8 @@ class AthleteStats(strava_model.ActivityStats):
 
 
 class MetaAthlete(strava_model.MetaAthlete, BoundClientEntity):
-    pass
+    # Undocumented
+    resource_state: Optional[int]
```
lwasser, line 414: does anyone every use / need this? i see it throughout . i'm guessing the answer is no?

src/stravalib/model.py
```diff
@@ -895,7 +896,53 @@ class SummaryActivity(MetaActivity, strava_model.SummaryActivity):
...
+class Trend(BaseModel):
+    speeds: Optional[list[float]] = None
+    current_activity_index: Optional[int] = None
+    min_speed: Optional[float] = None
+    mid_speed: Optional[float] = None
+    max_speed: Optional[float] = None
+    direction: Optional[int] = None
+
+
+class SimilarActivities(BaseModel):
```
lwasser, line 908: undocumented

src/stravalib/model.py
```diff
@@ -895,7 +896,53 @@ class SummaryActivity(MetaActivity, strava_model.SummaryActivity):
     )(check_valid_location)
 
 
-class Activity(
+class Trend(BaseModel):
+    speeds: Optional[list[float]] = None
```
lwasser, line 900: undocumented

src/stravalib/model.py
```diff
@@ -904,13 +951,18 @@ class Activity(
     """
 
     # field overrides from superclass for type extensions:
```
lwasser, line 953: i didn't check any of these objects - so i left TODO's as a reminder that we should check them.

src/stravalib/model.py
```diff
@@ -944,6 +998,14 @@ class Activity(
     from_accepted_tag: Optional[bool] = None
     segment_leaderboard_opt_out: Optional[bool] = None
     perceived_exertion: Optional[int] = None
+    prefer_perceived_exertion: Optional[bool] = None
+    visibility: Optional[str] = None
+    # TODO: add stats_visibilty - object type?
+    # stats_visibility
+    # TODO: similar_activities - new type?
```
lwasser, line 1005:
```suggestion
```
```suggestion
    # TODO: similar_activities - new type?
```
```suggestion
    # TODO: similar_activities - new type?
```

lwasser:
Ok - this is also ready for review. there is a part 2 - to this pr where we take a handful of missing and renamed objects related to gear, etc and fix them as well. i'm hesitant to do that in this pr - rather i'd prefer a solid review, we can evaluate the need for additional tests and merge and then work on the other new objects separately. some of them are related to segment returns, gear, athlete etc.

lwasser:
mypy is still unhappy however i'm not sure given the other issues we have open if that is critical to fix - NOW vs addressing the rest of the endpoints and issues that might resolve some of the type issues naturally. but please say the word if we should focus on mypy passing now - i will just need a bit of guidance if that is the case.

jsamoocha:
We first need to merge #517 and #519 and rebase this branch on those changes. It's good to rename the existing `model.Activity` to `DetailedActivity` for consistency with the Strava API. However, we cannot simply change all the existing references from `model.Activity` to `model.DetailedActivity`. Some of those should be `ClubActivity`, `SummaryActivity`, or `MetaActivity`.

lwasser:
this makes sense @jsamoocha i'm going to turn this into a draft again. i'll get the other pr #519 cleaned up for another review. and we can go from there!

src/stravalib/client.py
```diff
@@ -731,7 +735,7 @@ def create_activity(
         activity_type: ActivityType | None = None,
         description: str | None = None,
         distance: pint.Quantity | float | None = None,
-    ) -> model.Activity:
+    ) -> model.DetailedActivity:
```
lwasser, line 738: it's unlikely that create_activity will need Detailed level of attrs... double check.

src/stravalib/client.py
```diff
@@ -812,7 +820,7 @@ def update_activity(
         gear_id: int | None = None,
         device_name: str | None = None,
         hide_from_home: bool | None = None,
-    ) -> model.Activity:
+    ) -> model.DetailedActivity:
```
lwasser, line 823: Similarly double check that this is indeed detailed level vs summary.

src/stravalib/client.py
```diff
@@ -783,22 +797,29 @@ def create_activity(
...
         )
 
         raw_activity = self.protocol.post("/activities", **params)
 
-        return model.Activity.model_validate(
+        return model.DetailedActivity.model_validate(
             {**raw_activity, **{"bound_client": self}}
         )
 
+    # TODO: response is detailed activity, param sent should
```
lwasser, line 821: @jsamoocha i'm much less family with post . The spec says the object being "put" is UpdateableActivity. I am not sure what this would look like in our code - we are sending a dict / json to the API. Where would that object fit in? would we want to use it and serialize vs creating a params dict and sending directly?

ANd is there any benefit in this vs just leaving this all as is given it works and it's dict --> JSON when it is sent to the rest endpoint. 

i hope that makes sense.

src/stravalib/model.py
```diff
@@ -336,33 +336,38 @@ class SummaryClub(MetaClub, strava_model.SummaryClub):
...
-        Returns
-        -------
-        Iterator
-            An iterator of Activity objects representing club activities.
-        """
-        assert self.bound_client is not None, "Bound client is not set."
-        return self.bound_client.get_club_activities(self.id)
-
-
+    # TODO: I don't understand where these lazy properties are called.
```
lwasser, line 339: @jsamoocha i'm not sure we need these lazy properties any more. The endpoints are separate and we are handling lazy loading with the batched iterator.  Please say the word if i'm missing something!

src/stravalib/model.py
lwasser: i think i'd prefer to check the test of this in a separate PR if you are ok with that. this pr feels bigger than i wanted it to be.

lwasser (commented):
@jsamoocha ready for another review. there are a few items here that i am not confident about . thanks!!

src/stravalib/model.py
```diff
@@ -336,33 +336,8 @@ class SummaryClub(MetaClub, strava_model.SummaryClub):
     dimensions: Optional[list[str]] = None
     localized_sport_type: Optional[str] = None
 
-    @lazy_property
```
lwasser, line 339: I can put this back if i'm missing something. i don't see how this is used given members and activities are separate endpoints

e.g.: 

clubs/{id}/activities

jsamoocha (changes requested):
I'm in favor of retaining the lazy properties as a convenience to users and preventing more breaking changes.

About the added undocumented attributes, please decide which are useful enough to add.

Finally, I addressed your comment about the naming of `ActivityLap`, and corrected the type of `Lap.activity`.

Almost there 😅

src/stravalib/client.py
lwasser: ok perfect.

src/stravalib/model.py
lwasser: ahhhh ok @jsamoocha this was a pure miss on my part. i should have looked more closely at the code and what it was doing. i just saw that it was a separate endpoint + the spec and jumped to a conclusion that it didn't belong there. 

that is a really nice feature to have. i've added it back in and it does appear to be working.

src/stravalib/model.py
```diff
@@ -351,33 +351,8 @@ class SummaryClub(MetaClub, strava_model.SummaryClub):
     dimensions: Optional[list[str]] = None
```
lwasser: Related to your other comment, these three attrs that i've added are odd and likely not sure useful to users. I have removed them but pls say the word if you think we should add them given i added them and it's just more for us to maintain.

Screenshot 2024-06-12 at 12 22 25 AM

src/stravalib/model.py
```diff
@@ -950,6 +980,13 @@ class Activity(
     from_accepted_tag: Optional[bool] = None
     segment_leaderboard_opt_out: Optional[bool] = None
     perceived_exertion: Optional[int] = None
+    prefer_perceived_exertion: Optional[bool] = None
+    visibility: Optional[str] = None
+    # TODO: add stats_visibilty - object type?
+    # stats_visibility
+    similar_activities: Optional[SimilarActivities] = None
+    available_zones: Optional[list[str]] = None
+    private_note: Optional[str] = None
```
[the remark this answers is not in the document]
lwasser: ok cool. i've removed most of them. i left private note as i believe that is the text that you can write in the notes on an activity that is not visible to users. a user might like to retrieve that. 

but the other stuff is mostly settings regarding what strava can collect (heart rate data etc and what is visible).. mostly related to privacy settings / health data i believe.

src/stravalib/model.py
lwasser: removed this as well! that point is valid. no need to add extra undocumented attrs if users don't need them.

lwasser:
> Finally, I addressed your comment about the naming of `ActivityLap`, and corrected the type of `Lap.activity`.
> 
> Almost there 😅

thank you @jsamoocha so... before i started working on this again i rebased and force pushed 🙈 I think you added some commits to fix Lap.activity. and i am not sure that i can retrieve those after a force push. can you either re-add what you implemented or tell me what you wrote / did?

my apologies!! i should have `git pulled` first before rebasing, it's just habit to rebase and start working again after a pr has been merged. i don't think i can use reflog in this case because of the order of how i did things. 

i've addressed everything else. (i think?!)

lwasser:
Ok i was able to retrieve the commits visually using email links. i tried to cherry pick the commit itself as i could see it from my email link on github but i've already overwritten it with the force push so git didn't like be trying to cherry pick. i actually don't know how that works and it's hurting my brain to think about so i've just added the changes myself!  🤯 

hopefully this addresses all concerns. this change really just removes some of the stuff i added that was extraneous and puts back those (very cool to have) lazy properties that i didn't understand before.

jsamoocha (approved):
Great, thanks, let's merge this!

PR #524, opened 2024-06-14
Update: Activities/{id}/zones  & Laps

lwasser:
this addresses the activities/{id}/zones endpoint (and will impact the lazy loaded option in /activities/{id}

addresses in part #506

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -1026,9 +1026,7 @@ class Client:
 
         return ActivityUploader(self, response=initial_response)
 
-    def get_activity_zones(
-        self, activity_id: int
-    ) -> list[model.BaseActivityZone]:
+    def get_activity_zones(self, activity_id: int) -> list[model.ActivityZone]:
         """Gets activity zones for activity.
 
         Activity zones relate to a users effort (heartrate and power).
@@ -1043,7 +1041,7 @@ class Client:
         Returns
         -------
         :class:`list`
-            A list of :class:`stravalib.model.BaseActivityZone` objects.
+            A list of :class:`stravalib.model.ActivityZone` objects.
 
         Notes
         -----
@@ -1055,9 +1053,7 @@ class Client:
         zones = self.protocol.get("/activities/{id}/zones", id=activity_id)
 
         return [
-            model.BaseActivityZone.model_validate(
-                {**z, **{"bound_client": self}}
-            )
+            model.ActivityZone.model_validate({**z, **{"bound_client": self}})
             for z in zones
         ]
 
@@ -1195,7 +1191,7 @@ class Client:
         Returns
         -------
         class:`BatchedResultsIterator`
-            An iterator of :class:`stravalib.model.ActivityLaps` objects.
+            An iterator of :class:`stravalib.model.Lap` objects.
 
         """
         result_fetcher = functools.partial(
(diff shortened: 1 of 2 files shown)


src/stravalib/client.py
```diff
@@ -1043,7 +1041,7 @@ def get_activity_zones(
         Returns
         -------
         :class:`list`
-            A list of :class:`stravalib.model.BaseActivityZone` objects.
```
lwasser, line 1046: A rename to align with the spec

src/stravalib/model.py
```diff
@@ -993,39 +991,44 @@ class ClubActivity(strava_model.ClubActivity):
...
 ):
     """
     Base class for activity zones.
 
-    A collection of :class:`stravalib.model.DistributionBucket` objects.
+    A collection of :class:`stravalib.strava_model.TimedZoneDistribution` objects.
     """
 
     # Field overrides from superclass for type extensions:
+    # TODO: what was the legacy behavior and is it ok to break it?
```
lwasser, line 1024: @jsamoocha this is a sanity check. i will remove all comments once i'm clear this is correct. 

Is there legacy behavior that I need to consider here or can we just map to the API spec / returned values?

src/stravalib/model.py
```diff
@@ -993,39 +991,44 @@ class ClubActivity(strava_model.ClubActivity):
...
     # Using type that is currently mimicking legacy behavior.
-    distribution_buckets: Optional[list[DistributionBucket]] = None  # type: ignore[assignment]
 
-    # TODO: ignoring type given legacy support will be deprecated
+    # This should return a list of TimedZoneDistribution objects
+    # But this returns an error
```
lwasser, line 1028:
```suggestion
```

lwasser:
ok ready for review! ✨ i'll clean up comments once the review is complete // before we merge !

src/stravalib/model.py
```diff
@@ -652,10 +650,8 @@ class Lap(
     strava_model.Lap,
     BoundClientEntity,
 ):
-    # Field overrides from superclass for type extensions:
-    activity: Optional[MetaActivity] = None
-    athlete: Optional[Athlete] = None
 
+    # TODO: double check undocumented...
```
lwasser, line 654:
```suggestion
```

src/stravalib/model.py
```diff
@@ -993,39 +987,43 @@ class ClubActivity(strava_model.ClubActivity):
     pass
 
 
-class DistributionBucket(TimedZoneRange):
+# Should be TimedZoneDistribution
```
lwasser, line 990:
```suggestion
```

src/stravalib/model.py
```diff
@@ -652,9 +650,6 @@ class Lap(
     strava_model.Lap,
     BoundClientEntity,
 ):
-    # Field overrides from superclass for type extensions:
-    activity: Optional[MetaActivity] = None
-    athlete: Optional[Athlete] = None
```
jsamoocha, line 657: These should remain here, as they're _our_ `MetaActivity` and `MetaAthlete` types (which inherit from `BoundClientEntity`)

src/stravalib/model.py
jsamoocha: We override the `min` and `max` attributes to also allow floats as their type. You already took care of that.

jsamoocha (approved):
This looks good to merge, thanks!

PR #530, opened 2024-06-17
Fix: Migrate to covariant types to fix typing (in some places)

### LINKED ISSUE #510: Stravalib 2.0 -> Replace invariant by covariant container types

lwasser:
closes #510 

Note: i'm a bit unsure of how to handle docstrings in this case. I worry that some users will be confused by what a Sequence is. We are still technically passing lists and dictionaries so i'd love input on the best way to document this properly! i have only begun to migrate the list objects so far. 

## Description

This is a start to the migration to more flexible object typing that will support the overrides that we have to implement to support actual strava responses. 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -11,7 +11,7 @@ import collections
 import functools
 import logging
 import time
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable
 from datetime import datetime, timedelta
 from io import BytesIO
 from typing import (
@@ -1028,9 +1028,7 @@ class Client:
 
         return ActivityUploader(self, response=initial_response)
 
-    def get_activity_zones(
-        self, activity_id: int
-    ) -> Sequence[model.ActivityZone]:
+    def get_activity_zones(self, activity_id: int) -> list[model.ActivityZone]:
         """Gets activity zones for activity.
 
         Activity zones relate to a users effort (heartrate and power).
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -11,7 +11,7 @@ import collections
 import functools
 import logging
 import time
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from datetime import datetime, timedelta
 from io import BytesIO
 from typing import (
@@ -1028,7 +1028,9 @@ class Client:
 
         return ActivityUploader(self, response=initial_response)
 
-    def get_activity_zones(self, activity_id: int) -> list[model.ActivityZone]:
+    def get_activity_zones(
+        self, activity_id: int
+    ) -> Sequence[model.ActivityZone]:
         """Gets activity zones for activity.
 
         Activity zones relate to a users effort (heartrate and power).
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -178,7 +178,7 @@ def lazy_property(fn: Callable[[U], T]) -> Optional[T]:
 # This method checks a list of floats - ie a stream not just a single lat/lon
 def check_valid_location(
     location: Optional[Union[Sequence[float], str]]
-) -> Optional[Sequence[float]]:
+) -> Optional[list[float]]:
     """
     Validate a list of location xy values.
 
@@ -211,10 +211,13 @@ def check_valid_location(
         try:
             return [float(l) for l in location.split(",")]
         except AttributeError:
-            # location for activities without GPS may be returned as empty list
+            # Location for activities without GPS may be returned as empty list
             return None
+    # Because this could be any Sequence type, explicitly return list
+    elif location:
+        return list(location)
     else:
-        return location if location else None
+        return None
 
 
 class BoundClientEntity(BaseModel):
@@ -286,9 +289,7 @@ class LatLon(LatLng):
     """
 
     @model_validator(mode="before")
-    def check_valid_latlng(
-        cls, values: Sequence[float]
-    ) -> Sequence[float] | None:
+    def check_valid_latlng(cls, values: Sequence[float]) -> list[float] | None:
         """Validate that Strava returned an actual lat/lon rather than an empty
         list. If list is empty, return None
 
@@ -306,7 +307,11 @@ class LatLon(LatLng):
         """
 
         # Strava sometimes returns empty list in case of activities without GPS
-        return values if values else None
+        # Explicitly return a list to make mypy happy
+        if values:
+            return list(values)
+        else:
+            return None
 
     @property
     def lat(self) -> float:
@@ -740,9 +745,7 @@ class SummarySegment(strava_model.SummarySegment, BoundClientEntity):
     start_latlng: Optional[LatLon] = None
     end_latlng: Optional[LatLon] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
-    # TODO: mypy error - SummarySegment defines as  literal with two values
-    # activity_type: Optional[Literal["Ride", "Run"]] = None
-    # We could force ignore via cast??
+    # Ignore because the spec is incorrectly typed - Optional[Literal["Ride", "Run"]]
     activity_type: Optional[RelaxedActivityType] = None  # type: ignore[assignment]
     athlete_segment_stats: Optional[AthleteSegmentStats] = (
         None  # Actually, this is only part of a (detailed) segment response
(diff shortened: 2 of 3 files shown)


src/stravalib/model.py
```diff
@@ -187,14 +188,14 @@ def check_valid_location(
 
     Parameters
     ----------
-    location : list of floats
-        Either a list of x,y floating point values or strings or None
+    location : Sequence of floats
```
lwasser, line 191: so how do we update the docstrings? can we leave them as is given this is just a typing update rather than an actual structural update?

jsamoocha: Good question. If we want to be 100% _correct_ in our documentation, I'd go for Sequence (as we technically cannot guarantee it will be lists (vs., e.g., tuples) - that depends on how Pydantic parses the raw data). However, it looks like Pydantic currently parses json arrays into a list when the field type is a Sequence. So, the pragmatic approach would be to keep the docstring type as list, and only update the docstrings whenever the Pydantic implementation changes (which may never happen).

jsamoocha:
Overall, this looks good. Try to remove all the `# type: ignore[assignment]` comments on the Sequence fields; they should be okay now (except maybe for our `Relaxed[Activity,Sport]Type` since these are not proper subclasses of, e.g., `ActivityType`).

src/stravalib/model.py
```diff
@@ -738,7 +740,10 @@ class SummarySegment(strava_model.SummarySegment, BoundClientEntity):
     start_latlng: Optional[LatLon] = None
     end_latlng: Optional[LatLon] = None
     athlete_pr_effort: Optional[AthletePrEffort] = None
-    activity_type: Optional[RelaxedActivityType] = None  # type: ignore[assignment]
+    # TODO: mypy error - SummarySegment defines as  literal with two values
+    # activity_type: Optional[Literal["Ride", "Run"]] = None
+    # We could force ignore via cast??
+    activity_type: Optional[RelaxedActivityType] = None
```
lwasser, line 746: Mypy is failing here - 

```
src/stravalib/model.py:746: error: Incompatible types in assignment (expression has type "Optional[RelaxedActivityType]", base class "SummarySegment" defined the type as "Optional[Literal['Ride', 'Run']]")  [assignment]
```
i understand why it's failing but i don't understand the best way to address it. Do we need to use cast somewhere? or how do we tell mypy this typing mismatch is in fact ok?

src/stravalib/model.py
```diff
@@ -972,7 +977,7 @@ class ClubActivity(strava_model.ClubActivity):
     actual return.
     """
 
-    # Spec calls for a return of a metaAthlete object (which only has id in it)
+    # Class override - spec returns metaAthlete object (which only has id in it)
```
lwasser, line 980: similarly mypy fails here. we are overriding here again for good reason. i'm not sure how to handle this. 

```
src/stravalib/model.py:981: error: Incompatible types in assignment (expression has type "Optional[ClubAthlete]", base class "ClubActivity" defined the type as "Optional[MetaAthlete]")  [assignment]
```

lwasser:
> # type: ignore[assignment]

this is great - i'm working on this now - i have two questions ... i'll post them in just a moment

lwasser (commented):
ok i have made some more updates - there are two typing issues that id on't know how to handle using #ignore doesn't seem to work as i expect it to. but they are valid overrides.

then i am seeing two other mypy issues related to unithelper that i can't seem to recreate locally which is odd. 

but this is ready for review.

```
src/stravalib/unithelper.py:28: error: Unused "type: ignore[misc, valid-type]" comment  [unused-ignore]
src/stravalib/unithelper.py:80: error: Statement is unreachable  [unreachable]
```

src/stravalib/model.py
```diff
@@ -991,8 +996,8 @@ class TimedZoneDistribution(strava_model.TimedZoneRange):
     """
 
     # Type overrides to support pace values returned as ints
-    min: Optional[int | float] = None  # type: ignore
-    max: Optional[int | float] = None  # type: ignore
+    min: Optional[Sequence[int | float]] = None
```
lwasser, line 999: I tried to use sequence here as well.... but i am not sure that Sequence supports the types within the  list / tuple / etc. 

Options:
* either we keep the ignore statements 
* there is some other way to do this correctly?

lwasser:
i will work on the broken tests next - but please note my questions around types that we are overriding that are within the Sequence object.

lwasser:
ok 

✅ tests are passing -
✅  mypy is now passing locally for me. I did ignore assignments for a few intentional overrides that were mostly not Sequence related. Please let me know if that was not appropriate. 
✅ Question: why are we getting typing errors now in unithelper (i have not touched that module) [if i remove](https://github.com/stravalib/stravalib/actions/runs/9572955973/job/26393414373?pr=530#step:8:17) the ignore we get other typing issues. But the [unreachable error](https://github.com/stravalib/stravalib/actions/runs/9572955973/job/26393414373?pr=530#step:8:17) that one is new to me. is it mypy not being able to figure out what type is correct?

`src/stravalib/unithelper.py:80: error: Statement is unreachable  [unreachable]`

jsamoocha (changes requested):
Thanks, this looks good! I suggested a few tiny changes. If you agree with these, you can immediately merge after adding them.

jsamoocha:
> ✅ Question: why are we getting typing errors now in unithelper (i have not touched that module) [if i remove](https://github.com/stravalib/stravalib/actions/runs/9572955973/job/26393414373?pr=530#step:8:17) the ignore we get other typing issues. But the [unreachable error](https://github.com/stravalib/stravalib/actions/runs/9572955973/job/26393414373?pr=530#step:8:17) that one is new to me. is it mypy not being able to figure out what type is correct?

The unithelper issues will be resolved by fixing #500.

src/stravalib/client.py
```diff
@@ -1028,7 +1028,9 @@ def upload_activity(
 
         return ActivityUploader(self, response=initial_response)
 
-    def get_activity_zones(self, activity_id: int) -> list[model.ActivityZone]:
+    def get_activity_zones(
+        self, activity_id: int
+    ) -> Sequence[model.ActivityZone]:
```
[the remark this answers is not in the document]
lwasser: Ok i think i understand this now. We want to allow for input variability (any "sequence" object list, tuple etc) but we also control the output and can define it in the method itself. thank you. i think i may need to fix this in a few spots so i'll go through and check for that now.

src/stravalib/model.py
```diff
@@ -176,8 +177,8 @@ def wrapper(obj: U) -> Optional[T]:
 
 # This method checks a list of floats - ie a stream not just a single lat/lon
 def check_valid_location(
-    location: Optional[Union[list[float], str]]
-) -> Optional[list[float]]:
+    location: Optional[Union[Sequence[float], str]]
+) -> Optional[Sequence[float]]:
```
lwasser, line 181: sam here - returns a list but can ingest a Sequence

src/stravalib/model.py
lwasser: perfect. i've left the note just so we (i) remember but cleaned it up a bit.

src/stravalib/model.py
```diff
@@ -991,8 +997,8 @@ class TimedZoneDistribution(strava_model.TimedZoneRange):
     """
 
     # Type overrides to support pace values returned as ints
-    min: Optional[int | float] = None  # type: ignore
-    max: Optional[int | float] = None  # type: ignore
+    min: Optional[int | float] = None  # type: ignore[assignment]
+    max: Optional[int | float] = None  # type: ignore[assignment]
```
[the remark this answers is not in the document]
lwasser: awesome! i didn't realize int was a subtype of float! i've now typed it as: 

`float | None`

lwasser (commented):
Ok - fixed all of the items highlighted in the comments. 

NOTE: we use Optional throughout both of these modules. Perhaps in a separate PR i'll go through and remove those now that we've dropped python 3.9.  I'll open an issue about this.

lwasser:
> Thanks, this looks good! I suggested a few tiny changes. If you agree with these, you can immediately merge after adding them.

awesome. Thanks @jsamoocha i'll go ahead and merge this!! just a note i made a few tiny changes to the code to appease mypy. it just needed to ensure a more explicit return  of a list object.

example: in check valid location
i did this:
```python
    # Because this could be any Sequence type, explicitly return list
    elif location:
        return list(location)
    else:
        return None
```

replacing this:

```python
else:
        return location if location else None
```

PR #532, opened 2024-06-22
Fix: remove Optional throughout and replace with |

lwasser:
I haven't opened an issue 

This is a tiny code change to remove `Optional` from our typing and replace it with `|` given we did drop python 3.9 support. 

if we are worried about this breaking builds for anyone using 3.9 we could add 

`from typing_extensions import Annotated` for backwards compatibility. but 3.9 will be end of life in the fall so in theory folks should be upgrading to 3.10 at this point. 3.13 is being actively worked on. 

## Type of change

- [ ] Other (please describe) Typing! 

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -13,16 +13,14 @@ related entities from the API.
 from __future__ import annotations
 
 import logging
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from datetime import date, datetime, timedelta
 from functools import wraps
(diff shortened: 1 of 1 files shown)


lwasser:
Note: I used pyupgrade to implement this. it made things super easy and fast to do. literally this took a minute.

jsamoocha (approved):
Great! I'll immediately merge this to rebase a PR I'm working on.

Issue #502, opened 2024-05-11
Stravalib 2.0 -> Documentation Update


jsamoocha:
- [x] Incorporate all (breaking) changes in the documentation
- [x] Filter base type members from new custom types (e.g., `Duration` will, by default, include members of `int`)
- [x] Add examples of using the new custom types in docstrings
- [x] Add examples of using the new custom types in a new documentation section (about unit conversion)

lwasser:
i am happy to help with the docs. i could even start a page now pulling from the hackmd file that has notes in it. Just let me know how i can be helpful in all of this!

jsamoocha:
Thanks! Let's wait with this until some of the other issues are resolved and we have a definite understanding of how our API will look like.

lwasser:
sounds great!

lwasser:
i'm actually revisiting this original issue you opened in May that has the things i want to work on. Do you think we should chip away at the other items first before i dig too much into the docs? am I jumping the gun? i may be able to to do a bit more work this weekend on something .

jsamoocha:
Only a tiny bit of work is left in #501, which does not affect documentation or user-facing API. If you want me to help with specific sections of the documentation or examples, please let me know.

lwasser:
@jsamoocha that would be great. Id LOVE help with 

1. the section on deprecations/ breaking changes in stravalib 2.x. I was thinking i could add a banner to the top of the docs that linked to that page to alert users of the changes! Similar to what pydantic did. it's a built in feature of our theme to do this. 
2. with examples. 

Suggestion: How about i take my current pr  with the inheritance diagrams. (i can do this tonight). i'll split the pr into two:

1. clean the docs / add autoapi as our main autodoc tool + a banner at the top for the stravalib 2.x page.  i can create an empty deprecation page with some notes from our hackmd file. we can quickly merge this pr as it's just a structural update 
2. the inheritance diagrams + filtering of methods from objects like `int`

You can then work on the deprecated items page in a separate pr. and examples as you see fit!
 
i'll work on the inheritance diagrams and customizing new autoapi module which i've been playing with to filter things. Based on my research this extension should do what we want it to BUT I haven't tried that yet to do that specifically! I really like the extension because we literally do nothin to add things to the api docs. 

[demo of the docs here](https://github.com/pyOpenSci/pyosPackage)
[package structure here](https://github.com/pyOpenSci/pyosPackage)

it just creates the api docs on build and adds them to our docs!! 
[...]

PR #534, opened 2024-06-23
Custom types

### LINKED ISSUE #500: Stravalib 2.0 -> Introduce custom Pydantic types for field conversions

jsamoocha:
closes #500 

## Description

Adds custom types to our model that allow to preserve the legacy behavior of, e.g., unit conversion, while being a proper subtype of the generated strava model and not tripping up mypy.

## Type of change

## Does your PR include tests

- [x] Yes


src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+            ),
+        )
+
+    @classmethod
+    def __get_pydantic_json_schema__(
+        cls,
+        _core_schema: core_schema.CoreSchema,
+        handler: GetJsonSchemaHandler,
+    ) -> JsonSchemaValue:
+        return handler(cls.get_core_schema())
```
jsamoocha, line 280: See https://docs.pydantic.dev/latest/concepts/types/#as-an-annotation. This is quite verbose, but I didn't succeed in using the simplified way explained in https://docs.pydantic.dev/latest/concepts/types/#using-getpydanticschema-to-reduce-boilerplate.

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+
+class _CustomStrAnnotation(_CustomTypeAnnotation[str, Any], ABC):
+    @classmethod
+    def get_core_schema(cls) -> Mapping[str, Any]:
+        return core_schema.str_schema()
+
+
+class Duration(int):
+    def timedelta(self) -> timedelta:
+        return timedelta(seconds=self)
```
jsamoocha, line 303: This is an example of the new extended types. We simply subclass `int` (which is the base type of, e.g., `elapsed_time` in many classes) and add an accessor method to the enhanced type.

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+
+    @classmethod
+    def validate(cls, value: float) -> Velocity:
+        return Velocity(value)
+
+
+DurationType = Annotated[Duration, _DurationAnnotation]
+DistanceType = Annotated[Distance, _DistanceAnnotation]
+VelocityType = Annotated[Velocity, _VelocityAnnotation]
+TimezoneType = Annotated[Timezone, _TimezoneAnnotation]
```
jsamoocha, line 373: These are the types that we now use in the model classes.

src/stravalib/model.py
```diff
@@ -398,11 +551,15 @@ class DetailedClub(SummaryClub, strava_model.DetailedClub):
...
 
 
-class ActivityTotals(ActivityTotal):
+class ActivityTotals(strava_model.ActivityTotal):
     """An objecting containing a set of total values for an activity including
     elapsed time, moving time, distance and elevation gain."""
 
-    pass
+    # Attribute overrides for custom types:
+    distance: Optional[DistanceType] = None
```
jsamoocha, line 559: If `a` is an instance of `ActivityTotals`, we can now do the following:

```python
a.distance  # gets distance as a float
a.distance.quantity()  # gets distance as a Pint Quantity (in meters)
unit_helper.feet(a.distance)  # gets distance as a Pint Quantity (in feet)
```

src/stravalib/unithelper.py
```diff
@@ -4,84 +4,43 @@
...
-    @property
-    def unit(self) -> str:
-        warn_units_deprecated()
-        return str(self.units)
-
-    def __int__(self) -> int:
-        return int(self.magnitude)
-
-    def __float__(self) -> float:
-        return float(self.magnitude)
```
jsamoocha, line 48: We drop this backward compatibility with the `units` package we used before.

src/stravalib/unithelper.py
```diff
@@ -4,84 +4,43 @@
...
-
-
-def is_quantity_type(obj: Any) -> bool:
-    if isinstance(obj, (pint.Quantity, Quantity)):
-        return True
-    elif isinstance(obj, UnitsQuantity):  # check using Duck Typing
-        warn_units_deprecated()
-        return True
-    else:
-        return False
```
jsamoocha, line 84: This is not needed anymore because we drop backward compatibility with the `units` package.

jsamoocha:
Codecov complains about dropped coverage in model.py, which I couldn't find. Maybe we can review that before mering `pydantic-v2` into `main`.

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+class _TimezoneAnnotation(_CustomStrAnnotation):
+    _type = Timezone
+
+    @classmethod
+    def validate(cls, value: str) -> Timezone:
+        return Timezone(value)
+
+
+class Distance(_Quantity):
+    unit = "meters"
```
lwasser, line 347:
```suggestion
class Distance(_Quantity):
    """A class that supports conversion of units to meters. This output can then be 
    converted to other units such as feet or meters using the stravalib.unit_helper module. 
    
    Examples
    ----------
    Once you have distance in meters, you can then use the unit_helper module to 
    convert to other units such as feet 
    
    >>> from stravalib import unithelper
    >>> activity = client.get_activity(11416949675)
    >>> activity.distance.quantity()
    <Quantity(8055.9, 'meter')>
    >>>  unithelper.feet(activity.distance)
    <Quantity(26430.1181, 'foot')>
    >>> unithelper.miles(activity.distance)
    <Quantity(5.00570419, 'mile')>
    """
    unit = "meters"
```

docs/reference/unithelper.rst
lwasser, line 1: let's rename unit-helper for consistency with other .rst files. i can also help with this

lwasser (commented):
This is awesome. 
Jonatan, I think there is an opportunity for us to document this piece of stravalib better! It's really really helpful to have these built-in conversions. Reading the docs, I am not sure all users will know about this feature! And now we are competing in our api docs with things like the docs for int (that has examples). this is not a huge deal but i just think it will help users to throw in examples throughout so they know about this functionality! 

This could be a combination of adding some examples to the docstrings + a more descriptive narrative.  
Also -  we could do that outside of this pr as well. i just added a few suggestions here and there so you could see what i'm thinking about!

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+class _DurationAnnotation(_CustomIntAnnotation):
+    _type = Duration
+
+    @classmethod
+    def validate(cls, value: int) -> Duration:
+        return Duration(value)
+
+
+class Timezone(str):
+    def timezone(
```
lwasser, line 315: Can you help me understand how this would be used? when i pull down an activity i'd get a start and end time that are in utc by default. it looks like this is checking a string. is it for updating activities or adding activities? Or could i use it to convert the start time i have from UTC --> Mountain Time ? 

An example would be really helpful!

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+
+class _DistanceAnnotation(_CustomFloatAnnotation):
+    _type = Distance
+
+    @classmethod
+    def validate(cls, value: float) -> Distance:
+        return Distance(value)
+
+
+class Velocity(_Quantity):
```
lwasser, line 358: Let's add a short description here that directs users to unit_helper. And we could also include a small example  using unit_helper. i'm just starting to understand how you pulled this all together - bare with me 😸

lwasser (approved):
@jsamoocha ok i think i generally understand this all now - but my big take aways here are

* let's add some quick examples in the docstrings of the new unit conversion classes
* let's add a section in our docs on unit conversions. 

i can help with this but will need a bit of guidance from you as i'm still trying to wrap my head around how this works for the various raw values that are returned. Once i understand that i can be more helpful with docs! 

i don't have any big changes except it seems like we could potentially rename unithelper.py --> unit_helper.py

src/stravalib/model.py
jsamoocha: We should find some way to filter out the base type inheritance from the documentation for these specific cases. I added a TODO to #502.

src/stravalib/model.py
jsamoocha: That sounds good, I added #535

src/stravalib/model.py
```diff
@@ -220,6 +231,148 @@ def check_valid_location(
...
+        return core_schema.float_schema()
+
+
+class _CustomStrAnnotation(_CustomTypeAnnotation[str, Any], ABC):
+    @classmethod
+    def get_core_schema(cls) -> Mapping[str, Any]:
+        return core_schema.str_schema()
+
+
+class Duration(int):
```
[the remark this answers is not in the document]
jsamoocha: The suggestion is good; I like these examples! Regarding your question: the example should be

```python
activity.elapsed_time.timedelta()
```

as a method call. This will give you the timedelta object (which can be formatted with `str()`).

jsamoocha:
Thanks for the review! Merging this now.

I added TODOs to #502 regarding your documentation suggestions. I added #535 regarding renaming the `unithelper` module.

Issue #528, opened 2024-06-16
Check for non-existing model attributes


jsamoocha:
Some model classes (e.g., `DetailedAthlete`) have several undocumented attributes that aren't returned in Strava's responses. We can write a small test script that uses an access token (*) and compares the raw response from several endpoints with our model classes.

(*) Preferably for both premium and free Strava users.

PR #537, opened 2024-06-29
Docs: Clean up reference docs

lwasser:
This is a start at the documentation cleanup.
NOTE: i do need to go through and remove / update methods / classes etc that we've renamed / updated / removed. I've done some of them but not comprehensively. There are also some docstring typos that i need to clean up and missing docstrings. 

A few notes

* I had originaly wanted to use [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi). This tools is really nice as it automates many parts of api generation and i think would allow us to remove particular classes from the docs via custom templates. However, i was strugglin to get it to play nicely with autodoc_pydantic. 
* So i went back to autodoc but with autodoc if we are automatically generating those api stub files, we can't use the `:exclude-members:` feature because it just creates the .rst stub file on the fly. i'd have to manually create those which defeats the purpose of using autodoc! 

I am going to open an issue in sphinx-autoapi because it's possible i'm missing something with the pydantic piece! 

for the time being, this adds the banner at the top of our docs with a note about and link to the new features page for stravalib 2.0. it's not populated at all yet! 

it also adds structural organization to the model.py api page to make it easier to parse. 

TODOs:

* cleanup / lint / fix docstrings throughout. 
* Double check classes and methods align with what we actually have in our API now. If a class was removed or ranamed sphinx is quiet about not rendering it. 
* there are a handful of errors i need to clean up as well. 

i'm opening this as a draft as there's a bit more work here to do than i expected!


docs/conf.py
```diff
@@ -160,7 +161,7 @@
 # html_file_suffix = None
```
lwasser, line 161: There is a lot of older documentation stuff that i think we can remove from this file. i plan to do that in my next round of work.

docs/conf.py
```diff
@@ -208,3 +209,4 @@
 
 # Default to using the order defined in source.
 autodoc_member_order = "bysource"
+autodoc_default_options = {}
```
lwasser, line 212: This is where i thought maybe we could setup an exclude X class. (specifically for `int`) -- `Duration(int)`

docs/reference/model.rst
```diff
@@ -1,53 +1,138 @@
...
-..    Stream
-..    Route
-..    Subscription
-..    SubscriptionCallback
-..    SubscriptionUpdate
+============
+Model
+============
+
+.. currentmodule:: stravalib.model
```
lwasser, line 5: I understand that this approach means more maintenance. But the docs are infinently more readable if we organize them this way. i'll go back and double check each class / method here (i did this but not for things we've removed - just for new classes. 

so i propose we try this out and if it does end up being too hard to maintain i'll revert to automodule.

lwasser:
here is what the docs currently look like - https://stravalib--537.org.readthedocs.build/en/537/reference/model.html
see how nice model.py renders when we add sections? i can add more descriptive text and diagrams too. 

i really think this will make things more user friendly and am willing to help maintain it until we find it to be too cumbersome. I can't imagine it will change all that often once we get it all correct / checked, etc

src/stravalib/model.py
```diff
@@ -401,6 +402,11 @@ class BoundClientEntity(BaseModel):
 
 
 class RelaxedActivityType(ActivityType):
+    """This object supports allowing an array of Literal values to be used for
```
lwasser, line 405: Any classes that are missing docstrings return a warning when building the docs. i'm going through and adding docstrings as needed. This will NOT fix the recursion issue but will fix the warnings (and is more user friendly). so it's worth doing.

docs/reference/model.rst
```diff
@@ -1,53 +1,123 @@
...
-..    ActivityPhoto
-..    ActivityKudos
-..    ActivityLap
-..    Map
-..    Split
-..    SegmentExplorerResult
-..    AthleteSegmentStats
-..    AthletePrEffort
+..    SummarySegment
 ..    Segment
```
lwasser, line 114: Commented out because i am adding docstrings to each one. Then i'll double check the entire list. The bigger issue here is still that we have a recursion issue that i don't understand. and this happens regardless of using automodule, autosummary etc.

docs/whats-new/changelog.md
```diff
@@ -1,7 +1 @@
 # Changelog: What's new with stravalib
-
-
-:::{include} ./changelog.md
```
lwasser, line 4: i think this include was causing the recursion issue. if that is the case i can work on that later.  it was an issue with myst parses but i don't understand what the issue is -- yet.

docs/conf.py
```diff
@@ -127,85 +128,22 @@
...
-# docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
-# pixels large.
-# html_favicon = None
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["_static"]
 
-# If true, "Created using Sphinx" is shown in the HTML footer. Default is True.
```
lwasser, line 149: ALl of this config is dated. we do all of our config now via pydata_sphinx_theme above via a dict.

docs/conf.py
```diff
@@ -127,85 +129,59 @@
...
+# intersphinx_mapping = {'http://docs.python.org/': None}
 
-# One entry per manual page. List of tuples
-# (source start file, name, description, authors, manual section).
-man_pages = [
-    ("index", "stravalib", "stravalib Documentation", ["Hans Lellelid"], 1)
-]
+# Default to using the order defined in source.
+autodoc_default_options = {
+    "member-order": "alphabetical",
```
lwasser, line 146: sphinx has an exclude_members flag but it seems to only remove generating the stub / text from the members docstring in autosummary. the way to properly remove it is to use the approach below that leverages `auto_skip_members`. there may be a better way but this is the best way i've found so far.

lwasser:
ok @jsamoocha when you have time have a look at this and let me know what you think. 

✅ I setup a custom autoskip method that uses the sphinx autoskip_member. what's weird is when we do that then we have to specify everything that we want to skip. 
✅ I also added / cleaned up docstrings 

▶️  I was thinking then we could take the landing page for api docs -- (reference.rst)
▶️  we could then add inheritance diagrams there to help users understand what that looks like. we could do that in a separate pr. 

Please let me know what you think about this.

src/stravalib/model.py
```diff
@@ -820,6 +906,13 @@ class Lap(
     strava_model.Lap,
     BoundClientEntity,
 ):
+    """An object that represents an Activity lap.
+
+    This object inherits the BoundClient to support API access through lazy
+    methods.
```
jsamoocha, line 912: Seeing this kind of docstring being repeated often, I wonder if we should mention this inheritance of `BoundClient` at all, as it is not really of concern to users of this library. See, e.g., the documentation of [KNeighborsClassifier from scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html). Its docstring doesn't mention that it inherits from `KNeighborsMixin`, `ClassifierMixin`, or `NeighborsBase`.

src/stravalib/model.py
```diff
@@ -1180,19 +1295,17 @@ class ClubActivity(strava_model.ClubActivity):
...
-    The Strava API suggests returning a `MetaAthlete` object for activities' athlete
-    info. However, the actual return lacks expected attributes like resource_state,
-    first name, and last initial, differing from the spec but matching the actual
-    data.
-
+    The Strava API specification suggests that this should
+    return a `MetaAthlete` Object for athlete associated with the activities.
+    However, the object spec is missing what is actually returned,
+    i.e., resource_state, first name and last initial.
+    This object matches the actual return dat, not the spec.
```
jsamoocha, line 1302:
```suggestion
    The Strava API specification suggests that this should
    return a `MetaAthlete` Object for the athlete associated with this activity.
    However, the object spec is missing what is actually returned,
    i.e., resource_state, first name, and last initial.
    This object matches the actual return data, not the spec.
```

jsamoocha (commented):
Thanks for this overhaul and for adding docstrings to the new quantity types!

I added several wording suggestions, and have one question regarding mentioning the inheritance of `BoundClient` in docstrings (see my earlier comment on that for more details).

docs/conf.py
```diff
@@ -18,8 +18,10 @@
 
 import stravalib
 
+sys.setrecursionlimit(1500)
+
```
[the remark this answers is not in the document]
lwasser: oh this is definitely not needed anymore. great catch! i had forgotten about adding this.

docs/reference/model.rst
```diff
@@ -4,83 +4,105 @@ Model
...
-.. setting (BaseModel) see below. Testing a custom page that is
-.. cleaner to read.
-.. .. automodule:: stravalib.model
-..     :inherited-members: BaseModel
+    AthleteStats
+    MetaAthlete
+    SummaryAthlete
+    DetailedAthlete
+    AthletePrEffort
+    AthleteSegmentStats
```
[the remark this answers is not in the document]
lwasser: oh yes good catch.

docs/reference/model.rst
```diff
@@ -4,83 +4,105 @@ Model
...
+Club Classes
+-------------
+.. autosummary::
+   :toctree: api/
+
+   MetaClub
+   SummaryClub
+   DetailedClub
 
 Summary Functions
```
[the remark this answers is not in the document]
lwasser: oh good point. I'm fine with removing them.

docs/reference/model.rst
```diff
@@ -90,35 +112,9 @@ Route / Segment Classes
    SegmentEffort
    SegmentEffortAchievement
    SummarySegmentEffort
-   Map
    SegmentExplorerResult
+   Map
    BaseEffort
    BestEffort
```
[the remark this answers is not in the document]
lwasser: Gosh yes they are duplicate. Removed What do you think about this organization, Jonatan?

Should SegmentEffort go in Segment or Performance / Effort or here?

src/stravalib/model.py
lwasser: good point. i think i added it from a contributor perspective (ie it took me some time to understand why we had the pattern of inheritance) BUT i think you are correct - these are more for users so i'll remove mention of BoundClient. sklearn is a great model for us to follow!

lwasser (commented):
@jsamoocha a few notes

1. i just also caught that  docs for strava_model was commented out. have a look at the built docs to see what you think. i was thikning using automodule for our generated model made more sense because there could be new classes added. thoughts?

src/stravalib/model.py
```diff
@@ -599,11 +674,20 @@ class AthleteStats(strava_model.ActivityStats):
 
 
 class MetaAthlete(strava_model.MetaAthlete, BoundClientEntity):
```
lwasser, line 676: How would you document this class (the meta classes in general @jsamoocha ) if we don't talk about BoundClient. Should we focus on what is inheretied from strava model (so the x attributes this object carries?) (in this case just the ID?

src/stravalib/model.py
```diff
@@ -655,6 +740,7 @@ class DetailedAthlete(SummaryAthlete, strava_model.DetailedAthlete):
     sample_race_time: Optional[int] = None
     membership: Optional[str] = None
     admin: Optional[bool] = None
+    """Indicates if the athlete has admin privileges."""
     owner: Optional[bool] = None
```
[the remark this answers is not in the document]
lwasser: I did just remove these. But I suspect there are other classes / attributes that may have similar issues.  Can we target this issue more specifically [here](https://github.com/stravalib/stravalib/issues/528)?

lwasser:
Ok ready for another review. 

1. let's check for undocumented attributes via #528 in a more comprehensive way
2. How we describe those high level classes if we don't mention boundclient? 

let me know if i missed anything else!

docs/reference/model.rst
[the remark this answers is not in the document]
jsamoocha: The Strava API docs have a separate section for Segment Efforts. Maybe we can do the same?

src/stravalib/model.py
jsamoocha: I added a few commits with suggested changes to the docstrings in our Meta classes.

jsamoocha:
I added some commits that hopefully address your question about the docstrings for the Meta classes.

I tried to find the placeholder for the v2.0 announcement and/or migration guide but failed. Do we intend to add that in another PR? If so (and if you agree with my latest additions here), this PR seems finished and can be merged.

docs/conf.py
```diff
@@ -105,17 +106,16 @@
...
 html_theme_options = {
-    "header_links_before_dropdown": 4,
+    "header_links_before_dropdown": 5,
     "use_edit_page_button": True,
     "show_toc_level": 1,
     # "navbar_align": "left",  # [left, content, right] For testing that the navbar items align properly
     "github_url": "https://github.com/stravalib/stravalib",
     "footer_start": ["copyright"],
-    "announcement": "<a href='stravalib-2.html'>Stravalib 2.x is out 🚀! Check out our migration guide for tips on changes from Stravalib V1!</a>",
+    "announcement": "<a href='/whats-new/stravalib-2.html'>Stravalib 2.x is out 🚀! Check out our migration guide for tips on changes from Stravalib V1!</a>",
```
lwasser, line 118: The banner link was broken because i moved the file and my build wasn't clean. it's working again now. i fixed a few other issues with the changelog as well.

docs/contributing/documentation-changelog.md
lwasser, line 1: duplicate file. gosh i'm really off my game. i missed a LOT of small broken things!

src/stravalib/model.py
lwasser: thank you!

docs/conf.py
jsamoocha: Awesome, I see it now. I can add the content of that document sometime this week.

jsamoocha:
I added the "What's new" section with a migration guide and breaking changes. I also added docstring examples of all the new custom types for duration, distance, etc.

I tried to insert a link referencing /reference/model.html#custom-types, but failed due to lack of MyST experience. A hacky workaround would be to use a "plain" markdown link with an absolute reference to the published readthedocs. But maybe there's a way to have a working relative link.

Finally, I noticed that the summary section in Home > API Reference > Model does not have any links to the model classes. Can we add these?

lwasser:
Hi there! I am almost done w travel!! I will look at this either tomorrow night at the airport or when I'm back this upcoming week! We can definitely fill out whatever is missing in the api docs and i can work on it then!

lwasser:
> I tried to insert a link referencing /reference/model.html#custom-types, but failed due to lack of MyST experience.

ok no problem! i looked into this. this is actually an odd rst / sphinx issue! The fix was to add manual targets to the rst file and use those as a link like this

```
.. _custom-types-anchor:

## Custom types h2 here
```

i never knew about this quirk before ! good catch! 

regarding the strava_model file - i had commented out the automodule feature for that (probably in my debugging of that odd recursion issue!) and forgot about it. I have uncommented it!! 

so that is now fixed!!

@jsamoocha a few last things here..

The following items are not yet documented in our API docs. I will add them now - but i've been wondering if we want to include the helpers - check_valid_location, lazy_property and naive_datetime in a new module or not. On one hand they don't really fit nicely into a single module that we have. On the other they feel weird in the module.py file. 

i'm really open to leaving them as is for now too - i just wanted to bring it up! 
```
/Users/leahawasser/Documents/GitHub/stravalib/docs/reference/api/stravalib.model.BoundClientEntity.rst: WARNING: document isn't included in any toctree
/Users/leahawasser/Documents/GitHub/stravalib/docs/reference/api/stravalib.model.check_valid_location.rst: WARNING: document isn't included in any toctree
/Users/leahawasser/Documents/GitHub/stravalib/docs/reference/api/stravalib.model.lazy_property.rst: WARNING: document isn't included in any toctree
/Users/leahawasser/Documents/GitHub/stravalib/docs/reference/api/stravalib.model.naive_datetime.rst: WARNING: document isn't included in any toctree
```

for now i've added these in a stravalib helpers section in the model.rst file

docs/contributing/intro.md
```diff
@@ -69,5 +69,4 @@ Contributing Guide <../contributing/how-to-contribute>
 Development Guide <../contributing/development-guide>
 Build & Release Guide <../contributing/build-release-guide>
 New Contributor Resources <../contributing/resources-for-new-contributors>
```
lwasser, line 71: We have changelog in whats-new now!

docs/reference/model.rst
```diff
@@ -108,17 +90,24 @@ Subscription Classes
...
+--------------------------
+.. autosummary::
+    :toctree: api/
+
+    Distance
+    Duration
+    Timezone
+    Velocity
+
+Stravalib helpers
```
lwasser, line 105: I placed these at the bottom. Sphinx will provide warnings if we skip classes. If there is a strong reason to not include these helpers we can consider a manual removal! or just ignoring the warning. but it might be nice to include them for contributors.

docs/whats-new/stravalib-2.md
```diff
@@ -1,3 +1,71 @@
...
+supported. Instead, please
+use [Pydantic’s serialization mechanisms](https://docs.pydantic.dev/latest/concepts/serialization/).
+
+## Strava Meta/Summary/Detailed Type Hierarchy
+
+Many model classes were changed to reflect the type hierarchy defined in the
+Strava API. This means that classes such
+as `Activity` are now split into `MetaActivity`, `SummaryActivity`,
+and `DetailedActivity`.
+As `MetaActivity` has far fewer attributes than the former `Activity` class,
```
lwasser, line 29:
```suggestion
As `MetaActivity` has fewer attributes than the former `Activity` class,
```

docs/whats-new/stravalib-2.md
```diff
@@ -1,3 +1,71 @@
...
+object, e.g., `activity.distance` now returns a `Distance` type, which is an
+extension of `float`. The “plain”
+distance (in meters, the Strava default) can now be retrieved as a float
+by `activity.distance`. However, this distance
+can also be received as a `Quantity` with an explicit unit
+using `activity.distance.quantity`.
+This behavior differs from Stravalib V1, where `activity.distance` would
+immediately return a Quantity-like object.
+Detailed documentation about these new custom types can be
+[found here.](/reference/model.rst#custom-types-anchor)
```
lwasser, line 61: This should work now - it's calling the rst custom anchor.

lwasser (commented):
ok @jsamoocha i made a tiny edit to the migration page but otherwise i think i've addressed everything! let me know if you catch anything else!! otherwise, if you feel good about where things are with this PR, let's merge ✨ we are getting there!!

lwasser:
oddly getting a mypy error - but we haven't touched the code / typing  and it seems to run fine locally. 🤷🏻‍♀️

src/stravalib/client.py
```diff
@@ -297,7 +297,10 @@ def get_activities(
         after_epoch = self._utc_datetime_to_epoch(after) if after else None
         params = dict(before=before_epoch, after=after_epoch)
         result_fetcher = functools.partial(
-            self.protocol.get, "/athlete/activities", **params
+            self.protocol.get,
+            "/athlete/activities",
+            check_for_errors=True,
```
lwasser, line 302: This is silly. mypy is erroring because it expects a bool. but True is a default for check_for_errors. but what it sees is a dictionary to unpack as the second param provided to the get method. 

so i just added it here even tho it's redundant 🤷🏻‍♀️ if there is a better preferred way please let me know!

lwasser:
ok!! everything is building and green. i think this is good for another / final?? review?! then we can do some final cleanup / triage as needed?

jsamoocha (approved):
Great, thanks for fixing these remaining issues! Let's merge this.

lwasser:
Awesome merging!!

Issue #538, opened 2024-06-29
Stravalib 2.x: Recursion and overlapping attributes


lwasser:
When building our API docs i'm noticing some issues with the photo endpoint (we may not have checked these and i may a have a note about them in the docs ... Or maybe i missed something. 

but regardless there are multiple versions of the same  attribute

```
class ActivityPhotoMeta(PhotosSummary):
    """
    Represents the metadata of photos returned with the activity.

    Not to be confused with the fully loaded photos for an activity.

    Attributes
    ----------
    primary : ActivityPhotoPrimary, optional
        The primary photo for the activity.
    use_primary_photo : bool, optional
        Indicates whether the primary photo is used. Not currently documented
        by Strava.
```

and

```
class ActivityPhotoPrimary(Primary):
    """
    Represents the primary photo for an activity.

    Attributes
    ----------
    use_primary_photo : bool, optional
        Indicates whether the photo is used as the primary photo.

    Notes
    -----
    Attributes for activity photos are currently undocumented.
    """

    use_primary_photo: Optional[bool] = None
```

This causes a recursion error when building the docs. 

Error for photos and other attrs below

```
ERROR 1:

...path.../stravalib/.nox/docs/lib/python3.12/site-packages/stravalib/model.py:docstring of stravalib.model.ActivityComment.athlete:1: WARNING: duplicate object description of stravalib.model.ActivityComment.athlete, other instance in reference/api/stravalib.model.ActivityComment, use :no-index: for one of them

ERROR 2:
...path.../stravalib/.nox/docs/lib/python3.12/site-packages/stravalib/model.py:docstring of stravalib.model.ActivityPhotoMeta.primary:1: WARNING: duplicate object description of stravalib.model.ActivityPhotoMeta.primary, other instance in reference/api/stravalib.model.ActivityPhotoMeta, use :no-index: for one of them

ERROR 3:
...path.../stravalib/.nox/docs/lib/python3.12/site-packages/stravalib/model.py:docstring of stravalib.model.ActivityPhotoMeta.use_primary_photo:1: WARNING: duplicate object description of stravalib.model.ActivityPhotoMeta.use_primary_photo, other instance in reference/api/stravalib.model.ActivityPhotoMeta, use :no-index: for one of them

ERROR 4:
[...]

PR #539, opened 2024-06-29
Add: what's new section to the docs

lwasser:
This PR just cleans up our docs and adds a banner + a what's new section.

Right now the docs won't build because of a recursion issue mentioned in #538 which we need to fix first. i'm going to take a quick look at that next to see how involved it will be to address it. because of how we're building our docs, i'm not sure how to create the :noindex: flag to ignore the recursion issue (which is a duplicated attribute in two classes. (this happens a few times). 

UPDATE:

i actually looked into this and am not sure how to handle things here. We could merge just to work on the discussion around what's changed. but it seems like there may be some inheritance issues that we need to sort out first


jsamoocha (approved):
Thanks, this is good! I can work on expanding the migration guide.

PR #540, opened 2024-06-30
Fix: Photos endpoint cleanup

lwasser:
I know that this wasn't in our list of issues but the photos endpoint diverged from the Spec so I just cleaned it up a bit.  
The changes here just realign things with the strava spec where possible. i do understand that get_activity_photos is not a supported endpoint by strava currently. but i was able to get it to work after a lot of digging and then a realization that I should have just read the docstring for that method 🙃 first (instead of last) 🙈 i'm sorry for the excessive pings yesterday @jsamoocha

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -654,13 +654,6 @@ class Client:
         class: `model.DetailedActivity`
             An Activity object containing the requested activity data.
 
-        Examples
-        --------
-        # Assuming an authenticated client
-        >> activity = client.get_activity(activity_id = 12345677)
-        >> activity.moving_time
-        1234
-
         """
         raw = self.protocol.get(
             "/activities/{id}",
@@ -1161,11 +1154,6 @@ class Client:
         -------
         class:`BatchedResultsIterator`
             An iterator of :class:`stravalib.model.ActivityPhoto` objects.
-
-        Examples
-        --------
-        # Assuming an authenticated client
-        >> activity = client.get_activity_photos(activity_id = 12345677)
 
         """
         params: dict[str, Any] = {}
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -654,6 +654,13 @@ class Client:
         class: `model.DetailedActivity`
             An Activity object containing the requested activity data.
 
+        Examples
+        --------
+        # Assuming an authenticated client
+        >> activity = client.get_activity(activity_id = 12345677)
+        >> activity.moving_time
+        1234
+
         """
         raw = self.protocol.get(
             "/activities/{id}",
@@ -1154,6 +1161,11 @@ class Client:
         -------
         class:`BatchedResultsIterator`
             An iterator of :class:`stravalib.model.ActivityPhoto` objects.
+
+        Examples
+        --------
+        # Assuming an authenticated client
+        >> activity = client.get_activity_photos(activity_id = 12345677)
 
         """
         params: dict[str, Any] = {}
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -38,6 +38,7 @@ from dateutil import parser
 from dateutil.parser import ParserError
 from pydantic import (
     AliasChoices,
+    AnyHttpUrl,
     BaseModel,
     Field,
     GetCoreSchemaHandler,
@@ -56,9 +57,7 @@ from stravalib.strava_model import (
     Comment,
     ExplorerSegment,
     LatLng,
-    PhotosSummary,
     PolylineMap,
-    Primary,
     SportType,
     SummaryGear,
 )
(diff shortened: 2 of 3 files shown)


src/stravalib/client.py
```diff
@@ -654,6 +654,13 @@ def get_activity(
         class: `model.DetailedActivity`
             An Activity object containing the requested activity data.
 
+        Examples
+        --------
+        # Assuming an authenticated client
```
lwasser, line 659: we could setup doctest for these but i don't yet know how we'd do that without needing a token / proper authentication.

src/stravalib/model.py
```diff
@@ -38,6 +38,7 @@
 from dateutil.parser import ParserError
 from pydantic import (
     AliasChoices,
+    AnyHttpUrl,
```
lwasser, line 41: I read a bit more and this is an actual url object return. this could break people's code the way i've added it. so please let me know if this isn't ideal. if we agree to use it we can then check any field that expects a url (vs just accepting a string)

src/stravalib/model.py
```diff
@@ -717,32 +716,33 @@ class ActivityComment(Comment):
...
 
-class ActivityPhotoPrimary(Primary):
-    """
-    Represents the primary photo for an activity.
-
-    Attributes
-    ----------
-    use_primary_photo : bool, optional
-        Indicates whether the photo is used as the primary photo.
+class PhotosSummary_primary(BaseModel):
```
lwasser, line 719: renamed and updated it to align with the spec

src/stravalib/model.py
```diff
@@ -717,32 +716,33 @@ class ActivityComment(Comment):
...
 
-    use_primary_photo: Optional[bool] = None
+    unique_id: str | None = None
+    src: int | None = None
+    media_type: int | None = None
+    urls: dict[str, AnyHttpUrl] | None = None
 
 
-class ActivityPhotoMeta(PhotosSummary):
+class PhotosSummary(strava_model.PhotosSummary):
```
lwasser, line 736: similar to the above. NOTE that the spec doesn't align with what's actually returned but this is close

src/stravalib/model.py
```diff
@@ -751,15 +751,20 @@ class ActivityPhotoMeta(PhotosSummary):
...
-    primary: Optional[ActivityPhotoPrimary] = None
+    primary: Optional[PhotosSummary_primary] = None
 
     # Undocumented by strava
     use_primary_photo: Optional[bool] = None
 
 
+# TODO: When you call get_Ativity_photos it returns a list of this
+# object
 class ActivityPhoto(BaseModel):
```
lwasser, line 768:
```suggestion
class ActivityPhoto(BaseModel):
```

src/stravalib/model.py
```diff
@@ -769,23 +774,25 @@ class ActivityPhoto(BaseModel):
     endpoint to retrieve it
     """
 
-    athlete_id: Optional[int] = None
     activity_id: Optional[int] = None
```
lwasser, line 777: i alphabetizes these. it was driving me crazy trying to check against the API and both were in different orders. atleast now ours is in a specific order that we can follow.

src/stravalib/model.py
```diff
@@ -769,23 +774,25 @@ class ActivityPhoto(BaseModel):
...
     post_id: Optional[int] = None
-    default_photo: Optional[bool] = None
+    sizes: Optional[dict[str, Sequence[int]]] = None
     source: Optional[int] = None
+    type: Optional[int] = None
+    unique_id: Optional[str] = None
+    uploaded_at: Optional[datetime] = None
+    urls: Optional[dict[str, AnyHttpUrl]] = None
+
+    # TODO: i don't see these in the actual returns
```
lwasser, line 793: I did not see these in the actual return. is it ok to delete?

lwasser, line 793:
```suggestion
```

src/stravalib/model.py
```diff
@@ -769,23 +772,24 @@ class ActivityPhoto(BaseModel):
...
-    default_photo: Optional[bool] = None
+    sizes: Optional[dict[str, Sequence[int]]] = None
     source: Optional[int] = None
+    type: Optional[int] = None
+    unique_id: Optional[str] = None
+    uploaded_at: Optional[datetime] = None
+    urls: Optional[dict[str, AnyHttpUrl]] = None
+
+    # ref: Optional[str] = None
+    # uid: Optional[str] = None
```
lwasser, line 793:
```suggestion
```

lwasser: If you agree i didn't see these in the actual return.

src/stravalib/model.py
lwasser, line 41: We can remove this. i started to type URL's as this given it's the proper type for a url string (i think) for photos, etc. but strava_model doesn't use this at all.

src/stravalib/model.py
```diff
@@ -717,32 +716,49 @@ class ActivityComment(Comment):
...
+    # what is the best path forward here?
+    """
+    "unique_id": "8d9e546c-5da1-46cb-973e-c55bb3be88bd",
+            "urls": {
+                "600": "https://dgtzuqphqg23d.cloudfront.net/PNON86jIyK-xCnvSXhfESDYO--7cwvweK2fQ86NwOws-768x354.jpg",
+                "100": "https://dgtzuqphqg23d.cloudfront.net/PNON86jIyK-xCnvSXhfESDYO--7cwvweK2fQ86NwOws-128x59.jpg"
+            },
+            "source": 1,
+            "media_type": 1
+    """
```
lwasser, line 745:
```suggestion
```

src/stravalib/model.py
```diff
@@ -769,23 +788,25 @@ class ActivityPhoto(BaseModel):
...
     post_id: Optional[int] = None
-    default_photo: Optional[bool] = None
+    sizes: Optional[dict[str, Sequence[int]]] = None
     source: Optional[int] = None
+    type: Optional[int] = None
+    unique_id: Optional[str] = None
+    uploaded_at: Optional[datetime] = None
+    urls: Optional[dict[str, AnyHttpUrl]] = None
+
+    # TODO: i don't see these in the actual returns but they are used below
```
lwasser, line 807: I didn't see these in the photo return. but haven't looked into how / whether we need them. i just know we use them below in our code.

src/stravalib/model.py
lwasser: Note - i did notice after another pass that these attrs are used below. i can revisit if i'm missing something (which i think i must be).

src/stravalib/client.py
jsamoocha: An examples section in each client method's docstring seems to be overkill for this API. We already provide a link to the Strava API docs (which provide example responses). If particular parameters have a not-too-obvious effect on the method's behavior, we can always add examples on a per-instance basis.

As you mention, doctest wouldn't work here unless we're willing to jump through burning hoops to integrate doctest with our Strava API mock.

src/stravalib/model.py
```diff
@@ -717,32 +716,35 @@ class ActivityComment(Comment):
...
-    """
-    Represents the primary photo for an activity.
-
-    Attributes
-    ----------
-    use_primary_photo : bool, optional
-        Indicates whether the photo is used as the primary photo.
+# strava_model Primary is fine for this but we may
+# have to add some mapping
+class PhotosSummary_primary(strava_model.Primary):
```
jsamoocha, line 721: Is PEP8 happy with that class name? Maybe use `PhotosSummaryPrimary`?

src/stravalib/model.py
```diff
@@ -717,32 +716,35 @@ class ActivityComment(Comment):
...
+    """A object to store data about the primary photo of a Strava activity
```
jsamoocha, line 722:
```suggestion
    """An object to store data about the primary photo of a Strava activity
```

src/stravalib/model.py
```diff
@@ -717,32 +716,35 @@ class ActivityComment(Comment):
...
 
     Notes
     -----
-    Attributes for activity photos are currently undocumented.
+    https://developers.strava.com/docs/reference/#api-models-PhotosSummary_primary
+    This object is nor returned in strava_model but it is documented in the
+    Strava API V3 spec. The Spec documentation deviates from what is actually
+    returned. This model represents what is actually returned
```
jsamoocha, line 729:
```suggestion
    https://developers.strava.com/docs/reference/#api-models-PhotosSummary_primary
    This object is not returned in strava_model, but it is documented in the
    Strava API V3 spec. The Spec documentation deviates from what is actually
    returned. This model represents what is actually returned.
```

src/stravalib/model.py
```diff
@@ -717,32 +716,35 @@ class ActivityComment(Comment):
...
     """
 
-    use_primary_photo: Optional[bool] = None
+    # Strava_model has this defined as id but unique_id is the correct attr name
+    # it's also defined as a int but should be a string
```
jsamoocha, line 733:
```suggestion
    # it's also defined as an int but should be a string
```

jsamoocha (approved):
Thanks! Apart from some minor typos and a potential class name issue, this is good to merge.

lwasser:
Merging this!!

PR #541, opened 2024-06-30
Activity Comment cleanup

lwasser:
I could be missing something here. But part of the docs issues related to duplicated attributes. 

When i looked at model.ActivityComment is had a single attribute athlete and it inherited from BaseModel. however in strava_model there is a Comment object that has the same attr +. The spec calls for a return of a Comment object so I just removed the model.ActivityComment and imported strava_model.Comment instead. 

If there was some reason that we added this, lt's keep it. But it seems to be at this point it's a redundant class. I wondered if maybe the strava_model didn't have `Comment` before and now it does.

### DIFF
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -702,7 +702,7 @@ class DetailedAthlete(SummaryAthlete, strava_model.DetailedAthlete):
         return self.bound_client.get_athlete_stats(self.id)
 
 
-class ActivityPhotoPrimary(Primary):
+class ActivityPhotoPrimary(strava_model.Primary):
     """
     Represents the primary photo for an activity.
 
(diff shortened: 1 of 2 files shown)


src/stravalib/model.py
```diff
@@ -704,19 +704,6 @@ def stats(self) -> AthleteStats:
         return self.bound_client.get_athlete_stats(self.id)
 
 
-class ActivityComment(Comment):
```
lwasser, line 707: This is just a duplicate class of strava_model.Comment

jsamoocha (approved):
Thanks! This looks good to merge.

The `ActivityComment` class was probably there because it referenced our version of the `Athlete` class (which was a `BoundClientEntity`). However, due to our introduction of the meta-summary-detailed hierarchy, the comment now (correctly) refers to `SummaryAthlete` (which does not have any lazy properties that require the bound client). As such, our custom `ActivityComment` class is no longer necessary.

lwasser:
that makes sense! i'll merge this!

lwasser:
Thanks again for the review - i'm going to now integrate all of this into our docs pr and try to see if i can figure out what is happening with the recursion issues and if any of these fixes help!

PR #546, opened 2024-07-21
Fix enhanced types doc

### LINKED ISSUE #502: Stravalib 2.0 -> Documentation Update

jsamoocha:
closes #502 

This should be the last bit of documentation required to close #502. I just modified the existing "Overview" subsection of "Getting Started" to reflect the changes in the enhanced types.

### DIFF
--- a/docs/get-started/overview.rst
+++ b/docs/get-started/overview.rst
@@ -49,33 +49,37 @@ attributes on entities.::
 Attribute Types and Units
 =========================
 
-Many of the attributes in the Strava API are either temporal (or interval) types or quantities
-that have implicit units associated with them. In both cases these are represented using
-richer python types than the simple string or numeric values that Strava REST API returns.
+Many of the attributes in the Strava API are either temporal (or interval) types or
+quantities that have implicit units associated with them. In both cases, richer python
+types than the simple string or numeric values that the Strava REST API returns can be
+accessed as follows:
 
 Date/Time Types
 ---------------
 
-The date+time responses are encoded as python native :class:`datetime.datetime` objects.::
+The date+time responses are encoded as python native :class:`datetime.datetime`
+objects.::
 
    a = client.get_activity(96089609)
    print(a.start_date)
    # 2013-11-17 16:00:00+00:00
 
-Date values which have no time component are encoded as python native :class:`datetime.date` objects.::
+Date values which have no time component are encoded as python native
+:class:`datetime.date` objects.::
 
    me = client.get_athlete()
    print(me.dateofbirth)
    # 2010-12-26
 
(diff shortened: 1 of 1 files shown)


docs/get-started/overview.rst
```diff
@@ -17,19 +17,6 @@ types are instances of :mod:`stravalib.model` classes. For example::
...
-Entities in Strava's API exist at different detail levels, denoted by the numeric
-`resource_state` attribute (1=metadata, 2=summary, 3=detailed).  In general, detail level
-3 ("detailed") is only available to the user that is authenticated.  Detail level 2 ("summary")
-information is available to others.  For example::
-
-   other_athlete = client.get_athlete(123) # Some other Athlete ID
-   other_athlete.firstname  # This is accessible
-   # But this is not:
-   # other_athlete.email
-
```
jsamoocha, line 32: This section became irrelevant because we now use the Meta/Summary/Detailed type hierarchy of Strava

docs/get-started/overview.rst
```diff
@@ -99,15 +90,17 @@ what people would actually want to see (e.g. meters-per-second instead of
...
 
    activity = client.get_activity(96089609)
-   print(activity.distance)
-   # 22530.80 m
+   print(activity.distance.quantity())
+   # 22530.80 meter
 
-Hmmm, meters.  Well, here in the US we like to see miles.  While you can certainly do this using the units
-library directly, stravalib provides a preconfigured set of common units to simplify matters.::
+Hmmm, meters.  Well, here in the US we like to see miles.  While you can certainly do
```
lwasser, line 101: @jsamoocha i'm just noticing this language! should we consider making it less USA centric? I don't mind  being a US native. BUT you and others may prefer meters.  🙃 so we could either leave as is or make it a bit less country agnostic. 

like 

> Here, the value is returned in meters. If you prefer units in miles, you can use stravalib's preconfigured set of community units...

As written it's fun and light -- so you tell me if you want to drop the "here in the US to make it more agnostic" or not! :)

lwasser (approved):
This looks great to me. i have no feedback other than - we can decide to make the language less US focused -- or not. depending on how it reads to you! 

Also -- I didn't actually understand what `resource_state` was for so that is helpful to know!

docs/get-started/overview.rst
jsamoocha: I don't see any problem with that language, so let's not get too formal on that 😁.

PR #555, opened 2024-08-07
fix(codecov): relax code coverage requirements

lwasser:
This is a tiny pr to relax the requirements for code cov. this way it won't fail a pr if the test coverage is within 10% of main's current state.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,11 @@
+# Set non-blocking status checks
+coverage:
+  status:
+    project:
+      default:
+        # Allows 10% drop from previous base commit
+        threshold: 10%
+        informational: true
+    patch:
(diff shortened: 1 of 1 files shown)


lwasser:
i am going to merge this pr. the errors it has are not at all related to this pr which simple relaxings the code cov config to make sure it doesn't always fail. right now it fails a pr if the test changes are even .1 % different. this pr bumps that to 10% for the time being. it should still show us where test coverage gaps are but it won't fail the pr. @jsamoocha i will merge but it you want me to revert later i can. i can have a look at the other failed elements once we merge pydantic v2 branch. it doesn't make sense to fix things that should be fixed via merging pydantic-v2 on this pr!!

PR #567, opened 2024-08-22
Add get_athlete_zone method on client

### LINKED ISSUE #508: ENH: implement `athlete/zones` GET client method
### Feature Type

- [X] Adding new functionality to stravalib

- [ ] Changing existing functionality in stravalib

### Problem Description

This endpoint is not implemented

### Feature Description

Implement a client method

### Alternative Solutions

None

### Additional Context

_No response_

### Code of Conduct

enadeau:
closes #508 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

I'm having some issue getting the test to run. The new method runs as expected my I use my token but the test is actually failling. Printin the `raw` variable shows that it is this list that I don't understand when it comes from
```
[{'distribution_buckets': [{'max': 0, 'min': 0, 'time': 1498}, {'max': 50, 'min': 0, 'time': 62}, {'max': 100, 'min': 50, 'time': 169}, {'max': 150, 'min': 100, 'time': 536}, {'max': 200, 'min': 150, 'time': 672}, {'max': 250, 'min': 200, 'time': 821}, {'max': 300, 'min': 250, 'time': 529}, {'max': 350, 'min': 300, 'time': 251}, {'max': 400, 'min': 350, 'time': 80}, {'max': 450, 'min': 400, 'time': 81}, {'max': -1, 'min': 450, 'time': 343}], 'type': 'power', 'resource_state': 3, 'sensor_based': True}]
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add: `get_athelete_zones` method on client
+
 ## v2.0.0
 
 ### Major Changes in This Release
--- a/changelog.md
+++ b/changelog.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Add: `get_athelete_zones` method on client
+- Add: `get_athelete_zones` method on client (@enadeau, #508)
 
 ## v2.0.0
 
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -391,6 +391,21 @@ class Client:
             {**raw_athlete, **{"bound_client": self}}
         )
 
+    def get_athlete_zones(self) -> strava_model.Zones:
+        """
+        Get the authenticated athlete's heart rate and power zones
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        class:`stravalib.model.Zones`
+            The athlete zones model object.
+        """
+        raw = self.protocol.get("/athlete/zones")
+        return strava_model.Zones.model_validate(raw)
+
     def get_athlete_koms(
         self, athlete_id: int, limit: int | None = None
     ) -> BatchedResultsIterator[model.SegmentEffort]:
--- a/src/stravalib/tests/integration/test_client.py
+++ b/src/stravalib/tests/integration/test_client.py
@@ -10,7 +10,7 @@ from responses import matchers
 from stravalib.client import ActivityUploader
 from stravalib.exc import AccessUnauthorized, ActivityPhotoUploadFailed
 from stravalib.model import DetailedAthlete, SummaryAthlete, SummarySegment
-from stravalib.strava_model import SummaryActivity
+from stravalib.strava_model import SummaryActivity, Zones
 from stravalib.tests import RESOURCES_DIR
 from stravalib.unit_helper import miles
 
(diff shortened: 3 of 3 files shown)


lwasser:
hey there @enadeau 👋🏻  thank you for this pr🚀!! 

That value comes from the swagger.json file that is in our tests/resources dir [here](https://github.com/stravalib/stravalib/blob/main/src/stravalib/tests/resources/strava_swagger.json#L259)

It took me a bit to understand this as well. It could make sense to document it further as I always have to remind myself. 
Screenshot 2024-08-24 at 1 50 52 PM

so `mock_strava_api.get("/athlete/zones")` grabs the API mock that Jonatan created which returns the response in the swagger.json  file that is in the resources dir. That is where the distribution bucket data comes from.

the test is currently failing because of a validation error during the validation step when it tries to populate the model:

```
1 validation error for Zones
  Input should be a valid dictionary or instance of Zones [type=model_type, input_value=[{'distribution_buckets':..., 'sensor_based': True}], input_type=list]
    For further information visit https://errors.pydantic.dev/2.8/v/model_type
```

I haven't gotten further than that yet but I hope that is a helpful start just so you understand where the mocked data are coming from. I can try to debug a bit more later if things still aren't working as expected.

enadeau:
Hi thanks for the help I kind of knew this came for a swagger specification but didn't know where to actually find it.

I think this my actually be an error in the swagger file from strava as the data of the example looks nothing like the actual response I'm getting from the api.

In the documentation [here]( https://developers.strava.com/docs/reference/#api-Athletes-getLoggedInAthleteZones) the example is not of the same form as the described output which is supposed to be of type `Zones`

Not quite sure maybe we should be editiing the response from the mock to fit the actual response from the api.

lwasser:
> In the documentation [here](https://developers.strava.com/docs/reference/#api-Athletes-getLoggedInAthleteZones) the example is not of the same form as the described output which is supposed to be of type Zones
> Not quite sure maybe we should be editing the response from the mock to fit the actual response from the API.

I do see that mismatch between that swagger.json file and the actual data returned! Your method works fine when I run it. And we've seen this type of mismatch over and over for other endpoints just never on the test side of things with that json file!  

I say this cautiously because I've not had to modify that json file yet. But, let's modify the swagger.json file for the time being to better align with the actual returned data structure. 🙈 The method that you wrote works well as far as I can tell.  

What I'm curious about is - I only have running and hiking data so I have no power data on my account to test if more would be returned with additional data. Do you by chance have power data as well?

When Jonatan is back, he can have a look too in case I too am missing something. But looking at the [history for the file](https://github.com/stravalib/stravalib/commits/main/src/stravalib/tests/resources/strava_swagger.json) don't think that file has been updated from the API directly so any changes should be ok.

jsamoocha:
Unfortunately, the examples from the API spec not always match the actual responses from Strava. It's better not to rewrite the swagger.json, as Strava may update that file in the future. You can catch the response from the "real" Strava API and supply it to the mock using the `json` kwarg (see, e.g., `test_get_activity_zones()`).

lwasser:
Thank you jonatan!! I looked at that swagger.json file history and it only
had been modified once. Does our bot workflow update it? I missed where
that happens.

I will update the docs about this as my next pr and ask you to review. This
was what I wasnt sure about!

yihong0618 (approved):

changelog.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add: `get_athelete_zones` method on client
```
jsamoocha, line 7:
```suggestion
- Add: `get_athelete_zones` method on client (@enadeau, #508)
```

jsamoocha (approved):
Thanks for the PR! If you agree with my tiny change to the changelog, this is good to merge.

PR #569, opened 2024-08-29
Specialize type for on summary activity

### LINKED ISSUE #560: ENH: Specific exception type for rate limited response from strava api
### Feature Type

### Problem Description

If the client is created with `rate_limit_requests=False` then would expect a specific exception type to be raised when a rate limit is reached. This will allow apps to catch this exception themselves. At the moment a Fault exception is being raised which is too generic.

### Feature Description

In `protocol.py `in function `_handle_protocol_error` add specific error type for when 429 response is received and the client is setup with rate limit requests = False. 

### Alternative Solutions

Not sure

### Additional Context

When building APIs on top of Strava using this library, the API cannot use rate_limit_requests and needs the client to report back any 429 response so itself can report back to its clients.

### Code of Conduct

### LINKED ISSUE #570: Some fields have different types in stravalib.model.SummaryActivity  and stravalib.model.DetailedActivity
### Discussed in https://github.com/orgs/stravalib/discussions/568

<div type='discussions-op-text'>

<sup>Originally posted by **jlelong** August 29, 2024</sup>
Thanks a lot for library!

Fields related to elevation, duration and speed have a float type in `SummaryActivity` but more precise types in `DetailedActivity` such as `DistanceType`, `ElevationType`, VelocityType`. Could it be possible to use the same types in both `SummaryActivity` and `DetailedActivity` to make sure  a `DetailedActivity` can be cast to a `SummaryActivity`?

I would like to be able to use a `DetailedActivity` when a `SummaryActivity` is expected but I cannot figure out how to do it.</div>

enadeau:
closes #570 

## Description

Some field of ActivitySummary were only casted into our custom type on DetailActivity. This PR addresses that. 
Hopefully I've found them all.

I feel like this change is not breaking but not 100% sure. I'd take the opinion of the other dev on that

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,10 @@
 
 - Add: `get_athelete_zones` method on client (@enadeau, #508)
 
+### Fixed
+
+- Some override moved from DetailedActivity to SummaryActivity (@enadeau, #570)
+
 ## v2.0.0
 
 ### Major Changes in This Release
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -1242,12 +1242,19 @@ class SummaryActivity(MetaActivity, strava_model.SummaryActivity):
 
     # field overrides from superclass for type extensions:
     athlete: MetaAthlete | None = None
+    average_speed: VelocityType | None = None
+    distance: DistanceType | None = None
+    elapsed_time: DurationType | None = None
     # These force validator to run on lat/lon
     start_latlng: LatLon | None = None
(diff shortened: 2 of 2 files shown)


jsamoocha (approved):
Thanks for this fix! More overrides may have ended up in the wrong ("Detailed-") subclass, but we can handle these in other PRs. This can be merged.

PR #584, opened 2024-09-12
FIX: ensures ActivityType instances can be compared to str

### LINKED ISSUE #583: BUG: ActivityType comparison and serialization issues
### Stravalib version checks

- [X] I have tested this in a new clean environment with only stravalib and core python files.

- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the [latest version](https://pypi.org/project/stravalib/) of stravalib.

- [X] I have confirmed this bug exists on the [main branch](https://github.com/stravalib/stravalib) of stravalib.

### What operating system are you seeing the problem on?

Mac

### What version of python or you running?

```python
3.10.10
```

### Reproducible Example

```python
from stravalib.model import SummaryActivity

a = SummaryActivity.model_validate({'type': 'Run'})
a.type == 'Run'
# Expecting True, but returns False
```

### Issue Description

An activity's `type` attribute is of type `RelaxedActivityType`, but (since the pydantic v2 bump) doesn't behave as a `Literal[str]` anymore. This leads to subtle but potentially dangerous bugs, as seen above. This is a breaking change in 2.0 that we didn't document.

### Expected Behavior

`SummaryActivity.type` should return a type comparable with the previously used `Literal[]`, i.e., something that can be compared to `str` types and should be serialized as `str`.

### Your environment

Not appliccable

### Code of Conduct

jsamoocha:
closes #583

## Description

Ensures ActivityType instances can be compared to str. This prevents an undocumented breaking change from stravalib v1.

## Type of change

## Does your PR include tests

- [x] Yes

Updated existing test.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Some override moved from DetailedActivity to SummaryActivity (@enadeau, #570)
+- Ensures ActivityType instances can be compared to str (@jsamoocha, #583)
 
 ## v2.0.0
 
--- a/src/stravalib/model.py
+++ b/src/stravalib/model.py
@@ -522,6 +522,13 @@ class RelaxedActivityType(ActivityType):
             values = "Workout"
         return values
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, RelaxedActivityType):
+            return other == self
+        elif isinstance(other, str):
+            return other == self.root
+        return False
+
 
 class RelaxedSportType(SportType):
     """A class that extends the list of Literal values allowed for Sport Types
@@ -551,6 +558,13 @@ class RelaxedSportType(SportType):
             )
             values = "Workout"
         return values
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, RelaxedSportType):
+            return other == self
+        elif isinstance(other, str):
+            return other == self.root
+        return False
 
 
 class LatLon(LatLng):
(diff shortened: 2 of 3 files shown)


lwasser (approved):
@jsamoocha  this looks good to me. 

one suggestion - is it might be nice to use `@pytest.mark.parametrize` here like we've done in the other tests to make things a bit more concise. 

I haven't actually run this locally but it should be something like the below:

```python
@pytest.mark.parametrize("a_type, b_type, expected", [
    ({"type": "Run"}, {"type": "Run"}, True),  
    ({"type": "Run"}, {"type": "Ride"}, False),  
    ({"type": "Run"}, {"id": 42}, False)     
])
def test_relaxed_activity_type(a_type, b_type, expected):
    a = SummaryActivity.model_validate(a_type)
    b = SummaryActivity.model_validate(b_type)
    assert (a.type == b.type) == expected
 ```
```python
@pytest.mark.parametrize("a_sport_type, b_sport_type, expected", [
    ({"sport_type": "Run"}, {"sport_type": "Run"}, True),   # Equality
    ({"sport_type": "Run"}, {"sport_type": "Ride"}, False),  # Non-equality
    ({"sport_type": "Run"}, {"id": 42}, False)              # Comparison with different field
])
def test_relaxed_sport_type(a_sport_type, b_sport_type, expected):
    a = SummaryActivity.model_validate(a_sport_type)
    b = SummaryActivity.model_validate(b_sport_type)
    assert (a.sport_type == b.sport_type) == expected

```

Regardless, please merge as you see fit!!

changelog.md
```diff
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Some override moved from DetailedActivity to SummaryActivity (@enadeau, #570)
+- Ensures ActivityType instances can be compared to str (@jsamoocha, #583)
```
lwasser, line 12: 🚀

PR #588, opened 2024-09-20
fix(docs): tiny update to streaming data

### LINKED ISSUE #233: Streams API no longer accepts resolution="all"
"all" is not recognized as a valid value for the "resolution" parameter in the Streams API. When I make a request with resolution=all I get a 400 error. It still works for values of "high", "medium", and "low" (high seems to be the default now).

For what it's worth, the "resolution" parameter has been removed from the [official docs](https://developers.strava.com/docs/reference/#api-Streams-getActivityStreams), so maybe it will be completely removed in the future.

The loss of this parameter means it's impossible to get full resolution streams for long activities. Does anyone know if there's another way of getting that data?

https://github.com/hozn/stravalib/blob/b8d1181e32248c6570f0adfa995cb6c7adfccf62/stravalib/client.py#L1190-L1191

lwasser:
closes #233 

## Description

This is a tiny PR that clarifies how streaming data resolution works with stravalib and strava. I also added a note to the 
docstring so the user understands what `resolution=None` returns following @djcunningham0  issue.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

this is a tiny update, so i didn't include a changelog item for it.

### DIFF
--- a/docs/get-started/activities.md
+++ b/docs/get-started/activities.md
@@ -101,7 +101,7 @@ You can access each comment or item within a `BatchedResultsIterator` object usi
 Comment by: YourFriendsNameHere: Not the pool!
 ```
 
-## Get Activity streams information
+## Get Strava activity streams
 
 {py:func}`stravalib.client.Client.get_activity_streams` returns a dictionary containing time-series
 data associated with your activity.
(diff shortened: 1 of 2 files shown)


jsamoocha (approved):
Thanks for this update 🙏

PR #589, opened 2024-09-20
feat(rename): rename rst --> md

lwasser:
This is a simple pr that updates the activities.rst file to md format. The only changes to the text files I made here are syntax. I am going to merge this so that when I work on the activities.md file, we have an easy-to-view diff to look at!


docs/get-started/athletes.md
```diff
@@ -0,0 +1,65 @@
+(athletes)=
```
lwasser, line 1: There are no content changes in this pr. It is just a new syntax! I will submit a future pr with text changes that are reviewable!

PR #603, opened 2024-11-10
feat(docs): add specifics about using our mock fixture

lwasser:
closes #292

I've been fighting a bit with the test suite and so I took some time to better understand our mock setup.
This PR should close the existing issue around documenting the mock. 

 I could also add some examples of using it if that would be useful? 

## Description

I added some text around how the mock works. 
## Type of change

Select the statement best describes this pull request.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x] This is a documentation update
- [ ] This is an infrastructure update (docs, ci, etc)
- [ ] Other (please describe)

## Does your PR include tests

If you are fixing a bug or adding a feature, we appreciate (but do not require)
tests to support whatever fix of feature you're implementing.

- [ ] Yes
- [ ] No, i'd like some help with tests
- [x] This change doesn't require tests

## Did you include your contribution to the change log?

- [x] Yes, `changelog.md` is up-to-date.

PR #604, opened 2024-11-10
fix(endpoint): Update segment_efforts endpoint to match the currently documented endpoint

### LINKED ISSUE #321: Segment Efforts client method uses undocumented API endpoint
There are currently two client methods:

- `get_segment_efforts()`
- `get_segment_effort()`

The first uses an undocumented endpoint `/segments/{segment_id}/all_efforts`. Both should use the documented/supported Strava API endpoint `/segment_efforts`; see https://developers.strava.com/docs/reference/#api-SegmentEfforts.

lwasser:
closes #321

## Description

This pr fixes the segment_efforts api endpoint. I'm struggling a lot with understanding what is broken in the tests and would appreciate any input! This is why I created #603 as I think i generally understand the mock behavior but i clearly don't understand how to use it as these tests don't work as expected. 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

### DIFF
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -1427,7 +1427,7 @@ class Client:
         params: dict[str, Any] = {"segment_id": segment_id}
 
         if athlete_id is not None:
-            warn_param_unsupported(athlete_id)
+            warn_param_unsupported("athlete_id")
             params["athlete_id"] = athlete_id
 
         if start_date_local:
@@ -1445,7 +1445,7 @@ class Client:
             )
 
         if limit is not None:
-            warn_param_deprecation(limit, "date ranges")
+            warn_param_deprecation("limit", "date ranges")
             params["limit"] = limit
 
         result_fetcher = functools.partial(
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -1250,8 +1250,6 @@ class Client:
     def get_segment_effort(self, effort_id: int) -> model.SegmentEffort:
         """Return a specific segment effort by ID.
 
-        https://developers.strava.com/docs/reference/#api-SegmentEfforts
-
         Parameters
         ----------
         effort_id : int
@@ -1261,6 +1259,10 @@ class Client:
         -------
         class:`stravalib.model.SegmentEffort`
             The specified effort on a segment.
+
+        Notes
+        ------
+        https://developers.strava.com/docs/reference/#api-SegmentEfforts
 
         """
         return model.SegmentEffort.model_validate(
@@ -1358,6 +1360,7 @@ class Client:
             limit=limit,
         )
 
+    # Next TODO: add tests to deprecated items
     def get_segment_efforts(
         self,
         segment_id: int,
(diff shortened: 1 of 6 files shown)


src/stravalib/client.py
```diff
@@ -1262,6 +1260,10 @@ def get_segment_effort(self, effort_id: int) -> model.SegmentEffort:
         class:`stravalib.model.SegmentEffort`
             The specified effort on a segment.
 
+        Notes
+        ------
+        https://developers.strava.com/docs/reference/#api-SegmentEfforts
+
         """
         return model.SegmentEffort.model_validate(
             self.protocol.get("/segment_efforts/{id}", id=effort_id)
```
lwasser, line 1269: This endpoint [still exists in the docs as far as I can see](https://developers.strava.com/docs/reference/#api-SegmentEfforts-getSegmentEffortById).  But it could use more testing (I think?)

src/stravalib/client.py
```diff
@@ -1369,49 +1372,62 @@ def get_segment_efforts(
         """Gets all efforts on a particular segment sorted by start_date_local
 
         Returns an array of segment effort summary representations sorted by
-        start_date_local ascending or by elapsed_time if an athlete_id is
+        `start_date_local` ascending or by elapsed_time if an athlete_id is
         provided.
 
-        If no filtering parameters is provided all efforts for the segment
+        NOTES:
+        * The `athlete_id` sorting only works if you do NOT provide start and end dates.
```
lwasser, line 1379: This is undocumented but seems to work if you don't provide a start and end date. So it's a bit of a weird feature.

src/stravalib/client.py
```diff
@@ -1369,49 +1372,62 @@ def get_segment_efforts(
...
             a segment.
 
+        Notes
+        ------
+        https://developers.strava.com/docs/reference/#api-SegmentEfforts-getEffortsBySegmentId
+
         """
         params: dict[str, Any] = {"segment_id": segment_id}
 
         if athlete_id is not None:
```
lwasser, line 1429: Return a warning to let people know this could change at any time.

src/stravalib/client.py
```diff
@@ -1429,10 +1445,13 @@ def get_segment_efforts(
             )
 
         if limit is not None:
+            warn_param_deprecation(limit, "date ranges")
```
lwasser, line 1448: They have a per-page parameter but no limit. I noticed in our API that we use "limit" as a per-page parameter in a few places. But the way I read limit in the docs is it's a limit on how many items are returned. That feature doesn't work. Per page is still documented as a valid param.

src/stravalib/client.py
```diff
@@ -1429,10 +1445,13 @@ def get_segment_efforts(
...
             params["limit"] = limit
 
         result_fetcher = functools.partial(
-            self.protocol.get, "/segments/{segment_id}/all_efforts", **params
+            self.protocol.get,
```
lwasser, line 1452: This is the actual change here. it's pretty small and shouldn't impact our users at all!

src/stravalib/tests/integration/test_client.py
```diff
@@ -885,6 +888,35 @@ def test_get_segment_effort(mock_strava_api, client):
...
+    params, warning_type, warning_message, mock_strava_api, client
+):
+    """Test that if provided with deprecated params, the user receives a
+    warning."""
+
+    mock_strava_api.get(
+        "/segment_efforts", match=[matchers.query_param_matcher(params)]
+    )
+
+    with pytest.warns(warning_type, match=warning_message):
```
lwasser, line 916: I'm working in VS code, and this sometimes works. I can sometimes run the test and see the warning and it captures it if i'm just using my debugger. But it doesn't work if I run the entire test suite. 

I'm really confused as to why it's failing, and I would love any guidance on how to make sure it works properly.

lwasser:
This is the error I get, which looks like it's not matching the endpoint properly when trying to implement the pass-through. 

```
ERROR src/stravalib/tests/integration/test_client.py::test_get_segment_efforts_warnings[params0-DeprecationWarning-The '{athlete_id}' parameter is unsupported] - AssertionError: Not all requests have been executed [('GET', re.compile('https://www.strava.com/api/v3/segment_efforts'))]
```

lwasser:
Ahh ok this is good to go. The issue was I forgot about lazy loading. So I had to engage the iterator / use next to actually trigger an API call. hence the error. wow that was exceedingly tricky to debug and in the end was quite a simple issue.

src/stravalib/client.py
jsamoocha: This is the endpoint we _should_ be using, so nothing needs to change here.

src/stravalib/tests/integration/test_client.py
```diff
@@ -885,6 +891,52 @@ def test_get_segment_effort(mock_strava_api, client):
...
+    mock_strava_api.get("/segment_efforts", n_results=4)
+
+    efforts = client.get_segment_efforts(segment_id=2345, athlete_id=12345)
+
+    for index, effort in enumerate(efforts):
+        if index == 0:
+            name = effort.name
+
+    assert index + 1 == 4
+    assert name == "Alpe d'Huez"
```
jsamoocha, line 937:
```suggestion
    efforts = list(client.get_segment_efforts(segment_id=2345, athlete_id=12345))

    assert len(efforts) == 4
    assert efforts[0].name == "Alpe d'Huez"
```
I tried to make the test assertions more explicit here, but the test logic was already sound!

jsamoocha (approved):
Thanks for this fix! It prevents us from nasty surprises when Strava suddenly shuts down that undocumented endpoint. I added an (optional) rewrite of the test assertions. This can be merged!

src/stravalib/client.py
lwasser: Perfect!! now that I have a better grasp of our test mock setup and mocking in general, I may try to fill in some test gaps (but in a separate PR). Let's get this one merged!

src/stravalib/tests/integration/test_client.py
lwasser: Yay! thanks!! i see - .name doesn't really relate to the efforts endpoint so this is much better. merging now.

PR #605, opened 2024-11-13
feat: Return optional SummaryAthlete via exchange code for token with user request

### LINKED ISSUE #552: ENH: Exchange Code For Token to return Athelete summary
### Feature Type

### Problem Description

When calling `client.exchange_code_for_token` the response does not include the athlete summary details
However the strava api does provide an athelete summary in the token swap
Without this we have to call get Athlete thus incurring additional requests to Strava, and reducing overall rate limits

### Feature Description

Add the Athlete dictionary in the `exchange_code_for_token` response

### Alternative Solutions

call get Athlete but this consumes an apps rate limits with the strava api

### Additional Context

_No response_

### Code of Conduct

lwasser:
closes #552

## Description

This ensures `exchange_code_for_token` returns the optional `SummaryAthlete` object if requested.
## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [X] Yes

### DIFF
--- a/.github/.dependabot.yml
+++ b/.github/.dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
--- a/changelog.md
+++ b/changelog.md
@@ -5,7 +5,7 @@
 ### Added
 
 - Add: `SummaryAthlete` return to exchange_code_for_token (@lwasser, #552)
-- Add: Add more information about how our mock text fixture works (@lwasser, #292)
+- Add: Add more information about how our mock test fixture works (@lwasser, #292)
 
 ### Fixed
 - Fix: Update `segment_efforts` api and return warnings (@lwasser, #321)
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,12 @@
 
 ## Unreleased
 
-- Fix: Update segment_efforts api and return warnings (@lwasser, #321)
+### Fixed
+- Fix: Update `segment_efforts` api and return warnings (@lwasser, #321)
+
+### Added
+
+- Add: `SummaryAthlete` return to exchange_code_for_token (@lwasser, #552)
 
 ## v2.1.0
 
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -224,6 +224,7 @@ class Client:
             code=code,
             return_athlete=return_athlete,
         )
+        # This won't work because a dict has a length
         if len(raw) == 2:
             access_info, athlete_data = cast(
                 Tuple[AccessInfo, dict[str, Any]], raw
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -182,7 +182,7 @@ class Client:
         client_secret: str,
         code: str,
         return_athlete: bool = False,
-    ) -> AccessInfo | Tuple[AccessInfo, SummaryAthlete]:
+    ) -> AccessInfo | Tuple[AccessInfo, SummaryAthlete | None]:
         """Exchange the temporary authorization code (returned with redirect
         from Strava authorization URL) for a short-lived access token and a
         refresh token (used to obtain the next access token later on).
@@ -217,27 +217,21 @@ class Client:
         this exchange. However this return is currently undocumented
         and could change at any time.
         """
+        access_info, athlete_data = self.protocol.exchange_code_for_token(
+            client_id=client_id,
+            client_secret=client_secret,
+            code=code,
+            return_athlete=return_athlete,
+        )
+
+        # Return both access_info and athlete if requested. Athlete will be None if Strava
+        # doesn't return it in their end point response.
         if return_athlete:
-            access_info, athlete_data = self.protocol.exchange_code_for_token(
-                client_id=client_id,
-                client_secret=client_secret,
-                code=code,
-                return_athlete=return_athlete,
-            )
-        else:
-            access_info = self.protocol.exchange_code_for_token(
-                client_id=client_id,
-                client_secret=client_secret,
-                code=code,
-                return_athlete=return_athlete,
-            )
-
-        # Only return both if both exist, this will fail quietly if Strava
-        # modifies the end point return
-        if return_athlete and athlete_data:
-            summary_athlete = SummaryAthlete.model_validate(athlete_data)
+            if athlete_data:
+                summary_athlete = SummaryAthlete.model_validate(athlete_data)
+            else:
+                summary_athlete = None
             return access_info, summary_athlete
-
         else:
             return access_info
 
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -26,12 +26,11 @@ from typing import (
     TypeVar,
     cast,
 )
-from warnings import warn
 
 import arrow
 import pint
 import pytz
-from pydantic import BaseModel, ValidationError
+from pydantic import BaseModel
 from requests import Session
 
 from stravalib import exc, model, strava_model, unit_helper
@@ -209,7 +208,7 @@ class Client:
             token will expire)
         tuple
             Contains:
-             - the AccessInfo typed dict containing token values
+             - the `AccessInfo` typed dict containing token values
              - a `SummaryAthlete` object representing the authenticated user.
 
         Notes
@@ -218,33 +217,32 @@ class Client:
         this exchange. However this return is currently undocumented
         and could change at any time.
         """
-        raw = self.protocol.exchange_code_for_token(
-            client_id=client_id,
-            client_secret=client_secret,
-            code=code,
-            return_athlete=return_athlete,
-        )
-        # This won't work because a dict has a length
-        if len(raw) == 2:
-            access_info, athlete_data = cast(
-                Tuple[AccessInfo, dict[str, Any]], raw
-            )
         if return_athlete:
-            try:
-                athlete_info = SummaryAthlete.model_validate(athlete_data)
-                return access_info, athlete_info
-            except ValidationError as ve:
-                warn(
-                    f"Athlete data validation failed: {ve}. "
-                    "Returning AccessInfo without athlete info. "
-                    "Note: Returning athlete data is undocumented and could be "
-                    "dropped by Strava at any time.",
-                    UserWarning,
-                )
-                return access_info
+            access_info, athlete_data = self.protocol.exchange_code_for_token(
+                client_id=client_id,
+                client_secret=client_secret,
+                code=code,
+                return_athlete=return_athlete,
+            )
+        else:
+            access_info = self.protocol.exchange_code_for_token(
+                client_id=client_id,
+                client_secret=client_secret,
+                code=code,
+                return_athlete=return_athlete,
+            )
+
+        # # This won't work because a dict has a length
+        # if len(raw) == 2:
+        #     access_info, athlete_data = cast(
+        #         Tuple[AccessInfo, dict[str, Any]], raw
+        #     )
+        if return_athlete:
+            summary_athlete = SummaryAthlete.model_validate(athlete_data)
+            return access_info, summary_athlete
 
         else:
-            return cast(AccessInfo, raw)
+            return access_info
 
     def refresh_access_token(
         self, client_id: int, client_secret: str, refresh_token: str
--- a/src/stravalib/client.py
+++ b/src/stravalib/client.py
@@ -25,11 +25,12 @@ from typing import (
     TypeVar,
     cast,
 )
+from warnings import warn
 
 import arrow
 import pint
 import pytz
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 from requests import Session
 
 from stravalib import exc, model, strava_model, unit_helper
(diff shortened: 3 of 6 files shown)


.github/.dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
lwasser, line 1: Nothing has changed here it was just linted.

src/stravalib/tests/unit/test_protocol_auth.py
```diff
@@ -0,0 +1,70 @@
+from unittest.mock import patch
```
lwasser, line 1: Tests to ensure a return of either 1 or 2 items is provided. We could use this same structure to test refresh token IF it were useful.

src/stravalib/protocol.py
```diff
@@ -15,6 +15,7 @@
 import requests
 
 from stravalib import exc
+from stravalib.model import SummaryAthlete
```
jsamoocha, line 18: Letting the client method convert the raw response into a `SummaryAthlete` instead of here in the `protocol` module would be more consistent with how we handle other responses.

src/stravalib/tests/unit/test_protocol_auth.py
```diff
@@ -0,0 +1,70 @@
...
+
+    result = api_instance.exchange_code_for_token(
+        client_id=123,
+        client_secret="secret",
+        code="auth_code",
+        return_athlete=True,
+    )
+
+    # If athlete info is requested, return should be a tuple of 2 items
+    assert len(result) == 2
```
jsamoocha, line 52:
```suggestion
    exchange_response, athlete = api_instance.exchange_code_for_token(
        client_id=123,
        client_secret="secret",
        code="auth_code",
        return_athlete=True,
    )

    assert exchange_response["access_token"] == "mock_access_token"
    assert athlete["id"] == 12345
```
Making the assertion more explicit. In addition to the other comment, we should return the raw response instead of a `SummaryAthlete` object on this level. (Keep in mind that accepting this suggestion will -purposefully- break this test)

src/stravalib/protocol.py
```diff
@@ -200,7 +209,12 @@ def exchange_code_for_token(
             "expires_at": response["expires_at"],
         }
         self.access_token = response["access_token"]
-        return access_info
+
+        if return_athlete:
+            athlete_info = SummaryAthlete.model_validate(response["athlete"])
```
jsamoocha, line 214: This should happen in the client method

jsamoocha, line 214: Since `athlete` is undocumented, any response without an athlete would crash this with a `KeyError`. To prevent that, we could return `response.get("athlete")`. This method would then always return a tuple.

src/stravalib/tests/unit/test_protocol_auth.py
lwasser: OK, no problem. I'll fix the tests locally. I've refactored everything else locally based on your comments!

src/stravalib/client.py
```diff
@@ -189,17 +195,50 @@ def exchange_code_for_token(
...
-        )
+        raw = self.protocol.exchange_code_for_token(
+            client_id=client_id,
+            client_secret=client_secret,
+            code=code,
+            return_athlete=return_athlete,
+        )
+        if return_athlete:
+            try:
+                athlete_info = SummaryAthlete.model_validate(raw[1])
```
lwasser, line 228: @jsamoocha we don't seem to use ValidationError in other places that I can find. But I wasn't sure how to handle the case where athlete disappears in the return. If we use get (which is great) it returns None. So when it tries to validate this will fail. 

in this case we could either only return AccessInfo object OR we could raise a warning that the endpoint return has changed. I chose to raise a warning AND also return only the AccessInfo object.

Do you think this is handled appropriately or should I raise an error or handle this differently if the athlete return is None?   many thanks!

src/stravalib/client.py
```diff
@@ -189,17 +195,50 @@ def exchange_code_for_token(
...
-            client_id=client_id, client_secret=client_secret, code=code
-        )
+        raw = self.protocol.exchange_code_for_token(
+            client_id=client_id,
+            client_secret=client_secret,
+            code=code,
+            return_athlete=return_athlete,
+        )
+        if return_athlete:
+            try:
```
lwasser, line 227: I moved the validation / `SummaryAthlete` model population step here.

src/stravalib/client.py
```diff
@@ -189,17 +195,50 @@ def exchange_code_for_token(
...
+                return AccessInfo, athlete_info
+            except ValidationError as ve:
+                warn(
+                    "Oops, I can't seem to find athlete data in the return."
+                    "I will return the AccessInfo without athlete info.",
+                    "Please note that returning athlete data is undocumented.",
+                    "Support could be dropped by Strava at any time",
+                    f"The full error is here: {ve}",
+                )
+                return AccessInfo
```
lwasser, line 238: This return & associated warning is what I'm questioning!

src/stravalib/protocol.py
```diff
@@ -184,6 +192,7 @@ def exchange_code_for_token(
             expires_at (number of seconds since Epoch when the provided
             access token will expire)
         """
+        # The method returns: No rates present in response headers
```
lwasser, line 195: i'll hunt this down later but it relates to a discussion thread. it's just a message that's returned and isn't relevant to this PR. I'm leaving a note for later to be able to pinpoint where this is coming from.

src/stravalib/protocol.py
```diff
@@ -200,7 +209,12 @@ def exchange_code_for_token(
             "expires_at": response["expires_at"],
         }
         self.access_token = response["access_token"]
-        return access_info
+        if return_athlete:
+            # This will return None if the athlete key is missing so it doesn't
+            # fail if the undocumented part of the response is removed
```
lwasser, line 214:
```suggestion
            # This will potentially None if the undocumented athlete response is removed from the endpoint by Strava 
```

src/stravalib/protocol.py
lwasser: I always forget about .get when using dictionaries. Much better call!

src/stravalib/client.py
```diff
@@ -189,17 +196,54 @@ def exchange_code_for_token(
...
-        return self.protocol.exchange_code_for_token(
-            client_id=client_id, client_secret=client_secret, code=code
+        raw = self.protocol.exchange_code_for_token(
+            client_id=client_id,
+            client_secret=client_secret,
+            code=code,
+            return_athlete=return_athlete,
         )
+        if len(raw) == 2:
+            access_info, athlete_data = cast(
```
lwasser, line 228: mypy was getting confused by the tuple return with a dictionary that was typed.

src/stravalib/tests/integration/test_client.py
```diff
@@ -49,6 +50,32 @@ def default_request_params():
...
+            "state": "State",
+            "country": "Country",
+            "sex": "F",
+            "premium": True,
+            "summit": True,
+        },
+    )
+
+
 def test_get_athlete(mock_strava_api, client):
```
lwasser, line 79: These tests should fully cover this method. oddly code cov isn't updating or recognizing this. but they should test all three scenarios - athlete return missing as strava removes it, athlete return as expected, no athlete data return (default behavior)

lwasser:
ok yay - updating the action was the issue. the diff is now as expected all lines are covered. this is ready for another review 🚀 and I'm happy to fix other things that aren't quite right in both tests and approach!

src/stravalib/client.py
```diff
@@ -189,17 +196,55 @@ def exchange_code_for_token(
...
+
+        Notes
+        -----
+        Strava by default returns `SummaryAthlete` information during
+        this exchange. However this return is currently undocumented
+        and could change at any time.
         """
-        return self.protocol.exchange_code_for_token(
-            client_id=client_id, client_secret=client_secret, code=code
+        raw = self.protocol.exchange_code_for_token(
```
jsamoocha, line 221: Can you try to unpack the return here? So
```python
access_info, athlete_data = self.protocol.exchange_code_for_token()
```
You can then return the tuple if `athlete_data and return_athlete == True`. This prevents the check on `len(raw)` and handling the `ValidationError` in case of a `None` response.

src/stravalib/protocol.py
jsamoocha: I think I read somewhere that Strava doesn't count authentication requests re. rate limits, so that may be the edge case that leads to us logging a warning. We can finetune the rate limiter to ignore authentication calls to prevent that warning from being logged.

src/stravalib/protocol.py
```diff
@@ -162,8 +162,12 @@ def authorization_url(
...
 
     def exchange_code_for_token(
-        self, client_id: int, client_secret: str, code: str
-    ) -> AccessInfo:
+        self,
+        client_id: int,
+        client_secret: str,
+        code: str,
+        return_athlete: bool | None = False,
+    ) -> AccessInfo | tuple[AccessInfo, dict[str, Any]]:
```
jsamoocha, line 170:
```suggestion
    ) -> AccessInfo | tuple[AccessInfo, dict[str, Any] | None]:
```
I forgot that the athlete may be None

src/stravalib/tests/integration/test_client.py
```diff
@@ -1082,3 +1109,60 @@ def test_get_activity_kudos(mock_strava_api, client):
...
+
+@patch("stravalib.protocol.ApiV3.exchange_code_for_token")
+def test_exchange_code_for_token_athlete(
+    mock_request, client, raw_exchange_response
+):
+    """If a user requests athlete data, then it should be returned as a
+    SummaryAthlete object."""
+    mock_request.return_value = raw_exchange_response
+
+    result = client.exchange_code_for_token(
```
jsamoocha, line 1122: Can you also try to unpack the `result` tuple here (and adapt the assertions below) to make the test more explicit?

src/stravalib/tests/unit/test_protocol_auth.py
```diff
@@ -0,0 +1,68 @@
...
+    api_instance = ApiV3()
+    mock_request.return_value = mock_token_exchange_response
+
+    result = api_instance.exchange_code_for_token(
+        client_id=123,
+        client_secret="secret",
+        code="auth_code",
+    )
+
+    assert len(result) == 3
```
jsamoocha, line 67: This assertion can be removed as the result is a dict.

jsamoocha (changes requested):
Great, almost there! Can you try making the code more explicit by unpacking tuple responses (see the inline comments)?

src/stravalib/protocol.py
lwasser: ok i'm resolving things that i've addressed!

src/stravalib/tests/integration/test_client.py
lwasser: Ahhhh. I really appreciate this feedback. Thank you! This test is much cleaner / better now as is the code. I was trying to avoid additional conditional statements in the methods, but it seems by doing that I added complexity and less specificity to the tests and code.

src/stravalib/client.py
```diff
@@ -189,17 +195,51 @@ def exchange_code_for_token(
...
+        Notes
+        -----
+        Strava by default returns `SummaryAthlete` information during
+        this exchange. However this return is currently undocumented
+        and could change at any time.
         """
-        return self.protocol.exchange_code_for_token(
-            client_id=client_id, client_secret=client_secret, code=code
-        )
+        if return_athlete:
```
lwasser, line 220: This is much cleaner. However, mypy is not confused about protocol returning an "iterable". I'm not sure how to make it happy and could use any help @jsamoocha @enadeau . thank you!  I tried to "cast" it to a tuple but that didn't work.

src/stravalib/protocol.py
lwasser: ok perfect. I'll turn this into an issue, and we can handle it separately!

lwasser:
Okay, I've made more progress, and this can be reviewed again. 

I think this iteration is much cleaner. However, now I'm getting my errors because it doesn't "see" that `protocol.echange_token` ... is returning a Tuple / iterable. I'm not sure how to handle this. I tried to cast it, but that didn't work. To me, the typing looks ok, but I'm also not a typing expert.

changelog.md
```diff
@@ -4,11 +4,11 @@
 
 ### Added
 
-- Add: Add more information about how our mock fixture works (@lwasser, #292)
+- Add: `SummaryAthlete` return to exchange_code_for_token (@lwasser, #552)
+- Add: Add more information about how our mock text fixture works (@lwasser, #292)
```
lwasser, line 8:
```suggestion
- Add: Add more information about how our mock test fixture works (@lwasser, #292)
```

src/stravalib/protocol.py
```diff
@@ -166,8 +166,8 @@ def exchange_code_for_token(
         client_id: int,
         client_secret: str,
         code: str,
-        return_athlete: bool | None = False,
-    ) -> AccessInfo | tuple[AccessInfo, dict[str, Any]]:
+        return_athlete: bool = False,
+    ) -> tuple[AccessInfo, dict[str, Any] | None]:
```
jsamoocha, line 170: The typing issue might have been caused by some ambiguity in the previous version's signature. The typing issue was resolved by making `protocol.exchange_code_for_token()` always return a tuple. The athlete part of the returned tuple will be None if not requested or not present in Strava's response.

jsamoocha (approved):
I fixed the typing issue, and everything else looks fine. This can be merged now. (I haven't tested this change on the real Strava API, assuming you already did. If not, let's do that before merging.)

src/stravalib/protocol.py
lwasser: @jsamoocha thank you. I had tested this locally. I'll test again because I took some time off GitHub completely and haven't looked at this pr in a while. Then I'll merge. Thank you for figuring this out. I was stumped by it. 

Happy happy new year!! 🚀

src/stravalib/client.py
lwasser: this is complete resolving!

lwasser:
Ok this all looks good locally!! i'm going to merge it now 🚀

PR #606, opened 2024-11-13
Add: dependabot to update actions

lwasser:
I'm just getting this up because our actions keep getting out of date. so this will make those node deprecation warnings go away as it updates things for us.


lwasser:
pre-commit autofix

PR #612, opened 2025-01-12
First stab at updating token

### LINKED ISSUE #585: ENH: built-in token refresh
### Feature Type

### Problem Description

It would be nice if stravalib would automatically refresh access tokens whenever they expire.

### Feature Description

Calling a client method will automatically refresh an access token if it's expired (or about to expire).

### Alternative Solutions

Letting users handle token expiry themselves (which is the current situation)

### Additional Context

_No response_

### Code of Conduct

lwasser:
closes #585

## Description

This is a stab at an automatic token refresh built into stravalib using environment variables if the user wishes to do so. it should NOT impact current functionality. 

My notes are below regarding how this works. Once we are happy with it I will add docs as well. 

This workflow assumes that you can access your refresh token and token 
expiration date. This information is created the first time that you 
authenticate a user with Strava. 

The basic authentication information you need:

* **access_token**: this token value expires after 6 hours
* **refresh_token**: this token is needed to refresh the access_token after it expires
* **expires_at**: this is the date/time when your access time expires and you need 
to refresh it

The access token that you initially retrieve can be used for the full period it's 
active. Once it expires, you will need to use the refresh token to refresh it so you 
can continue to make API calls. 

The expires_at value allows stravalib to check to see if your token is active or 
not before making API calls allowing stravalib ti automatically refresh your token 
when needed. 

How this works... 

1. Create the .env file with the client secret and id 
2. ake sure python-dotenv is installed (optional dependency)
3. Populate the .env file with the client secret and ID values:

CLIENT_SECRET=
CLIENT_ID=

4. the rest will be done for you!
## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [x] Yes

I'M GOING TO WORK ON TESTS NEXT. right now this pr breaks everything but I'll slowly fix it. I wanted folks to see this approach in case there is feedback as I'm working!

- [X] Yes, `changelog.md` is up-to-date.

all coming soon...


README.md
```diff
@@ -157,6 +153,42 @@ if time.time() > client.token_expires_at:
     expires_at = refresh_response["expires_at"]
 ```
 
+## Automatic token refresh
```
lwasser, line 156: I think it would be nice to have a tiny quick start and then we can link to our docs so we aren't trying to maintain examples in two places!

src/stravalib/protocol.py
```diff
@@ -8,11 +8,14 @@
...
 import functools
 import logging
+import os
+import time
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Literal, TypedDict
 from urllib.parse import urlencode, urljoin, urlunsplit
 
 import requests
+from dotenv import load_dotenv
```
lwasser, line 18: Note: this is a new dependency, but we could make it optional. Users can always bypass dotenv and just set envt variables themselves too.

lwasser:
Anyone is welcome to test this out. i have to fix all of the mypy and test issues next but the functionality (I think) works. but i'd love some folks to test!! 🚀  I've moved all messaging to logging rather than print statements following @jsamoocha feedback example decorator!

src/stravalib/protocol.py
```diff
@@ -89,6 +102,85 @@ def __init__(
...
+        except ImportError:
+            logging.debug(
+                "python-dotenv is not installed. Skipping .env file loading."
+                " If you need .env support, please install `python-dotenv`."
+            )
+        # os.environ.get returns strings by default
+        client_id = os.environ.get("CLIENT_ID")
+        client_secret = os.environ.get("CLIENT_SECRET")
+
+        if client_id and client_secret:
```
lwasser, line 157: I'm stuck here on typing and how to ensure client_id is an int. it is possible that a user could write the wrong value to their envt or env file. mypy however is giving me errors .

src/stravalib/protocol.py
```diff
@@ -89,6 +102,168 @@ def __init__(
...
+            load_dotenv()
+            logging.info("Loaded .env file successfully.")
+        except ImportError:
+            logging.debug(
+                "python-dotenv is not installed. Skipping .env file loading."
+                " If you need .env support, please install `python-dotenv`."
+            )
+        # os.environ.get returns strings by default
+        client_id = int(os.environ.get("CLIENT_ID"))
+        client_secret = os.environ.get("CLIENT_SECRET")
```
jsamoocha, line 239: Maybe rename these to `STRAVA_CLIENT_ID` and `STRAVA_CLIENT_SECRET`? In larger apps, there will be dozens of env variables.

src/stravalib/protocol.py
```diff
@@ -89,6 +102,168 @@ def __init__(
...
+            # https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
+            from dotenv import load_dotenv  # type: ignore
+
+            load_dotenv()
+            logging.info("Loaded .env file successfully.")
+        except ImportError:
+            logging.debug(
+                "python-dotenv is not installed. Skipping .env file loading."
+                " If you need .env support, please install `python-dotenv`."
+            )
```
jsamoocha, line 236: I think (in hindsight?) we should entirely leave out any possible .env file processing. Larger apps would include `stravalib` as a dependency and have their own management of development/test/production environments and would use `load_dotenv` on a higher level.

For smaller hobby projects, we could add references to our documentation on how to use .env files + pydotenv (there are plenty of great tutorials online).

The only thing our code should do is assume these env variables are set in whichever way the user prefers.

src/stravalib/client.py
```diff
@@ -67,6 +67,8 @@ def __init__(
         rate_limit_requests: bool = True,
         rate_limiter: limiter.RateLimiter | None = None,
         requests_session: Session | None = None,
+        token_expires: int | None = None,
+        token_refresh: str | None = None,
```
jsamoocha, line 71: Let's rename this throughout to `refresh_token` to retain consistency with Strava's naming and our naming of `access_token`.

src/stravalib/protocol.py
```diff
@@ -89,6 +102,168 @@ def __init__(
...
+            Whether to automagically refresh the users token or not.
+            Requires environment setup to work.
+
+        Returns
+        -------
+        Dict[str, Any]
+            The parsed JSON response.
+        """
+        # Try to refresh the token unless told to not try
+        if refresh_expired_token:
```
jsamoocha, line 144: To prevent that infinite loop, could we check the URL at this point and ignore refreshing if the URL itself ends with `/oauth/token`? In that case, we could leave the existing `exchange_code_for_token()` method unchanged.

jsamoocha (commented):
Thanks! Almost there, I left a few comments for improvement.

src/stravalib/protocol.py
lwasser: Good call! renamed!

src/stravalib/protocol.py
lwasser: That's a good point. I struggled with this because i didn't like adding an (optional) dependency. this will make the code much cleaner as we only need the os library to access envt variables. 

I have a doc locally where i'm building out the steps so I'll just update our tutorials accordingly and we can link to the dot file tutorials as well. It's pretty simple to setup and now I don't have to worry about tests for it!

src/stravalib/protocol.py
lwasser: Ahhhh so much better. One less parameter as well in this method. Thanks!!

src/stravalib/client.py
lwasser: I've done this throughout! Are you comfortable with keeping `token_expires` as that makes it a verb?

lwasser:
I will come back to this and work more on tests and mypy. but the basic code is working and I've addressed all comments in the review so far! (I think?)

README.md
lwasser: Update: in #613 i just linked to the docs to avoid maintaining instructions in several places. so I'll rebase this pr after we review merge that other pr - Or I'll rebase that one against this if this gets merged first. but the goal is to not dump all docs into the readme rather keep them confined to the docs and link to them.

src/stravalib/protocol.py
```diff
@@ -89,6 +102,187 @@ def __init__(
...
+
+        # If the request is a token request, skip, otherwise
+        # refresh token
+        # This feels sloppy but if check credentials is None then that means
+        # users environment isn't setup
+        credentials = self._check_credentials()
+
+        # Only refresh token if we know the users' environment is setup
+        if "/oauth/token" not in url and credentials:
+            self.refresh_expired_token(credentials[0], credentials[1])
```
lwasser, line 184: This is what i don't like. I wanted to unpack a tuple with client_id, client_secret. It would make the code more expressive. The alternative would be that the tuple contains a third value that is a bool for valid credentials? that would make that method a bit more complex but would make the code here more expressive.

README.md
```diff
@@ -157,6 +153,42 @@ if time.time() > client.token_expires_at:
...
+
+The above example shows you how to manually authenticate with Strava using
+stravalib. It then shows you how to refresh your token manually.
+
+If you want, you can  automate the refresh token process by setting up your
+environment with your Strava app's `clientID` and `clientSecret` values.
+
+To do this add the following values to your environment yourself.
+
+### Using the dot file approach
```
lwasser, line 169: Remove this given we've adjusted the approach.

pyproject.toml
```diff
@@ -62,7 +62,10 @@ build = ["build"]
 tests = [
   "pytest",
   "pytest-cov",
-  "responses"
+  # Run tests in parallel
+  "pytest-xdist",
```
lwasser, line 66: this supports distributed processing if available.

src/stravalib/client.py
```diff
@@ -111,21 +121,63 @@ def access_token(self) -> str | None:
...
+             User's access token
 
         Returns
         -------
 
         """
         self.protocol.access_token = token_value
 
+    # TODO: if we go this route create a function to set up these attrs
+    @property
```
lwasser, line 139:
```suggestion
    @property
```

src/stravalib/protocol.py
```diff
@@ -89,6 +102,190 @@ def __init__(
             lambda _request_params, _method: None
         )
 
+    def _check_credentials(self) -> tuple[int, str] | None:
```
lwasser, line 105: This is a new helper that checks credentials. This makes it easier to test and hopefully maintain and avoids adding a lot of logic to the refresh steps.

src/stravalib/protocol.py
```diff
@@ -89,6 +102,190 @@ def __init__(
...
+        Returns
+        -------
+        Dict[str, Any]
+            The parsed JSON response.
+        """
+
+        # If the request is a token request, skip, otherwise
+        # refresh token
+        # This feels sloppy but if check credentials is None then that means
+        # users environment isn't setup
```
lwasser, line 182:
```suggestion
        # Check if refresh token credentials are setup
```

src/stravalib/protocol.py
```diff
@@ -264,77 +471,6 @@ def resolve_url(self, url: str) -> str:
             )
         return url
 
-    def _request(
```
lwasser, line 267: i didn't delete or modify much in this method. It looks that way because of the new methods and where I added them.

src/stravalib/tests/unit/test_protocol_ApiV3_token_refresh.py
```diff
@@ -0,0 +1,105 @@
...
+
+import stravalib
+from stravalib.protocol import ApiV3
+
+
+@pytest.fixture
+def apiv3_instance():
+    """Fixture to create an ApiV3 instance for testing."""
+
+    return ApiV3()
```
lwasser, line 15:
```suggestion
```

moved to conftest

src/stravalib/tests/unit/test_protocol_ApiV3_token_refresh.py
```diff
@@ -0,0 +1,105 @@
...
+        client_id=12345,
+        client_secret="secret",
+        refresh_token="refresh_value123",
+    )
+    # Should return none
+    assert not out
+
+
+def test_request_skips_refresh_for_oauth_token(apiv3_instance):
+    """Test that _request skips refresh_expired_token for /oauth/token URL."""
```
lwasser, line 90: I had to mock a lot here to isolate the single URL check. could I use our mock API object? i haven't dug into that yet.

src/stravalib/tests/unit/test_refresh_token.py
```diff
@@ -0,0 +1,123 @@
...
+exchange_code_for_token
+* make sure that access token and token expires are set properly in the
+ - protocol.py:360-370 & 414-419 (how are these different?)
+
+ ## Client.py
+ * in client - make sure the access token, refresh value and expires at are all
+  set client.py:120-179
+* refresh access token: 258
+i think these will be integration tests to add to client?
+vs unit
```
lwasser, line 15:
```suggestion
```

src/stravalib/tests/unit/test_refresh_token.py
```diff
@@ -0,0 +1,123 @@
...
+    and not called when valid.
+    """
+
+    apiv3_instance.refresh_token = "dummy_refresh_token"
+
+    # Patch os get envt variables and populate envt vals
+    with patch.dict(
+        os.environ,
+        {"STRAVA_CLIENT_ID": "12345", "STRAVA_CLIENT_SECRET": "secret"},
+    ):
```
lwasser, line 73:
```suggestion
```

this context manager is no longer needed because it only is needed in the _check_credentials helper (I think)

src/stravalib/tests/unit/test_refresh_token.py
```diff
@@ -0,0 +1,123 @@
...
+        apiv3_instance._token_expired = MagicMock(return_value=token_expired)
+        apiv3_instance.refresh_access_token = MagicMock()
+
+        apiv3_instance.refresh_expired_token(
+            client_id=12345, client_secret="123secret"
+        )
+
+        # Check how many times refresh_access_token was called
+        assert apiv3_instance.refresh_access_token.call_count == expected_call
```
lwasser, line 82:
```suggestion
        assert apiv3_instance.refresh_access_token.call_count == expected_call
```

lwasser:
Ok this is ready for another review and I will pause working on things. 
✅ I added tests but am uncertain whether what I added is sufficient. I focused on unit tests that isolate specific behaviors. 
✅ I cleaned up the feature code following the review. 

All feedback welcome! 🎉

src/stravalib/protocol.py
```diff
@@ -203,62 +242,42 @@ def _token_expired(self) -> bool:
...
-                    # None
-                    refresh_token=self.token_refresh,
-                )
-        # The user hasn't setup their environment
         else:
             logging.warning(
-                "Client ID and secret not found. You need to manually refresh"
-                " your token or set up environment variables."
+                "Please make sure you've set up your environment."
+                " I can't find your refresh_token"
```
jsamoocha, line 280: This warning message may be confusing as the refresh token is not coming from the environment, and the warning may also be triggered if the refresh token is present but has not expired yet. We may need separate warnings for each case.

src/stravalib/protocol.py
```diff
@@ -131,18 +169,19 @@ def _request(
...
         -------
         Dict[str, Any]
             The parsed JSON response.
         """
-        # Try to refresh the token unless told to not try
-        if refresh_expired_token:
-            self.refresh_expired_token()
+
+        # Check if refresh token credentials are setup
+        credentials = self._check_credentials()
```
jsamoocha, line 180: It's great that you refactored this 👍

However, it may be better to move this check to `APIV3`s initializer. Otherwise, this check (and its associated warning/error messages) will be performed on every request.

src/stravalib/protocol.py
jsamoocha: If you move the check and extraction of credentials to `APIV3`s initializer, this should be no problem anymore.

src/stravalib/protocol.py
jsamoocha: Great refactoring. We can simplify this by always returning a tuple of potentially nullables (so the return type is `tuple[int | None, str | None]`) and storing the result as instance variables of `APIV3` for later reuse. See the other comments.

jsamoocha (changes requested):
Thanks for these changes! There is one remaining issue about the environment being checked for credentials on every request. It can be fixed by moving the check to the `APIV3` initializer.

src/stravalib/protocol.py
```diff
@@ -88,6 +103,175 @@ def __init__(
...
+                # Make sure client_id is a valid int
+                client_id: int = int(client_id_str)
+            except ValueError:
+                logging.error("STRAVA_CLIENT_ID must be a valid integer.")
+        else:
+            logging.error(
+                "Please make sure your STRAVA_CLIENT_ID is set in your environment."
+            )
+
+        if client_id and client_secret:
```
lwasser, line 137: Populate credentials to self if they exist and are in the correct format.

src/stravalib/tests/conftest.py
```diff
@@ -18,3 +21,26 @@ def mock_strava_api():
...
+    return ApiV3(
+        access_token="dummy_access_token",
+        client_id=123456,
+        client_secret="clientfoosecret",
+        refresh_token="refresh_value123",
+    )
+
+
+@pytest.fixture
+def mock_strava_env():
```
lwasser, line 39: vscode will STORE environment variables like this that are mocked. This was breaking things and I couldn't figure out why. This fixture allows us to mock that piece by just calling it.

src/stravalib/protocol.py
```diff
@@ -74,12 +80,27 @@ def __init__(
...
+            client secret for the Strava app pulled from the users envt
         """
         self.log = logging.getLogger(
             "{0.__module__}.{0.__name__}".format(self.__class__)
         )
-        self.access_token = access_token
+        self.access_token: str | None = access_token
+        self.token_expires: int | None = token_expires
+        self.refresh_token: str | None = refresh_token
+        self.client_id: int | None = None
```
lwasser, line 102: This two vars as now populated by the method _check_credentials which avoids having to pass things around and perform continual checks!

src/stravalib/protocol.py
```diff
@@ -88,6 +109,189 @@ def __init__(
         self.rate_limiter = rate_limiter or (
             lambda _request_params, _method: None
         )
+        # Check for credentials when initializing
+        self._check_credentials()
```
lwasser, line 113: Runs on init... if credentials exist, populate to self.

src/stravalib/protocol.py
```diff
@@ -88,6 +109,189 @@ def __init__(
...
+            Whether to raise an error or not.
+
+        Returns
+        -------
+        Dict[str, Any]
+            The parsed JSON response.
+        """
+
+        # Only refresh token if we know the users' environment is setup
+        if "/oauth/token" not in url and self.client_id and self.client_secret:
```
lwasser, line 191: The only things changed in _request is this check

src/stravalib/protocol.py
```diff
@@ -88,6 +109,189 @@ def __init__(
...
+                "the auto token-refresh feature"
+            )
+            return
+
+        # Token is not yet expired, move on
+        if not self._token_expired():
+            return
+
+        # This should never be false, BUT mypy wants a reminder that these values
+        # are populated
```
lwasser, line 281: Double check this - it made mypy happy but I wonder if we need test coverage or if this is just a mypy work around.

noxfile.py
```diff
@@ -5,7 +5,7 @@
 
 import nox
 
-nox.options.reuse_existing_virtualenvs = False
+nox.options.reuse_existing_virtualenvs = True
```
lwasser, line 8:
```suggestion
nox.options.reuse_existing_virtualenvs = False
```
mistakenly committed this. we want these to rebuild but I've been running tests over and over so there's no point in rebuilding  every time

src/stravalib/protocol.py
```diff
@@ -238,6 +445,12 @@ def refresh_access_token(
             Dictionary containing the access_token, refresh_token and
             expires_at (number of seconds since Epoch when the provided
             access token will expire)
+
+        Notes
+        -----
+        This method is user facing so we don't populate client_id and
+        client_secret from self in case user wish to manually populate
+        those values.
```
lwasser, line 453:
```suggestion
        This method is user facing. Here, we don't populate client_id and
        client_secret from self; A user can call this method and refresh the token manually.
        those values.
```

src/stravalib/protocol.py
lwasser: Ok this has been moved to init. it's much better now. I'm really glad that you caught that. I'm refactoring tests a bit and I'll look at that comment above.

src/stravalib/protocol.py
```diff
@@ -88,6 +109,189 @@ def __init__(
...
+        if the user has setup their environment with the client
+        secret information.
+
+        Returns
+        -------
+        None
+            If all is setup properly, updates and resets the access_token
+            attribute. Otherwise it logs a warning
+        """
+        if not self.refresh_token:
```
lwasser, line 269: This now explicetly provides the user with feedback if they want to use token_refresh. hopefully it's more clear than before!

src/stravalib/protocol.py
lwasser: I don't see a way around adding these asserts to make mypy happy. but client_id should never be missing here as it's checked before this method is even run.

src/stravalib/protocol.py
lwasser: fixed - i cleaned that method up and added more explicit feedback if refresh_token isn't populated.

src/stravalib/protocol.py
lwasser: it's much better! thank you for catching that. it just gets assigned to self now which makes everything simpler.

lwasser:
Ok this is ready for another review when you have time. I've cleaned up the tests some and addressed the review comments ✨

src/stravalib/protocol.py
```diff
@@ -238,6 +445,12 @@ def refresh_access_token(
             Dictionary containing the access_token, refresh_token and
             expires_at (number of seconds since Epoch when the provided
             access token will expire)
+
+        Notes
+        -----
+        This method is user facing. Here, we don't populate client_id and
+        client_secret from self; A user can call this method and refresh the token manually.
```
jsamoocha, line 452:
```suggestion
        client_secret from self; A user can call this method and refresh the token manually with
```

jsamoocha (approved):
Great stuff! I found a typo in a docstring somewhere; otherwise, this can be merged.

lwasser:
yay thanks! i'll merge it now. Should we push a release with the new features?

jsamoocha:
A release sounds good. It will allow me to reduce some complexity in code at my end :-)

lwasser:
awesome. ok i'll merge now. And i'll do a release in the next few days 🚀

PR #619, opened 2025-02-08
fix: add trusted publisher setup to stravalib

### LINKED ISSUE #617: Feat: update release workflow to follow current security recommendations
We should do the following

- [ ] separate build from deploy
- [ ] create environments and then add permissions so a human (one of our maintainers) has to ok the trigger of the deploy step always
- [ ] add Jonatan and myself as authorized to release which means the build for the package will run but then we will have to authorize the deploy manually! I'll document it but it will just make everything more secure (this can't be done until we merge to main and github recognizes the environments. Right now it's only on my branch and in a pr

maybe sigstore signature too but I'm less familiar with that step. 

This also might entail adjusting our trusted publisher setup.
An example of what this will look like [is here](https://github.com/pyOpenSci/pyosMeta/blob/main/.github/workflows/publish-pypi.yml)
I'm noting that we dont' use environments at all right now which means we should update trusted publisher too

lwasser:
closes #617

## Description

This pr describes and adds trusted publishing to stravalib. Right now we are publishing via a token. This is ok but token leaks are real AND trusted publishing is a more secure and preferred route to publishing. 

A few things left to do here

- [ ] document the slight change in the workflow - 
- [x] Setup trusted publisher in pypi

Screenshot 2025-02-08 at 2 05 05 PM- 

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/.github/workflows/push-pypi.yml
+++ b/.github/workflows/push-pypi.yml
@@ -6,10 +6,11 @@ on:
     branches:
       - main
 jobs:
-  build-publish:
+  build:
     runs-on: ubuntu-latest
     # Only run build action on base repo - not forks
     if: github.repository_owner == 'stravalib'
+    environment: build
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


lwasser:
I will merge this because I have to test it on main before deleting our token. I also can't document it until the environments are setup so I can add screenshots and such.

PR #643, opened 2025-05-30
ENH: bypass client ID & client secret error and warning when `silence_token_warning` is true

### LINKED ISSUE #625: ENH: Don't show message about client ID and secret environment variables when using other workflows
### Feature Type

- [x] Changing existing functionality in stravalib

### Problem Description

Using Stravalib 2.2, when doing a manual token exchange, there is now a pair of messages telling people to set environment variables.

```
ERROR:root:Please make sure your STRAVA_CLIENT_ID is set in your environment.
WARNING:root:STRAVA_CLIENT_ID and STRAVA_CLIENT_SECRET not found in your  environment. Please refresh your access_token 
```

This is useful functionality, but it interferes with uses where the client ID and secret are handled in a different way.

### Feature Description

Perhaps the better way to do this would be to only show it if a request fails due to an outdated token?

At the very least, this should be clearly documented in the migration guide.

Maybe there could be a way to opt out of this specifically somehow to never see the message. 

### Alternative Solutions

See above.

### Additional Context

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

KevsterAmp:
closes #625

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/protocol.py
+++ b/src/stravalib/protocol.py
@@ -131,8 +131,10 @@ class ApiV3(metaclass=abc.ABCMeta):
 
         client_id_str = os.environ.get("STRAVA_CLIENT_ID")
         client_secret = os.environ.get("STRAVA_CLIENT_SECRET")
+        silence_token_warning = os.environ.get("SILENCE_TOKEN_WARNINGS")
+
         # Make sure client_id exists and can be cast to int
-        if client_id_str:
+        if client_id_str or silence_token_warning:
             try:
                 # Make sure client_id is a valid int
                 client_id = int(client_id_str)
@@ -143,7 +145,7 @@ class ApiV3(metaclass=abc.ABCMeta):
                 "Please make sure your STRAVA_CLIENT_ID is set in your environment."
             )
 
-        if client_id and client_secret:
+        if (client_id and client_secret) or silence_token_warning:
             self.client_id = client_id
             self.client_secret = client_secret
         else:
@@ -277,7 +279,6 @@ class ApiV3(metaclass=abc.ABCMeta):
         assert (
             self.client_secret is not None
         ), "client_secret is required but is None."
-
 
         self.refresh_access_token(
             client_id=self.client_id,


src/stravalib/protocol.py
```diff
@@ -131,8 +131,10 @@ def _check_credentials(self) -> None:
...
         client_id_str = os.environ.get("STRAVA_CLIENT_ID")
         client_secret = os.environ.get("STRAVA_CLIENT_SECRET")
+        silence_token_warning = os.environ.get("SILENCE_TOKEN_WARNINGS")
+
         # Make sure client_id exists and can be cast to int
-        if client_id_str:
+        if client_id_str or silence_token_warning:
             try:
                 # Make sure client_id is a valid int
                 client_id = int(client_id_str)
```
KevsterAmp, line 140: Failed CI was caused by this.

I'm not exactly sure how to proceed since client_id is not initialized anywhere else 
https://github.com/stravalib/stravalib/blob/8cb1fc7daed8177fe81321f42a080a9330978706/src/stravalib/protocol.py#L129-L130

So I reached out on the issue about how they manually exchange tokens.
https://github.com/stravalib/stravalib/issues/625#issuecomment-2921992276

KevsterAmp:
tests and channelog will be added once I get a clear idea on how to resolve the issue

src/stravalib/protocol.py
jsamoocha: The type checker now fails because `client_id_str` is not guaranteed to have a value. Can you try fixing this by moving the check for `silence_token_warning` to the exact spot where it's needed? I.e.:
```python
        if client_id_str:
            try:
                # Make sure client_id is a valid int
                client_id = int(client_id_str)
            except ValueError:
                logging.error("STRAVA_CLIENT_ID must be a valid integer.")
        elif not silence_token_warning:  # <--
            logging.error(
                "Please make sure your STRAVA_CLIENT_ID is set in your environment."
            )

jsamoocha:
Thanks for working on this issue @KevsterAmp ! Can you also add text to the error + warning messages that these can be disabled by setting the environment variable?

KevsterAmp:
I'm not exactly sure how to add tests for this as we are using `MagicMock()` for testing `client.refresh_access_token()`

KevsterAmp:
I'll try to take a look at the tests and add one for this PR. Thanks for the assistance 

cc: @jsamoocha  @lwasser

lwasser:
@jsamoocha i agree. Lets merge as is and cut a  release. 

We can always add docs later as needed. @KevsterAmp thank you for all of the work on the pr!! We appreciate you!

PR #645, opened 2025-06-18
chore(ci): pin action versions to hashes

miketheman:
## Description

Increase repository security by pinning GitHub Actions to hashes, to minimize the ability for the underlying author to change the tag of the release without the user knowing about it.

Surfaced via zizmor
Ref: https://docs.zizmor.sh/audits/#unpinned-uses

## Type of change

## Does your PR include tests

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -12,9 +12,9 @@ jobs:
     env:
       PYTHON-VERSION: "3.11"
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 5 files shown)


lwasser:
@all-contributors please add @miketheman for code, review, security

.github/workflows/build-docs.yml
```diff
@@ -12,9 +12,9 @@ jobs:
     env:
       PYTHON-VERSION: "3.11"
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
lwasser, line 15: Pin dependency versions to a specific hash for security

PR #656, opened 2025-08-24
Bumps swagger2pydantic action

jsamoocha:
I forgot to bump the action version ¯\_(ツ)_/¯


lwasser (approved):
Thank you for catching all of this. I"m sorry that I merged a PR that broke things. i will make sure notifications are turned on for when CI breaks here as well. I didn't even notice the build wasn't working.

PR #667, opened 2025-09-06
FIX: Remove false warning about response headers on authentication calls

### LINKED ISSUE #608: Stravalib returns No rates present in response headers for authentication calls
stravalib returns a quiet message upon authentication in the protocol module

`No rates present in response headers`

@jsamoocha noted:  I think I read somewhere that Strava doesn't count authentication requests re. rate limits, so that may be the edge case that leads to us logging a warning. We can finetune the rate limiter to ignore authentication calls to prevent that warning from being logged.

_Originally posted by @jsamoocha in https://github.com/stravalib/stravalib/pull/605#discussion_r1855403338_

While this doesn't break any builds, it would be good to fine tune things to avoid confusing users! This was brought up in [a discussion](https://github.com/orgs/stravalib/discussions/593) by g-normand  .

BPR02:
closes #608 

## Description

This PR updates the `_request` method in ApiV3 so that the rate limiter (`self.rate_limiter`) does not run for authentication calls. This is done by only running the rate limiter if `"/oauth/token"` is not in the url.

## Type of change

- [ ] Other (please describe)

## Does your PR include tests

- [ ] Yes

### DIFF
--- a/src/stravalib/protocol.py
+++ b/src/stravalib/protocol.py
@@ -219,7 +219,7 @@ class ApiV3(metaclass=abc.ABCMeta):
         raw = requester(url, params=params)  # type: ignore[operator]
         # Rate limits are taken from HTTP response headers
         # https://developers.strava.com/docs/rate-limits/
-        if "/oauth/token" not in url:
+        if "/oauth/" not in url:
             self.rate_limiter(raw.headers, method)
 
         if check_for_errors:
--- a/src/stravalib/protocol.py
+++ b/src/stravalib/protocol.py
@@ -219,7 +219,8 @@ class ApiV3(metaclass=abc.ABCMeta):
         raw = requester(url, params=params)  # type: ignore[operator]
         # Rate limits are taken from HTTP response headers
         # https://developers.strava.com/docs/rate-limits/
-        self.rate_limiter(raw.headers, method)
+        if "/oauth/token" not in url:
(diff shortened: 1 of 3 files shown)


yihong0618:
yes seems we can add a test?

yihong0618 (commented):
LGTM

src/stravalib/protocol.py
```diff
@@ -219,7 +219,8 @@ def _request(
         raw = requester(url, params=params)  # type: ignore[operator]
         # Rate limits are taken from HTTP response headers
         # https://developers.strava.com/docs/rate-limits/
-        self.rate_limiter(raw.headers, method)
+        if "/oauth/token" not in url:
```
jsamoocha, line 222: All oauth endpoints are without rate limit headers and are not throttled. Can you make the exclusion URL more generic, e.g. `/oauth/`? That will allow us to also catch e.g. `/oauth/deauthorize`.

jsamoocha (commented):
Thanks for this fix! As far as I can tell, there's no need to test logging side effects.

It would be nice if you could add a unit test on `protocol.ApiV3._request()`, mocking all dependencies, and asserting whether the rate limiter is called (or not) depending on the url.

lwasser:
Hi 👋🏻 thank you for this pr @BPR02 !! It looks like we are waiting for a test to be added. After this is merged we can cut a release 🚀

PR #746, opened 2026-09-14
feat: raise a specific fault for HTTP 429 responses

### LINKED ISSUE #560: ENH: Specific exception type for rate limited response from strava api

bensynapse:
I maintain Live Tennis API.

Closes #560.

## Description

A client configured with `rate_limit_requests=False` currently receives a generic `Fault` when Strava returns HTTP 429. Add `exc.TooManyRequests(Fault)` so applications can catch that response specifically. Existing `Fault` and `requests.exceptions.HTTPError` handlers still catch it, and the original response, rate-limit headers and diagnostic message are preserved.

The five new regression cases cover a normal error payload, empty/non-JSON/non-object bodies, and a public client request with automatic limiting disabled. The client test checks that there is no sleep or retry. The exception docstring appears in the API reference, and the changelog is updated.

## Type of change

## Does your PR include tests

- [x] Yes

Validation on Python 3.11 and 3.14:

- All **300 tests pass** when the test stub uses its existing bundled Swagger fallback, with no tests skipped or changed.
- The ordinary `nox -s tests-3.11 tests-3.14` command reports **296 passed and four failed** on each version. The same four failures occur on untouched main: the current live Swagger document lacks `/clubs/{id}/activities`, `/clubs/{id}/admins`, and `/clubs/{id}/members`, so their existing mock registrations raise `KeyError`.
- A separate public-client reproduction fails on main with `Fault` and passes after this change.
- `nox -s mypy`, `pre-commit run --all-files`, `nox -s build`, and `twine check dist/*` pass.
- `nox -s docs` exits 0 and renders the new exception, but reports existing `get_segment_efforts` docstring indentation errors and duplicate-object warnings.

Live authenticated Strava requests and macOS/Windows runs were not performed.


jsamoocha:
Hi, thanks for your contribution. Because Strava removed several endpoints from their API, the test suite on your branch fails. I fixed this in #745 , which is merged now. Can you please rebase this branch onto main and push it again?

bensynapse:
The rerun fails while loading three club endpoints that have disappeared from Strava’s live Swagger file. Your merged #745 removes those endpoints and tests.

I prepared a local rebase onto 45e7fd05, retained the HTTP 429 change, and resolved the changelog conflict. The project’s nox test session passes all 291 tests on Python 3.11, mypy passes, and all configured checks pass on the contribution.

I cannot publish that rebase yet: it also brings in four upstream workflow updates, and my push credential lacks workflow permission. The remote branch remains 29ba2147, so its CI is still red. Please hold review until I can publish the tested rebase and rerun CI.

PR #747, opened 2026-09-17
Use the new OAuth endpoint for deauthorization

### LINKED ISSUE #738: Replace oauth/deauthorize with oauth/revoke (deauthorize retires 1 June 2027)
Part of the 2026 Strava Developer Program changes tracked in #730.

## What changes

Strava replaced `oauth/deauthorize` with `oauth/revoke`. From the
[authentication docs](https://developers.strava.com/docs/authentication/):

- `POST https://www.strava.com/oauth/revoke` is available as of 1 June 2026.
- `POST https://www.strava.com/oauth/deauthorize` "will no longer be supported
  after June 1, 2027".

The two endpoints do not have the same interface:

| | `oauth/deauthorize` | `oauth/revoke` |
| --- | --- | --- |
| Auth | `access_token` parameter | HTTP Basic, `client_id:client_secret` |
| Body | `access_token` | `token` (access **or** refresh token), optional `token_type_hint` |
| Effect | Removes the app from "My Apps" | Revoking either token revokes the other |

## What stravalib has

```python
# src/stravalib/client.py:315-322
def deauthorize(self) -> None:
    """Deauthorize the application. ..."""
    self.protocol.post("oauth/deauthorize")
```

## Suggested work

Add a `revoke` path and deprecate `deauthorize` with
`exc.warn_method_deprecation`, which already takes an `alternative` and an
`alt_url`.

The new endpoint needs client credentials rather than the access token, so it
does not fit the existing `protocol.post` flow. `ApiV3` already holds
`client_id` and `client_secret` (populated by `_check_credentials`), so the
inputs are there, but the Basic Auth header is new.

Two things worth checking while doing this work — both observations, not
verified failures:

1. `resolve_url` prefixes anything that does not start with `http` with
   `/api/v3`, so `post("oauth/deauthorize")` appears to resolve to
   `https://www.strava.com/api/v3/oauth/deauthorize`. Elsewhere the OAuth
   endpoints are built with an explicit absolute URL
   (`f"https://{self.server}/oauth/token"`). Worth confirming which URL
   `deauthorize()` actually calls today.
2. On `main`, the access token is sent as a query parameter
   (`params["access_token"] = self.access_token`, `src/stravalib/protocol.py:245`),
   which is what `oauth/deauthorize` requires. PR #733 moves the token into the
   `Authorization` header. Whether `deauthorize()` still gets its required
   parameter after that change is worth a test either way.

Open question for a maintainer: keep `deauthorize()` as a thin wrapper that
calls the new endpoint, or deprecate it and add `revoke()` as a separate method
with the `token_type_hint` argument exposed?

## Timing

`oauth/deauthorize` keeps working until 1 June 2027. No deadline pressure, but
item 2 above may interact with PR #733 sooner than that.

mifuha:
Closes #738.

## Description

Keeps `Client.deauthorize()` as discussed in the issue and routes it to Strava's new `/oauth/revoke` endpoint.

It uses the application credentials available through `Client.protocol`, with Basic authentication and the access token sent as form data. It also handles the empty 200 response without refreshing the token first or following redirects.

The method name and return value stay the same, but callers now need application credentials as well as an access token. Missing inputs raise a clear `ValueError` before making a request.

I also updated the authentication guide and changelog, and added tests for the request format, missing credentials, error responses, redirects, and keeping authentication scoped to this request.

### Testing

* 306 unit and mocked integration tests passed, including 12 revocation cases
* lint, formatting, spelling and type checks passed
* no live Strava requests were made
* the changed authentication guide and API page rendered locally
* the full documentation build did not finish; the unchanged upstream comparison build also did not finish

## Type of change

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [x] This change requires a documentation update
* [ ] This is a documentation update
* [ ] This is a infrastructure update (docs, ci, etc)
* [ ] Other (please describe)

## Does your PR include tests

* [x] Yes
* [ ] No, I'd like some help with tests
* [ ] This change doesn't require tests

* [x] Yes, `changelog.md` is up-to-date.


jsamoocha:
Hi, thanks for your contribution. Because Strava removed several endpoints from their API, the test suite on your branch fails. I fixed this in #745 , which is merged now. Can you please rebase this branch onto main and push it again?

jsamoocha:
Thanks for the update! There are two issues needing attention before merge:

### Potentially expired access tokens

From Strava's [documentation](https://developers.strava.com/docs/authentication/#deauthorization), it is unclear whether they'll be able to find the to-be-revoked account in case of an expired access token. If they fail to find it, they'll respond with a 200 status and we'll never know the token was never revoked. To improve our chance of success, let's use the refresh token, e.g. :

```python
if self.refresh_token:
  data = {"token": self.refresh_token, "token_type_hint": "refresh_token"}
elif self.access_token:
  data = {"token": self.access_token, "token_type_hint": "access_token"}
else:
  raise ValueError("Deauthorization requires an access or refresh token.")
```

### Changelog

1. Can you add you user handle to the changelog item for contribution (like the other changelog entries)?
2. As the endpoint now requires the client id and - secret (and raises if these are missing), can you add one line to the "Breaking Changes" section?

mifuha:
Thanks! I’ve made those changes and pushed the update. Also added my handle and the breaking change note to the changelog. All 303 tests pass locally.

jsamoocha:
Awesome, thanks!

